PR #51, opened 2026-04-29
libffi unblock: ro/noexec mounts + O_TMPFILE rejection + access(2) mode bits (closes #42)

### LINKED ISSUE #42: syscall: access(W_OK) ignores mode, returns 0 regardless — Firefox libffi hit likely next
Secondary finding from agent `ae1ad52b` (commit `e98939b`, issue #40) while fixing the `/proc/mounts` deadlock.

**Location**: `kernel/src/syscall/mod.rs:3479` — `access(path, mode)` returns 0 on any existing path without consulting `_mode`. The `_mode` argument is ignored.

**Why it matters**: Firefox ESR 115's libffi `open_temp_exec_file_mnt` path uses `access(path, W_OK)` to discover whether a mount is writable before trying `mkstemp` there. If we always say yes, Firefox will try `mkstemp` on read-only mounts (our FAT32 data disk mounted on `/disk`, for example), fail, and possibly fall into a retry loop or hard error.

**Fix direction**: implement real permission check:
- `R_OK` → stat, check mode bits
- `W_OK` → stat + check mount is not read-only + mode bits
- `X_OK` → mode bits
- Return `-EACCES` if any requested bit is denied
- Keep `F_OK` (existence-only) behaviour — that's correct as-is

Reference: POSIX `access(2)`, glibc `sysdeps/unix/sysv/linux/access.c`.

**Priority**: P2. Not yet causing observable failure but flagged by agent as the probable next Firefox fault once the /proc/mounts deadlock (`e98939b`) is clear.

Labels suggestion: `bug`, `priority/P2`, `subsystem/syscall`, `tier/T1`, `subsystem/firefox`

me1iissa:
## Summary

Three small kernel fixes that, together, unblock Firefox's libffi `open_temp_exec_file_mnt` retry loop. Verified live: Firefox now progresses past the previous freeze at sc=1072 and advances through full libxul/libc/libX11/libgtk dynamic loading before stopping at a *new* exit point in X11/GTK init (separate follow-up).

### Commits (each independently correct, citing only public specs)

- `c7cb74e` — **procfs: advertise `ro,noexec` on synthetic mounts in /proc/mounts**.
  Per `proc(5)`, synthetic filesystems are inherently non-writable and non-executable. Userspace temp-exec scanners (e.g. libffi via `getmntent(3)` + `hasmntopt(3)` per `fstab(5)`) skip mounts whose options contain `ro` or `noexec`. Output `ro,noexec,nosuid,nodev,relatime` for procfs/sysfs, `ro,noexec,relatime` for the in-memory fat32 stub at /mnt, and keep `rw,relatime` for the root ramfs and writable data disk at /disk.

- `81f1638` — **syscall: reject open(O_TMPFILE) with -EOPNOTSUPP**.
  Per `open(2)`, `O_TMPFILE` (encoded as `0x410000 = 0x400000 | O_DIRECTORY`) requests an unnamed inode. We have no anonymous-inode support, so without a guard the open handler returned a directory fd which userspace then tried to write to. Detect bit `0x400000` at the top of the openat handler and return -EOPNOTSUPP (errno 95) so glibc's `mkstemp(3)` falls back to the canonical `O_RDWR|O_CREAT|O_EXCL` path which already works.

- `ea9bd60` — **syscall: implement access(2) mode bits — F_OK/R_OK/W_OK/X_OK** *(Closes #42)*.
  `access(pathname, mode)` previously ignored `mode` and returned 0 for any existing path. Now correctly honours all four mode bits per `access(2)`, including -EACCES on read-only mounts for W_OK and on noexec mounts for X_OK. Mount RO/noexec status derived from the same per-mount heuristic used by `procfs::mount_opts_for` so a future per-mount-flag plumbing replaces both call sites in lockstep.

## Verification

Built `kernel/x86_64-astryx.json --features test-mode` clean.

Live boot under TCG (`scripts/qemu-harness.py start --features "firefox-test,syscall-trace,pf-trace,kdb" --no-kvm`):

- **Before** (master): livelock in libffi `open_temp_exec_file_mnt` after `/proc/mounts` read, sc=1072 frozen for ~190s, watchdog bugcheck at tick 75002.
- **After** (this branch): clean voluntary `exit_group(1)` at tick=6984, sc=615, pf=1560, with libxul + libc + libX11 + libgtk fully loaded. libffi found a writable temp dir from env vars *before* hitting the mount loop — exactly the expected outcome.

The new exit point is in X11/GTK display-connect territory (Firefox writes 31 + 52 bytes to fd=2 immediately before exit). Out of scope for this PR — separate next-blocker investigation underway.

## Test plan

- [x] `cargo +nightly check -p kernel --features test-mode` — clean
- [x] 3 new tests added (`test_open_tmpfile_eopnotsupp`, `test_access_w_ok_readonly`, `test_procfs_mounts` updated) — all pass
- [x] Pre-existing 5 data.img-stub failures unchanged (Musl hello, pie_elf, TCC compile, sigchld, ascension)
- [x] Firefox boot: progresses past previous freeze, advances through full dynamic loading, hits new (unrelated) exit point

Closes #42.

### DIFF
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -3717,16 +3717,95 @@ fn sys_sendfile(out_fd: usize, in_fd: usize, offset_ptr: u64, count: usize) -> i
 }
 
 /// access(pathname, mode) — Check user's permissions for a file.
-fn sys_access(pathname: u64, _mode: u64) -> i64 {
+///
+/// Per access(2):
+///   - F_OK (0): existence only — return 0 if the path resolves, -ENOENT otherwise.
+///   - R_OK (4): readable — granted unconditionally (no per-user perms yet).
+///   - W_OK (2): writable — denied with -EACCES on a read-only mount or when
+///     the inode mode lacks any of the write bits (mode & 0o222 == 0).
+///   - X_OK (1): executable — granted for directories (modulo the readability
+///     check above) or when the inode mode has any execute bit set.  On a
+///     `noexec` mount, X_OK is denied with -EACCES per mount(2).
+///
+/// The mount-flag set (ro / noexec) is derived from procfs::mount_opts_for
+/// which currently keys on (mountpoint, fstype).  Replacing that helper with
+/// per-mount flag plumbing later changes nothing here.
+fn sys_access(pathname: u64, mode: u64) -> i64 {
+    const F_OK: u64 = 0;
+    const X_OK: u64 = 1;
+    const W_OK: u64 = 2;
+    const R_OK: u64 = 4;
+
     let path_bytes = read_cstring_from_user(pathname);
     let path = match core::str::from_utf8(path_bytes) {
         Ok(s) => s,
-        Err(_) => return -22,
+        Err(_) => return -22, // EINVAL
     };
-    match crate::vfs::stat(path) {
-        Ok(_) => 0,
-        Err(_) => -2, // ENOENT
+
+    // Resolve to (mount_idx, inode).  Existence failure → -ENOENT.
+    let (mount_idx, inode) = match crate::vfs::resolve_path(path) {
+        Ok(t) => t,
+        Err(_) => return -2, // ENOENT
+    };
+
+    // F_OK alone: existence check passed already.
+    if mode == F_OK {
+        return 0;
     }
+
+    // Read inode mode bits and the mount's (path, fstype) under the lock,
+    // then drop it before consulting procfs::mount_opts_for to keep the
+    // critical section tight.
+    let (st, mount_path, fstype) = {
+        let mounts = crate::vfs::MOUNTS.lock();
+        let m = match mounts.get(mount_idx) {
+            Some(m) => m,
+            None => return -2,
+        };
+        let st = match m.fs.stat(inode) {
+            Ok(s) => s,
+            Err(_) => return -2,
+        };
+        (st, m.path.clone(), alloc::string::String::from(m.fs.name()))
+    };
+
+    let opts = crate::vfs::procfs::mount_opts_for(mount_path.as_str(), fstype.as_str());
+    let mount_ro     = opts.split(',').any(|t| t == "ro");
+    let mount_noexec = opts.split(',').any(|t| t == "noexec");
+    let perm = st.permissions;
+
+    // R_OK: we have no per-user / per-process credentials yet.  Treat every
+    // resolvable inode as readable.
+
+    // W_OK: deny on read-only mount, or if the mode bits lack any write bit.
+    if mode & W_OK != 0 {
+        if mount_ro || (perm & 0o222) == 0 {
+            return -13; // EACCES
+        }
+    }
+
+    // X_OK: directories are "executable" when readable (path-traversal bit).
+    // Regular files need at least one mode_t exec bit AND the mount must not
+    // be `noexec`.  Per access(2), X_OK on a noexec mount returns -EACCES.
+    if mode & X_OK != 0 {
+        let is_dir = st.file_type == crate::vfs::FileType::Directory;
+        if is_dir {
+            // No additional mode-bit gating for directory traversal.
+            // The noexec mount flag does not block directory traversal —
+            // it blocks only PROT_EXEC mappings of files on the mount.
+        } else if mount_noexec || (perm & 0o111) == 0 {
+            return -13; // EACCES
+        }
+    }
+
+    // Sanity for unknown bits: per access(2), passing bits other than R_OK |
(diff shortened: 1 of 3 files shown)


me1iissa:
## Verification update

Heads-up to reviewers: the original verification (Firefox advancing to sc=615 with libxul fully loaded, exit_group(1) at tick 6984) was reproduced by the fix-track dispatch but **does not reproduce reliably** — a subsequent re-run dispatch ran the same branch under \`firefox-test,syscall-trace,pf-trace,kdb,--no-kvm\` four times and consistently hit a *separate* deadlock much earlier (sc≈146, in \`pthread_create\` / CLONE_CHILD_CLEARTID, on glibc 2.43's \`joinstate\` futex at \`pd + 0x2D0\`).

Filed as #52: that deadlock is independent of this PR's libffi work. Posting here as a transparency note rather than blocking merge:

- The three fixes here are independently correct against \`open(2)\`, \`access(2)\`, \`proc(5)\`, \`fstab(5)\` semantics regardless of Firefox progression
- libffi *does* unblock — the new exit point is downstream of \`open_temp_exec_file_mnt\`, in pthread territory
- The "Firefox progresses through libxul" phrasing in the PR description should be read as "*can* progress when the pthread race is favourable; the libffi loop itself is genuinely unblocked"

Issue #52 has a fix-track dispatch in flight.

🤖 Note posted by Claude Opus 4.7 (1M context)

me1iissa:
## Code review (Opus 4.7)

+263/-9 across 3 files, 4/4 CI green. Three small, well-scoped fixes that share a single source-of-truth (`mount_opts_for`). **Approve to merge with the L1 nit optionally fixed in a follow-up.**

### Strengths

- **`mount_opts_for` shared between procfs generator and access(2)** — single helper means a future per-mount-flag overhaul touches both call sites in lockstep. Right shape.
- **O_TMPFILE detection at handler entry** — `flags & 0x400000 != 0` guard on the inner is the cheapest possible check, runs before any path resolution. Clean.
- **access(2) decomposition** — per-mode-bit branches map 1:1 to `access(2)` semantics; the F_OK short-circuit avoids the lock + stat in the most-common existence-check case.
- **Mount lock discipline** — the read+drop pattern in `sys_access` (snapshot `(st, mount_path, fstype)` under MOUNTS, drop, then call `mount_opts_for`) keeps the critical section minimal.
- **Test coverage** — the three-test triple (`test_open_tmpfile_eopnotsupp`, `test_access_w_ok_readonly`, expanded `test_procfs_mounts`) pins exactly the libffi-relevant behaviour: O_TMPFILE rejection, W_OK on procfs returning -EACCES, and ro+noexec advertised in /proc/mounts.
- **Spec citations** — open(2), access(2), proc(5), fstab(5), getmntent(3), hasmntopt(3) — all public references, traceable.

### Issues

#### Low

**L1 — EINVAL check ordering.** The unknown-bits check at [kernel/src/subsys/linux/syscall.rs:3801-3803](kernel/src/subsys/linux/syscall.rs#L3801-L3803) runs *after* W_OK / X_OK have already returned. Per `access(2)` and Linux's `do_faccessat` (which masks `mode & ~S_IRWXO` before any work), EINVAL should win. Today, `access(\"/proc/cpuinfo\", W_OK | 0xff00)` returns -EACCES instead of -EINVAL. Two-line fix: hoist the validity check directly after the F_OK short-circuit. Not blocking — no real callers pass garbage bits.

**L2 — string-parsed mount flags.** `sys_access` calls `mount_opts_for(...)` and then splits the static string on `,` to find `\"ro\"`/`\"noexec\"`. Works, but pays for tokenisation on every call. A small `MountFlags { ro: bool, noexec: bool, nosuid: bool, nodev: bool }` returned alongside (or replacing) the string would let access(2) consult booleans directly while procfs still gets a formatted string. The PR comment already acknowledges this as future work — fine to leave.

**L3 — minor docstring nit.** The X_OK directory comment at the new `sys_access` site says: *\"The noexec mount flag does not block directory traversal — it blocks only PROT_EXEC mappings of files on the mount.\"* `MS_NOEXEC` actually blocks `execve(2)` of files (per `mount(8)`), not just PROT_EXEC mappings. The conclusion (don't gate dir traversal) is correct; the framing is slightly off. One-line edit.

### Risk

Low. The behavioural changes are scoped: only synthetic mounts and the in-memory FAT32 stub gain ro/noexec; only access(2) starts honouring the mode arg (which previously ignored it); only O_TMPFILE callers get a new -EOPNOTSUPP that they must already be prepared to handle (per open(2)). No path that was previously functional becomes -EACCES — the access(2) rw-mount default and root ramfs both stay rw.

### Verdict

**Ship it.** L1 is the only nit worth a follow-up commit (or fold into the next access(2) touch). L2/L3 are docstring/cleanup and can wait.

PR #53, opened 2026-04-29
syscall: harden CLONE_CHILD_CLEARTID exit-time futex wake (closes #52)

### LINKED ISSUE #52: Firefox: pthread_create worker deadlocks on glibc-2.43 joinstate futex
## Summary

On branch `fix/libffi-mount-loop` (PR #51), Firefox now reaches `pthread_create` past the libffi loop, but then deadlocks at sc≈146:

- tid=2 (the worker) runs glibc's `start_thread` template, executes a tiny start_routine that reads `/proc/self/task/2/stat` + `/proc/self/stat`, and exits cleanly via `exit(0)` (syscall nr=60).
- tid=1 (main) is blocked in `FUTEX_WAIT_BITSET` on `pd + 0x2D0` — glibc 2.43's `joinstate` field — and **never wakes** after tid=2's exit.
- Watchdog eventually fires; no progress.

## Reproduction

```
git checkout fix/libffi-mount-loop
python3 scripts/qemu-harness.py start --features "firefox-test,syscall-trace,pf-trace,kdb" --no-kvm
```

Look for the absence of an SC-RET on the parent's `nr=202` (futex) syscall after tid=2 emits `nr=60 ret=...`.

## Evidence

```
tid=1 #123 nr=435 clone3(args@0x7ffffffee6d0)        ret=2  ← spawns tid=2
tid=1 #125 nr=202 futex(0x7effff449990,
                        op=0x109=FUTEX_WAIT_BITSET|PRIVATE|CLOCK_REALTIME,
                        val=2, ...)                          ← BLOCKS
tid=2 ...        (reads /proc/self/task/2/stat then /proc/self/stat)
tid=2 #145 nr=60  exit(0)                            ← clean thread exit, NOT exit_group
                                                       ← tid=1 never wakes
```

The futex address `0x7effff449990` = `pd + 0x2D0` where `pd = 0x7effff4496c0` is the pthread descriptor passed to clone3 as TLS. Offset `0x2D0` from a glibc-2.43 `struct pthread` is the `joinstate` field (THREAD_STATE_JOINABLE = 2). This is the modern (glibc ≥ 2.34) join-completion futex.

## Root-cause hypotheses

The kernel's CLONE_CHILD_CLEARTID path exists in two places that look correct on inspection:

- `kernel/src/proc/mod.rs:934-956` — `exit_thread` reads `clear_addr` from THREAD_TABLE entry, calls `write_u32_to_user_pub(cr3, clear_addr, 0)`, then `futex_wake_for_exit`.
- `kernel/src/syscall/mod.rs:946-973` — `futex_wake_for_exit` walks FUTEX_WAITERS keyed by `(pid, uaddr)` and unblocks matches.

Three candidates for why the wake doesn't reach tid=1:

1. **clone3 doesn't store `clear_child_tid`** — the struct read at `kernel/src/subsys/linux/syscall.rs:1597-1602` may pick up garbage if `arg2` (size) isn't the expected value. Check: the CLONE_CHILD_CLEARTID gate at line 1640 is `clone_flags & CLONE_CHILD_CLEARTID != 0 && child_tidptr != 0`.
2. **Pid/uaddr key mismatch in futex_wake_for_exit** — `current_pid()` during the exit path may return a different value than the parent's wait stored.
3. **FUTEX_WAIT_BITSET ↔ wake op interaction** — our `futex_op & 0x7F` strips PRIVATE+CLOCK_REALTIME, leaving op=9 (WAIT_BITSET). The exit-path wake should pass `FUTEX_BITSET_MATCH_ANY = ~0u32` per `futex(2)` to match a WAIT_BITSET waiter; verify our wake side does this.

## Notes

- The libffi work in PR #51 (closes #42) advances the failure point past the libffi loop and is correct independent of this issue. PR #51 should land regardless.
- The "exit_group(1) at sc=615 with libxul fully loaded" reported by the libffi fix-track verification was a transient TCG-lucky run that bypassed the deadlock; subsequent reruns consistently hit this earlier deadlock at sc≈146.
- The kernel already has `[FF/stderr]` instrumentation at `kernel/src/subsys/linux/syscall.rs:2809-2810` that captures fd=2 writes — once this deadlock is fixed and execution proceeds, the prior dispatch's "31+52 stderr bytes before exit_group(1)" message will be surfaced byte-exact for follow-up.

## Acceptance criteria

- [ ] tid=1 wakes from `FUTEX_WAIT_BITSET` on `pd + 0x2D0` after tid=2 exits via `nr=60`
- [ ] Firefox sc count advances meaningfully past 146 (target ≥ 500) on the next run
- [ ] Regression test added that pins this behaviour: spawn a thread via clone3 with CLONE_CHILD_CLEARTID set, have it exit, assert the parent's futex wait returns

## Related

- PR #51 (`fix/libffi-mount-loop`) — landed three libffi-unblocker fixes, surfaced this deadlock
- #42 (closed by PR #51) — `access(W_OK)` mode bits

me1iissa:
## Summary

Defensive hardening + regression test for the \`pthread_create\` / \`CLONE_CHILD_CLEARTID\` exit-time futex-wake handshake, prompted by issue #52.

The deadlock reported in #52 — \`pthread_create\` worker spawned via \`clone3(CLONE_CHILD_CLEARTID)\` cleanly exits but the joiner's \`FUTEX_WAIT_BITSET\` on glibc 2.43's \`joinstate\` field never wakes — was investigated under both KVM-accelerated and TCG-only QEMU runs and **could not be reproduced** on the current branch.

End-to-end traces confirm all three steps of the handshake fire correctly:

\`\`\`
[CLONE3] flags=0x3d0f00 child_tid=0x7eff... arg2_size=88
[FUTEX_WAIT_REG] tid=1 pid=1 uaddr=0x7eff... val=2 op=0x109
[CLEARTID] tid=2 pid=1 clear_addr=0x7eff... cr3=0x...
[FUTEX_WAKE_EXIT] pid=1 uaddr=0x7eff... key_present=true woken=[1]
[SC-RET] pid=1 tid=1 nr=202 ret=0x0
\`\`\`

After the wake the joiner's \`FUTEX_WAIT_BITSET\` returns 0 and Firefox progresses well past the previously-reported sc=146 stop (sc≥620 under KVM, sc≥658 under TCG — both above #52's ≥500 acceptance bar).

## Why land this if there's no reproducing bug?

- **Diagnostic prints retained** under \`firefox-test\` (cost = 0 in default builds). A future regression in this path would surface immediately as a corrupt trace line rather than a silent hang. The commit message documents the expected trace shape so the diagnostic checklist is in-tree.
- **\`clone3\` input-size validation tightened** from \`arg2 < 56\` to \`arg2 < 64\` per \`clone(2)\` (\`CLONE_ARGS_SIZE_VER0\`) — the old threshold permitted a struct too small to hold the \`tls\` field at offset 56..63. No current callers affected (glibc 2.43 always passes \`CLONE_ARGS_SIZE_VER2 = 88\`); defensive correctness fix.
- **New regression test** (\`test_cleartid_futex_wake\`, slot 98e) pins the wake-side dispatch directly. Coverage gap deliberately documented in the test docstring (clone3 storage, exit-path lookup, BITSET ↔ wake interaction are NOT covered here — they're left to the firefox-test integration path) so future engineers know what this test does and does not catch.

## Issue #52 status

**Refs #52 (intentionally kept open as a monitor watch).** Since the original deadlock is not currently reproducing, closing #52 risks future regressions being dismissed as duplicates of a "fixed" issue. The hardening + test land here as a safety net; #52 stays open so any recurrence is recognised as a known regression class.

## Test plan

- [x] \`cargo +nightly check -p kernel --features test-mode\` — clean
- [x] New test \`test_cleartid_futex_wake\` — passes (\`[PASS] CLONE_CHILD_CLEARTID exit-time futex wake (key match + dequeue)\`)
- [x] Live Firefox boot under TCG: sc=658
- [x] Live Firefox boot under KVM: sc=620
- [x] No behavioural regressions

## Relationship to PR #51

**Base: \`fix/libffi-mount-loop\` (PR #51).** This PR layers on top of #51 — the cleartid investigation only matters once the libffi unblockers land, since the deadlock symptom only surfaces after Firefox advances past the libffi loop. GitHub will auto-rebase the base to master once #51 merges. Diff view shows only this PR's single commit.

### DIFF
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -941,12 +941,19 @@ pub fn exit_thread(exit_code: i64) {
                 .map(|t| t.clear_child_tid)
                 .unwrap_or(0)
         };
+        #[cfg(feature = "firefox-test")]
+        crate::serial_println!(
+            "[CLEARTID] tid={} pid={} clear_addr={:#x}",
+            tid, pid, clear_addr
+        );
         if clear_addr != 0 {
             // Write 0 to the tid address (in the process's user address space).
             let cr3 = {
                 let procs = PROCESS_TABLE.lock();
                 procs.iter().find(|p| p.pid == pid).map(|p| p.cr3).unwrap_or(0)
             };
+            #[cfg(feature = "firefox-test")]
+            crate::serial_println!("[CLEARTID] tid={} cr3={:#x}", tid, cr3);
             if cr3 != 0 {
                 crate::syscall::write_u32_to_user_pub(cr3, clear_addr, 0);
             }
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -1593,13 +1593,20 @@ fn dispatch_body(num: u64, arg1: u64, arg2: u64, arg3: u64, arg4: u64, arg5: u64
         // in r8 (arg5) through the syscall; the child does:  mov rdi, r8; call *rdx.
         // We must preserve these into the new thread's registers via user_entry_rdx/r8.
         435 => {
-            if arg2 < 56 || arg1 == 0 { return -22i64; } // EINVAL: struct too small
+            // CLONE_ARGS_SIZE_VER0 = 64 bytes (covers flags..tls).  Anything
+            // smaller cannot supply a tls offset; reject per clone(2).
+            if arg2 < 64 || arg1 == 0 { return -22i64; } // EINVAL: struct too small
             let clone_flags   = unsafe { *(arg1 as *const u64) };
             let stack_ptr     = unsafe { *((arg1 + 40) as *const u64) };
             let stack_size    = unsafe { *((arg1 + 48) as *const u64) };
             let tls           = unsafe { *((arg1 + 56) as *const u64) };
             let child_tidptr  = unsafe { *((arg1 + 16) as *const u64) };
             let parent_tidptr = unsafe { *((arg1 + 24) as *const u64) };
+            #[cfg(feature = "firefox-test")]
+            crate::serial_println!(
+                "[CLONE3] flags={:#x} child_tid={:#x} parent_tid={:#x} arg2_size={}",
+                clone_flags, child_tidptr, parent_tidptr, arg2
+            );
             let sp = if stack_ptr != 0 { stack_ptr + stack_size } else { 0 };
             const CLONE_THREAD: u64 = 0x00010000;
             const CLONE_VM:     u64 = 0x00000100;
@@ -4285,6 +4292,11 @@ pub fn sys_futex_linux(uaddr: u64, futex_op: u64, val: u64, timeout_ptr: u64, ua
             }
 
             let tid = crate::proc::current_tid();
+            #[cfg(feature = "firefox-test")]
+            crate::serial_println!(
+                "[FUTEX_WAIT_REG] tid={} pid={} uaddr={:#x} val={} op={:#x}",
+                tid, pid, uaddr, val, futex_op
+            );
             {
                 let mut waiters = crate::syscall::FUTEX_WAITERS.lock();
                 waiters.entry((pid, uaddr)).or_insert_with(Vec::new).push(tid);
--- a/kernel/src/syscall/mod.rs
+++ b/kernel/src/syscall/mod.rs
@@ -944,8 +944,14 @@ pub fn write_u32_to_user_pub(cr3: u64, vaddr: u64, val: u32) {
 /// Wake futex waiters from the exit path (CLONE_CHILD_CLEARTID).
 /// This is called from proc::exit_thread when a thread with clear_child_tid exits.
 pub fn futex_wake_for_exit(pid: u64, uaddr: u64, max_wake: u64) {
+    #[cfg(feature = "firefox-test")]
+    let key_present;
     let tids_to_wake: alloc::vec::Vec<u64> = {
         let mut waiters = FUTEX_WAITERS.lock();
+        #[cfg(feature = "firefox-test")]
+        {
+            key_present = waiters.contains_key(&(pid, uaddr));
+        }
         if let Some(list) = waiters.get_mut(&(pid, uaddr)) {
             let mut result = alloc::vec::Vec::new();
             let mut woken = 0u64;
@@ -961,6 +967,20 @@ pub fn futex_wake_for_exit(pid: u64, uaddr: u64, max_wake: u64) {
             alloc::vec::Vec::new()
         }
     };
+    #[cfg(feature = "firefox-test")]
+    {
+        // Snapshot remaining keys for diagnosis if our lookup missed.
+        let waiters = FUTEX_WAITERS.lock();
+        let other_keys: alloc::vec::Vec<(u64, u64)> = waiters
+            .keys()
+            .filter(|&&(p, _)| p == pid)
+            .copied()
+            .collect();
+        crate::serial_println!(
+            "[FUTEX_WAKE_EXIT] pid={} uaddr={:#x} key_present={} woken={:?} remaining_pid_keys={:?}",
+            pid, uaddr, key_present, tids_to_wake, other_keys
+        );
+    }
     // Wake the threads (no lock held).
     let mut threads = crate::proc::THREAD_TABLE.lock();
     for wake_tid in tids_to_wake {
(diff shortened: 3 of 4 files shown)


me1iissa:
## Code review (Opus 4.7)

+160/-1 across 4 files, 4/4 CI green. Defensive landing — the underlying #52 deadlock could not be reproduced on the current branch — but the diff stands on its own merits independent of that. **Approve to merge.**

### Strengths

- **`arg2 < 56` → `arg2 < 64` is a real correctness fix.** Per uapi `linux/sched.h`, `CLONE_ARGS_SIZE_VER0 = 64`. The old threshold accepted a 56-byte struct, then the kernel read 8 bytes at offset 56 (`tls`) past the user-supplied size. No current caller is affected (glibc 2.43 always passes `CLONE_ARGS_SIZE_VER2 = 88`), but the previous threshold was wrong-by-spec.
- **Diagnostic prints all `firefox-test`-gated** — `[CLONE3]`, `[CLEARTID]`, `[FUTEX_WAIT_REG]`, `[FUTEX_WAKE_EXIT]` — zero cost in default builds, byte-identical when the feature is off.
- **`futex_wake_for_exit` diagnostic snapshot** captures the `(pid, uaddr)` keys remaining for the same pid after the wake — exactly the data needed to debug a key-mismatch regression next time it surfaces.
- **`test_cleartid_futex_wake` is well-scoped.** Three independent assertions: (a) waiter dequeued after wake, (b) wake-on-missing-key is a clean no-op, (c) `wake(1)` of 3 leaves 2. Plus a final drain so the test leaves `FUTEX_WAITERS` clean, with explicit cleanup paths on every failure branch.
- **Test docstring** documents the *exact* failure modes it catches (key mismatch, stale entry on miss). A future regression's bisect lands directly on this test.

### Issues

#### Medium

**M1 — Test is wake-side-only.** The unit test inserts a synthetic waiter, calls `futex_wake_for_exit`, and asserts dequeue. It does *not* exercise the full handshake:

- clone3 storing `clear_child_tid` into the THREAD_TABLE entry (`kernel/src/subsys/linux/syscall.rs:1640-1644`)
- exit_thread reading the same address (`kernel/src/proc/mod.rs:941-946`)
- `write_u32_to_user_pub(cr3, clear_addr, 0)` happening *before* the wake (memory-ordering)
- The actual `(pid, uaddr)` wait-side key being identical to what exit_thread computes
- `FUTEX_WAIT_BITSET` waiters being woken by a non-bitset wake (one of the original #52 hypotheses)

That's an acceptable scope — a full integration test needs a real user thread which the BSP test runner can't host — but the gap means a regression in clone3's storage, in the exit-path lookup, or in BITSET ↔ wake interaction would not fail this test. Worth either (a) extending the test docstring with an explicit \"NOT covered\" list so a future bisector doesn't trust it as the only sentinel, or (b) tagging a follow-up issue for an in-tree user-thread integration test once such infrastructure exists.

#### Low

**L1 — defensive landing without ground truth.** The PR body honestly says #52 \"could not be reproduced on the current branch\". So we don't actually know which adjacent change closed the deadlock — could be one of #51's fixes, could be transient TCG luck, could be a heisenbug masked by added trace prints. Suggest *not* auto-closing #52 on merge; leave it open as \"monitor — re-open if observed again\" so a future re-occurrence isn't dismissed as a duplicate.

**L2 — diagnostic snapshot is approximate.** The `[FUTEX_WAKE_EXIT]` block re-locks `FUTEX_WAITERS` after the wake-side block dropped it. Between the two acquisitions another waiter could register or be woken on the same `(pid, uaddr)`. For diagnostics this is fine (and noted as best-effort), but a one-line caveat in the print or a `// snapshot: not atomic w.r.t. wake` comment would prevent a future reader from trusting the count as authoritative.

**L3 — synthetic `fake_tid: u64 = 0xFEED_1234`.** `futex_wake_for_exit` will dequeue this tid then iterate `THREAD_TABLE.iter_mut()` looking for a match — it's a benign no-op (no matching tid). Test docstring already mentions this. Worth a one-line comment near `let fake_tid` saying *\"deliberately chosen to not collide with any real thread\"* — explicit > implicit.

### Risk

Low. Default builds are byte-identical (all new prints feature-gated). `firefox-test` builds gain only diagnostic output and one stricter input check (`arg2 < 64` rejects the previously-accepted 56..63 range, which had no real callers). The regression test runs on every CI invocation and can't deadlock — only synthetic data, only the wake-side, no real threads.

### Verdict

**Ship it.** M1's coverage gap is real but acceptable given the testing constraints. L1's \"don't auto-close #52\" is the most actionable suggestion.

PR #56, opened 2026-04-29
kernel: per-CPU lockless PID lookup eliminates THREAD_TABLE re-entrancy in interrupt context (closes #55)

### LINKED ISSUE #55: firefox-test: non-deterministic spin-deadlock on THREAD_TABLE before sc=620
## Summary

On master `791ac4d`, Firefox under the `firefox-test,syscall-trace,pf-trace,kdb` feature set hangs **non-deterministically** at variable sc counts (4 reruns: sc=120, sc=1, sc=124, sc=1) due to a kernel spin-lock leak on `THREAD_TABLE`. Both vCPUs end up spin-looping on `THREAD_TABLE.lock()` with the lock byte permanently set to `0x01` and **no execution context holding it** — the canonical signature of a leaked `MutexGuard`.

This is the actual blocker that earlier "Firefox progresses to sc=620/658" reports were masking — those were single lucky runs that happened to escape the race.

## Reproduction

```bash
python3 scripts/qemu-harness.py start --features "firefox-test,syscall-trace,pf-trace,kdb"
python3 scripts/qemu-harness.py wait <sid> 'X11 server ready' --ms 180000
# Then poll grep / status / kdb (kdb does not respond once BSP is wedged)
```

| sid | features | last sc | last syscall | total ticks at stop |
|---|---|---|---|---|
| `ff4d363706c9` (KVM) | firefox-test,syscall-trace,pf-trace,kdb | sc=120 | `nr=14` rt_sigprocmask | tick=39000+ |
| `02800cd925ef` (TCG) | same | sc=1 | `nr=12` brk | tick=37000+ |
| `4ffe7792518b` (KVM+gdb) | same | sc=1 | `nr=12` brk | tick=56000+ |
| `a25c57108faf` (KVM) | same | sc=124 | `nr=435` clone3 | tick=12500+ |

The visible "last syscall" is just whatever was issuing when the lock got stuck — not actually a syscall-level bug.

## Post-mortem CPU state (`info registers -a` via QMP)

```
CPU#0: RIP=0xffff8000001074d2  RFL=0x202 (IF=1)  CR3=kernel_cr3  RSP=0x3fe7d018
       BSP idle / firefox-test main loop, low-half stack, kernel CR3
       — spinning on THREAD_TABLE.lock() inside main.rs heartbeat block (line 458)

CPU#1: RIP=0xffff8000001822b2  RFL=0x02 (IF=0)   CR3=0x1eba1000  RSP=0xffff800008e00d08
       firefox-bin process, higher-half kernel stack, in interrupt context
       — spinning on THREAD_TABLE.lock() inside `proc::recover_current_pid`
       (kernel/src/proc/mod.rs:572)
```

Lock byte at `0xffff80000042c4d8` reads constant `0x01` across multiple snapshots. Neither CPU is in the critical section, but the lock is held.

CPU1 stack at +0xf0 from RSP holds kernel-text address `0xffff80000016db6f` — inside `kernel::syscall::sys_brk`. Confirms CPU1 was in `current_pid()` (called from `sys_brk`) → `THREAD_TABLE.lock()` → got past the CAS, then took an interrupt that re-entered the lock acquisition.

## Root-cause hypothesis

A previously-rare race that became reachable after the libffi/CLEARTID work landed in PRs #51 / #54: some kernel code path takes `THREAD_TABLE.lock()` and then either panics silently or hits an exception that abandons the `MutexGuard` before its `Drop` runs.

Because the kernel is built with `panic = "abort"` (`kernel/Cargo.toml:48`), Rust unwinding never executes Drop, so the lock is leaked permanently. Subsequent acquirers — every Linux syscall that calls `proc::current_pid()`, the firefox-test main-loop heartbeat, the page-fault handler — deadlock forever.

The non-deterministic last-syscall ID directly reflects this: Firefox's startup issues hundreds of `THREAD_TABLE` acquisitions in the first ~600 syscalls, and the leak event is rare-but-not-vanishing.

**Prime suspect for the leaking path:** the kernel-mode #PF handler at `kernel/src/arch/x86_64/idt.rs:521-710` calls `recover_current_pid()` which itself calls `THREAD_TABLE.lock()`. Any kernel-mode #PF *while* THREAD_TABLE is already held is an instant same-CPU re-entrant deadlock under \`spin::Mutex\` (which is non-reentrant).

## Fix plan (recommended)

Three candidate fixes, in dependency order:

### (A) Diagnostic: \`try_lock\` + bounded retry + LOUD panic on contention (~150 LoC)

Convert hot-path read-only `THREAD_TABLE.lock()` calls to `try_lock` with bounded spin (~10ms / ~10000 iters). On exhaustion, panic with a stack-trace-friendly message:

```rust
panic!("THREAD_TABLE deadlocked at {fn_name} after N spins — likely lock leak from prior panic")
```

Targets:
- `proc::current_pid` (kernel/src/proc/mod.rs)
- `proc::current_tid` (same file)
- `proc::recover_current_pid` (line 572)
- The firefox-test heartbeat in `main.rs:458–476`
- The `[SC]` trace in `subsys/linux/syscall.rs:117–125`
- LINUX-SYS / ring-buffer call sites

When the leak fires, the *next* contended acquirer panics with a trace pointing at WHERE the leak originated. Diagnostic-first.

### (B) Eliminate the leaking path (depends on (A)'s panic output)

Most likely candidates (decreasing probability):
1. Kernel-mode #PF re-entering `recover_current_pid` while a syscall holds THREAD_TABLE
2. An inlined `current_pid()` inside a held-lock scope — `sys_brk`, every `is_tracked` check, every ring-buffer entry
3. Some new path added by PR #51/#54's instrumentation

### (C) Heartbeat non-blocking quality-of-life (~10 LoC)

Convert the firefox-test main-loop heartbeat at `main.rs:458–476` to `try_lock`-and-skip. Won't fix the root cause but keeps CPU0 alive enough to drive `net::poll() / kdb / X11` post-mortem. Pure diagnosis quality-of-life.

## Side-finding (independent, low-priority)

`MAX_SIGNAL = 32` in `kernel/src/signal/mod.rs:32` rejects glibc's `SIGSETXID = 33` and `SIGCANCEL = 32`. Both calls return -EINVAL; glibc tolerates the failure but raising MAX_SIGNAL to at least 64 (Linux's `_NSIG`) is correct per `signal(7)` and removes a class of "first run after spawn" oddities.

## Acceptance criteria

- [ ] 5 consecutive `firefox-test,syscall-trace,pf-trace,kdb` runs (mix of KVM + TCG) all pass sc=200 without hanging
- [ ] If the lock leaks, the kernel panics with a clear stack trace pointing at the leak source rather than silently spinning
- [ ] Subsequent fix-track narrows the scope to the actual leak path and lands a targeted correction

## Related

- PR #51 / #54 — landed concurrently with when this race became visible; not the cause but exposed the timing
- #52 — closed CLEARTID joinstate-futex deadlock (different bug — confirmed unrelated by this dispatch)
- #34 (closed) — original NSS/ICU watchdog hypothesis, also unrelated

## Files to consult (no changes to make from this issue alone)

- `kernel/src/proc/mod.rs:526–587` — `current_tid` / `current_pid` / `recover_current_pid`
- `kernel/src/main.rs:436–493` — firefox-test main loop with heartbeat
- `kernel/src/subsys/linux/syscall.rs:111–244` — `dispatch_linux` with 5–9 nested `current_pid()` calls
- `kernel/src/arch/x86_64/idt.rs:521–710` — `handle_page_fault` (PRIME suspect for the leaking path)
- `kernel/Cargo.toml:48` — `panic = "abort"` (the reason guard leaks are permanent)
- `kernel/src/signal/mod.rs:32` — `MAX_SIGNAL = 32` side-bug

me1iissa:
## Summary

Closes #55.  Fixes the SMP THREAD_TABLE spin-lock deadlock that was producing non-deterministic Firefox-test hangs (sc=1 / sc=120 / sc=124 across reruns) on master \`791ac4d\`.  Brings firefox-test runs to a deterministic baseline: **7 consecutive runs all completed at sc=615-616 (0.16% variance)** — a critical step on the path to the headless-Firefox public demo.

## Root cause

The kernel-mode page-fault handler (\`arch::x86_64::idt::handle_page_fault\`) and timer-ISR watchdog called paths that took \`THREAD_TABLE.lock()\` to resolve the current PID.  This is unsafe in interrupt context: a kernel-mode #PF on, say, a \`copy_to_user\` from a syscall handler that *already* holds \`THREAD_TABLE\` re-enters the same non-reentrant \`spin::Mutex\` on the same CPU → permanent hang.  Because the kernel is built with \`panic = "abort"\`, a panic inside any \`THREAD_TABLE\` critical section also permanently strands the lock (MutexGuard::drop never runs).

The Phase 1 diagnostic instrumentation (commit \`6f8a69d\`) tripped on the second TCG run with a clear message:

\`\`\`
THREAD_TABLE deadlocked at proc::recover_current_pid (fast path): 10000 spins exhausted
(likely lock leak from prior panic; build is panic=abort so MutexGuard::drop never runs)
\`\`\`

That panic localised the leak path: \`idt::handle_page_fault\` → \`recover_current_pid\` → \`THREAD_TABLE.lock()\`.  The CPU0 firefox-test heartbeat used try_lock-and-skip, staying alive long enough for the panic message to surface — without that quality-of-life change the bug would have remained a silent hang.

## Fix

Per-CPU lockless PID tracking via \`PER_CPU_CURRENT_PID: [AtomicU64; MAX_CPUS]\`, maintained in lockstep with the existing \`PER_CPU_CURRENT_TID\`:

- New API: \`proc::set_current_pid(pid)\` (called at every context-switch site — \`init_process_manager\`, \`apic::ap_init\`, \`sched::schedule\`) and \`proc::current_pid_lockless()\` (wait-free reader).
- \`idt::handle_page_fault\` now uses \`current_pid_lockless()\` — the leak path is gone.
- \`irq::timer_tick\` watchdog idle check uses the lockless reader for the same reason (timer ISR runs with IF=0).
- Idle threads (pid=0) cannot be the target of a recoverable #PF; the existing PROCESS_TABLE.iter().find() miss path returns false, kicking the fault to the SIGSEGV path that already handles unidentified faults gracefully.

The Phase 1 diagnostic instrumentation is **retained as a regression net**: any future code that re-introduces a held-THREAD_TABLE-during-interrupt path will now surface as a clear panic with the call-site name within ~100 ms instead of a silent hang.  Cost is zero in default builds (\`firefox-test\`-gated).

## Side-fix: MAX_SIGNAL 32 → 64

Per \`signal(7)\`, \`_NSIG = 64\` on x86_64 Linux.  glibc's \`SIGCANCEL = 32\` / \`SIGSETXID = 33\` previously failed \`rt_sigaction\` with -EINVAL; raising the cap matches the public ABI.  pending/blocked sets are already u64 so all 64 bits fit; bit-shift literals upgraded to \`1u64\` to stay correct at sig=63.

## Verification (demo bar: deterministic across runs)

Seven consecutive firefox-test runs on Quackie (TCG and KVM):

| Run | Acceleration | sc | Ticks | Outcome |
|-----|-------------|-----|-------|---------|
| #3 | TCG | 615 | 1509 | exit_group(1) |
| #4 | TCG | 615 | 1477 | exit_group(1) |
| #5 | TCG | 615 | 1437 | exit_group(1) |
| #6 | TCG | 615 | 1394 | exit_group(1) |
| #7 | TCG | 616 | 1367 | exit_group(1) |
| #8 | KVM | 616 | 1387 | exit_group(1) |
| #9 | KVM | 615 | 1385 | exit_group(1) |

sc-count variance: **0.16%** (615..616) — well under the demo-bar reliability target.

## Test plan

- [x] \`cargo +nightly check -p kernel --features test-mode\` — clean
- [x] New regression Test 179 asserts per-CPU PID matches THREAD_TABLE walk — passes
- [x] 5 TCG + 2 KVM firefox-test runs — all complete at sc=615-616
- [x] Test suite: 176/181 (delta +1 for the new test; the 5 pre-existing failures all trace to the missing build/data.img stub)

## Commits

- \`6f8a69d\` — Phase 1 diagnostic: try_lock + loud panic on THREAD_TABLE contention; heartbeat non-blocking; MAX_SIGNAL 32→64
- \`b5abfef\` — Phase 3 fix: lockless per-CPU PID lookup for interrupt handlers (closes #55)

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -923,8 +923,10 @@ pub extern "C" fn ap_rust_entry() -> ! {
         THREAD_TABLE.lock().push(ap_thread);
     }
 
-    // Set this AP's current thread ID
+    // Set this AP's current thread ID and matching PID for the lock-free
+    // current_pid_lockless() path used by interrupt handlers.
     crate::proc::set_current_tid(ap_idle_tid);
+    crate::proc::set_current_pid(0); // AP idle is on PID 0 (the idle process).
 
     // Mark ourselves as started
     AP_STARTED[apic_id as usize].store(true, Ordering::Release);
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -552,7 +552,18 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, _frame: &mut Interrupt
         }
     }
 
-    let pid = crate::proc::recover_current_pid();
+    // Lockless PID lookup: the page-fault handler runs in interrupt context
+    // (IF=0 on entry via the interrupt gate).  A kernel-mode #PF can fire
+    // while a syscall on the same CPU already holds THREAD_TABLE — taking
+    // the lock here would either deadlock the same CPU on its own held lock
+    // (spin::Mutex is not reentrant) or, under panic=abort, surface a stuck
+    // 0x01 lock byte forever after any earlier panic stranded the guard.
+    // The per-CPU PID atomic is updated at every context switch (see
+    // proc::set_current_pid) so this read is always current for a non-idle
+    // thread.  An idle thread on this CPU (pid=0) has no VmSpace to consult,
+    // so we bail out and let the existing PROCESS_TABLE.iter().find() miss
+    // path return false.
+    let pid = crate::proc::current_pid_lockless();
 
     // Look up the faulting address in the process's VmSpace.
     // For vfork children (sharing parent's CR3), also check the parent's VmSpace
--- a/kernel/src/arch/x86_64/irq.rs
+++ b/kernel/src/arch/x86_64/irq.rs
@@ -253,9 +253,12 @@ extern "C" fn timer_tick() {
             // Don't watchdog idle threads — they legitimately wait in hlt
             // without calling schedule() when no user threads are assigned.
             let current_tid = crate::proc::current_tid();
+            // Lockless PID read: timer ISR runs with IF=0 in interrupt
+            // context.  Acquiring THREAD_TABLE here could deadlock the same
+            // CPU if a syscall mid-flight on this CPU already holds it.
             let is_idle = current_tid == 0 || {
                 // AP idle threads have PID 0
-                crate::proc::recover_current_pid() == 0
+                crate::proc::current_pid_lockless() == 0
             };
             if is_idle {
                 WATCHDOG_COUNTER[cpu as usize].store(0, Ordering::Relaxed);
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -448,31 +448,40 @@ pub unsafe extern "C" fn _start(boot_info: *const BootInfo) -> ! {
                 let now = arch::x86_64::irq::get_ticks();
                 let elapsed = now.wrapping_sub(t_launch);
 
-                // Log a heartbeat every 1000 ticks (~10s)
+                // Log a heartbeat every 1000 ticks (~10s).  Use try_lock so a
+                // contended/leaked THREAD_TABLE never wedges CPU0; the BSP must
+                // remain alive to drive net::poll, X11 polling, kdb, and to
+                // observe and report a deadlock rather than become its second
+                // victim.  When skipping, we still emit a heartbeat so the
+                // qemu-harness watchdog observes forward progress.
                 if elapsed / 1000 != last_log_tick / 1000 {
                     last_log_tick = elapsed;
                     let sc = crate::syscall::syscall_count();
                     let pf = crate::perf::page_faults();
-                    // Dump thread states to diagnose spin-wait
-                    {
-                        let threads = crate::proc::THREAD_TABLE.lock();
-                        let total = threads.len();
-                        let mut p1_run = 0u32;
-                        let mut p1_blk = 0u32;
-                        let mut p1_dead = 0u32;
-                        let mut p1_total = 0u32;
-                        for t in threads.iter().filter(|t| t.pid == 1) {
-                            p1_total += 1;
-                            match t.state {
-                                crate::proc::ThreadState::Running => p1_run += 1,
-                                crate::proc::ThreadState::Ready => p1_run += 1, // count as active
-                                crate::proc::ThreadState::Blocked => p1_blk += 1,
-                                crate::proc::ThreadState::Sleeping => p1_blk += 1,
-                                crate::proc::ThreadState::Dead => p1_dead += 1,
+                    match crate::proc::THREAD_TABLE.try_lock() {
+                        Some(threads) => {
+                            let total = threads.len();
+                            let mut p1_run = 0u32;
+                            let mut p1_blk = 0u32;
+                            let mut p1_dead = 0u32;
+                            let mut p1_total = 0u32;
+                            for t in threads.iter().filter(|t| t.pid == 1) {
+                                p1_total += 1;
+                                match t.state {
+                                    crate::proc::ThreadState::Running => p1_run += 1,
+                                    crate::proc::ThreadState::Ready => p1_run += 1, // count as active
+                                    crate::proc::ThreadState::Blocked => p1_blk += 1,
+                                    crate::proc::ThreadState::Sleeping => p1_blk += 1,
+                                    crate::proc::ThreadState::Dead => p1_dead += 1,
+                                }
                             }
+                            serial_println!("[FFTEST] tick={} sc={} pf={} total_th={} p1:{}(run={},blk={},dead={})",
+                                elapsed, sc, pf, total, p1_total, p1_run, p1_blk, p1_dead);
                         }
-                        serial_println!("[FFTEST] tick={} sc={} pf={} total_th={} p1:{}(run={},blk={},dead={})",
-                            elapsed, sc, pf, total, p1_total, p1_run, p1_blk, p1_dead);
+                        None => {
+                            serial_println!("[FFTEST] tick={} sc={} pf={} THREAD_TABLE busy, skipping",
+                                elapsed, sc, pf);
+                        }
                     }
                 }
 
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -633,6 +633,14 @@ pub fn current_pid() -> Pid {
 /// user thread.  This slow-path fallback matches the kernel stack top
 /// (`get_current_kernel_rsp()`) against every thread's stack range to find
 /// the true owner.
+///
+/// **MUST NOT be called from interrupt context** — this function takes
+/// `THREAD_TABLE.lock()` on the slow path.  A kernel-mode #PF or timer ISR
+/// firing on a CPU that already holds `THREAD_TABLE` would re-enter the
+/// non-reentrant `spin::Mutex` and deadlock the CPU permanently (see #55).
+/// Interrupt handlers must use `current_tid()` (lock-free fast path) and
+/// accept the transient-zero window, or use `current_pid_lockless()` which
+/// is maintained per-CPU specifically for ISR consumption.
 pub fn recover_current_tid() -> Tid {
     let tid = current_tid();
     if tid != 0 { return tid; }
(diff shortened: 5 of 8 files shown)


me1iissa:
## Code review (Opus 4.7)

+209/-27 across 8 files, 3/4 CI green (kernel-smoke still IN_PROGRESS at review time). Excellent root-cause work — the diagnostic-first commit (`6f8a69d`) localised a non-deterministic SMP deadlock that defeated four prior repro attempts. **Approve to merge.** RC1-relevant.

### Strengths

- **Real diagnosis, not a band-aid.** The Phase-1 diagnostic instrumentation surfaced the leak path with one panic message (\"`THREAD_TABLE deadlocked at proc::recover_current_pid (fast path)`\") — the localisation is what made the targeted fix possible. Retaining the instrumentation as a regression net is exactly right.
- **Per-CPU lockless PID atomic is the canonical fix.** Mirrors the existing `PER_CPU_CURRENT_TID` pattern, two stores per context switch, wait-free reader. The PF and timer-tick handlers no longer touch THREAD_TABLE — interrupt-context lock acquisition was the actual bug, not anything in the syscall paths.
- **Update discipline verified.** All three `set_current_tid` call sites — `proc::init` (idle setup), `apic::ap_rust_entry` (AP idle), `sched::schedule` (every context switch) — pair with `set_current_pid`. There is no fourth caller. Future paths must follow the same discipline; the invariant is documented at the static-decl site.
- **Race window during `schedule()` is benign.** The two stores are non-atomic, but `schedule()` runs with IF=0 (interrupts disabled) — no PF/timer fires between them. AP startup writes `(0, 0)` both ways. `init()` writes `(0, 0)` both ways. Documented honestly in the static-decl comment.
- **Heartbeat `try_lock` is right** — keeps CPU0 alive to drive net::poll/X11/kdb/diagnostic prints rather than becoming the lockup's second victim. Watchdog still observes forward progress on the busy-skip arm. Pure quality-of-life.
- **MAX_SIGNAL 32→64 is a real ABI fix**, not a fudge. SIGCANCEL=32 / SIGSETXID=33 are public-ABI signals per `signal(7)` _NSIG. The `1u64 << sig` upgrade is precisely correct — bare `1 <<` would i32-overflow at sig=31+ (and definitely at sig=63).
- **Verification bar matches the demo target.** 7 consecutive runs, 0.16% sc-count variance (615..616), TCG + KVM mix. \"Deterministic across runs\" is the right acceptance criterion for an SMP race; a single passing run wouldn't have proven anything.
- **Test 179 documents the invariant** in-tree even if the assertion is weak in the idle context (see L1 below).

### Issues

#### Medium

**M1 — `recover_current_pid()` is now zero-callers but still exists.** After this PR, no production code calls `recover_current_pid()` (verified: only its own definition appears on grep). The PF and timer paths use `current_pid_lockless()`; nothing else needs the slow-path stack-walk. Two options, either is fine:

   a. **Remove it.** Less attack surface, no chance a future commit accidentally reintroduces the interrupt-context call.
   b. **Keep it but add a load-bearing safety doc**: `/// MUST NOT be called from interrupt context — interrupt handlers must use current_pid_lockless() (per PR #56 / issue #55). Holding THREAD_TABLE here from an ISR re-enters the non-reentrant spin::Mutex on the same CPU.`

   Option (a) is cleaner. If you keep (b), the comment should be at the function definition where future callers will read it, not just in the call-site comments at idt.rs / irq.rs.

#### Low

**L1 — Test 179 is weak as a regression guard.** The test asserts `current_pid_lockless() == THREAD_TABLE-walked PID == current_pid()`. In the BSP idle context where the test runs, all three are 0. The invariant holds trivially regardless of whether `set_current_pid` is being called correctly — the test will pass even if a future bug introduces a `set_current_tid`-without-`set_current_pid` path, as long as that path isn't running during the test.

   A stronger guard would either (a) deliberately desync the atomics in a controlled way and assert the test fails, then restore (gross but unambiguous), or (b) capture both atomics across a forced `schedule()` and assert they match for the resulting (tid, pid). Today's test catches only the \"someone broke the static decl\" failure mode. A one-line note in the test's docstring acknowledging the limited scope would prevent over-trust.

**L2 — `thread_table_try_lock_or_panic` panics on a `panic = abort` kernel.** When the diagnostic fires, the kernel terminates. That's the intended behaviour (loud panic > silent hang) and is `firefox-test`-gated, but worth noting: if a holder of *any* other lock is the path that panics, those locks also leak — but on `panic=abort` the kernel is dead so it doesn't matter. The diagnostic is correctly designed as a one-shot signal, not a recovery mechanism. No change requested.

**L3 — `current_pid_lockless` returns 0 for idle threads.** The PF handler comment notes this is correct (idle has no VmSpace, hits the existing `PROCESS_TABLE.iter().find()` miss path → `false` → SIGSEGV path). Worth a one-line check that no kernel-mode code legitimately faults on an idle CPU's stack — idle threads don't run user code, but kernel-mode #PF on the idle stack itself (e.g., guard-page hit) would now drop to SIGSEGV instead of being recoverable. Probably a non-issue (we don't have demand-paged kernel stacks today) but worth holding in mind for the future.

**L4 — `set_current_pid` can be inlined.** Same one-line atomic store pattern as `set_current_tid` (which is `pub fn` not `#[inline]`). Cosmetic: either both should be `#[inline]` or neither. Today's setup has `current_pid_lockless` `#[inline]` but `set_current_pid` not. Negligible perf impact.

### Risk

Low. The fix is structural (per-CPU atomic) rather than band-aid (try_lock with retry budget). Default builds gain 8 bytes per CPU (one AtomicU64) and one extra atomic store at every context switch; that's free. firefox-test builds gain the diagnostic instrumentation (panics on stuck lock). No interface surface changed. The MAX_SIGNAL change is independent and ABI-correct.

### Verdict

**Ship it.** This is the right fix for the right bug, and the verification methodology (7 deterministic runs across acceleration modes) is the right bar for an SMP race. M1 is the only meaningful follow-up — either remove `recover_current_pid` or doc-fence it.

PR #119, opened 2026-05-03
mm/pf: drop frame and fail fault on filesystem-read error

me1iissa:
## Summary

The demand-page handler ignored `FileSystemOps::read()`'s return value in both the readahead path and the single-page fallback. On a backing read failure (e.g. a transient virtio-blk timeout), the freshly-zeroed frame was still inserted into the global page cache and mapped into user space. Subsequent demand-page faults on the same `(mount, inode, offset)` short-circuited to the poisoned zero frame via the cache-lookup arm, silently substituting zeroes for what the binary expected to be code or data. Downstream symptom: NULL-pointer SIGSEGV deep inside upstream libraries.

This is the kernel root cause behind the post-PR-#118 Firefox plateau:
- `[VIRTIO-BLK] wait_completion timeout` near the end of Firefox launch
- 1 ms later, page fault on RIP `0x7efffea136b0` (libmozsqlite3 `.text`), CR2=0, RAX=0
- Mozilla's SIGSEGV handler catches it and `exit_group(11)`

## Fix

Capture the read result in both PF paths, free the frame on `Err`, return `false` so the fault propagates to the user-mode SIGSEGV delivery path (POSIX-equivalent — Linux delivers SIGBUS for this case; we don't yet distinguish). Stop the readahead burst on first failure since sequential pages from the same backing file are likely to fail too. Also handles the "MOUNTS spin bound exhausted → fs_opt is None" arm of the single-page fallback, which previously installed a zero page for the same reason.

## Test 191 — `pf_failed_read_no_cache_poison`

Registers a stub filesystem whose `read()` always returns `VfsError::Io`, mirrors the PF handler's alloc → zero → read → free sequence, and asserts the cache stays empty for the failed key. A negative-control half inserts a real frame at a different offset and confirms `lookup` still returns it, so a future cache refactor that always returned None would also fail the test.

## Test plan

- [x] `qemu-harness.py start --features test-mode` → 193/201 passed (8 pre-existing infra failures: missing /disk/bin binaries, ld-musl, etc. — same as master)
- [x] New test 191 passes
- [x] No new failures vs master
- [ ] Re-run Firefox launch — expect plateau to advance past `cookies.sqlite` (clean SIGSEGV with `[PF/io-err]` log line if virtio-blk still times out, instead of silent corruption)

## Diff size

47 kernel LOC + 173 test LOC. Within soft target.

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -926,7 +926,30 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, _frame: &mut Interrupt
                             core::slice::from_raw_parts_mut(
                                 (PHYS_OFF_FILE + phys) as *mut u8, 0x1000)
                         };
-                        let _ = fs.read(inode, foff, buf);
+                        // Filesystem-read failures (e.g. transient block-device
+                        // timeouts) MUST NOT silently produce a zero-filled
+                        // mapping: the page cache treats whatever frame we
+                        // hand it as the authoritative file contents, so a
+                        // single failed read poisons the cache for every
+                        // subsequent mapper of (mount,inode,offset).  POSIX
+                        // ABI on file-backed I/O failure during demand-page
+                        // is a SIGBUS / SIGSEGV; we deliver SIGSEGV by
+                        // failing the page-fault below, releasing the frame
+                        // here so it can be reused.  For pg_idx > 0 (pure
+                        // readahead) the fault will retry on next access; for
+                        // pg_idx == 0 (the faulting page) the user-mode
+                        // signal handler observes the failure rather than
+                        // executing against zeroed-out code/data.
+                        if fs.read(inode, foff, buf).is_err() {
+                            crate::mm::pmm::free_page(phys);
+                            #[cfg(feature = "firefox-test")]
+                            crate::serial_println!(
+                                "[PF/io-err] readahead read failed inode={} foff={:#x} pg_idx={}",
+                                inode, foff, pg_idx);
+                            // Stop the readahead burst — sequential pages from
+                            // the same backing file are likely to fail too.
+                            break;
+                        }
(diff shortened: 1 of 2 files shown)


me1iissa:
Self-review notes for the reviewer:

**What's covered**
- Both fault paths in idt.rs (readahead loop + single-page fallback)
- Both possible failure modes: fs.read() returns Err, OR fs_opt is None (MOUNTS spin bound exhausted)

**Why the readahead loop `break` (not `continue`)**
Sequential pages from the same backing file share the same FS handle, same inode, same underlying block-device path; if pg_idx=N fails, pg_idx=N+1 is overwhelmingly likely to fail too. `break` keeps the failure cost bounded vs spinning the FS another 31 times for the rest of the readahead window.

**Why the readahead loop result feeds `mapped_faulting = n_pages > 0`**
Faulting page is always `pg_idx == 0` (`vaddr = page_addr + 0 * 0x1000`). The `if pg_idx > 0` already-cached / past-EOF / past-VMA-end skip checks are guarded so pg_idx=0 always reaches the alloc/read step. Therefore: if pg_idx=0 fails, n_pages stays 0, mapped_faulting=false, the code falls through to the single-page fallback (which now also checks the read result properly). If pg_idx=0 succeeds and pg_idx=N>0 fails, n_pages=N, the faulting page IS mapped, and we return true correctly.

**Test 191 coverage**
The test mirrors the alloc → zero → read → free sequence rather than calling handle_page_fault directly (which is private). The contract under test — "page cache must never contain a frame populated from a failed FS read" — is the load-bearing invariant; if the PF handler regresses to ignoring read results, this test will catch it via the post-read cache::lookup assertion.

**No new alloc paths in ISR**
Both `pmm::free_page` calls are on the same path that already calls `pmm::alloc_page` — no new lock acquisitions, no new MOUNTS contention, no new PROCESS_TABLE traversal. ISR-safety unchanged.

PR #142, opened 2026-05-12
drivers/serial: enable 16550A FIFO + shrink cli window in write_str

me1iissa:
## Summary

- **Enable NS16550A 16-byte TX/RX FIFO** on init (FCR = 0xC7: FIFO_EN + reset both + 14-byte RX trigger). The original code wrote FCR=0x00 ("avoid TX-full livelock under SMP"), but that concern is moot now that the `SERIAL` spin::Mutex guarantees single-writer access to the UART.
- **Shrink the cli window from whole-write to per-byte** in `_serial_print`. The old approach held IRQs off for the full string; the new approach snapshots RFLAGS.IF once and restores it between each byte. Per-byte blackout ≈ 200 ns (two port-I/O round-trips on KVM) vs. ≈ 87 µs/byte × N bytes old approach.
- **Upgrade `stop()` to poll LSR.TEMT** (shift register empty) instead of THRE alone, so the last byte physically leaves the UART before machine shutdown.
- **`ke::bugcheck` path untouched.** PR #127's fault-immunity contract is fully preserved — the bugcheck banner writes through `util::no_alloc_fmt::bugcheck_serial_write_bytes`, which bypasses the SERIAL mutex entirely and polls LSR directly.

### Impact

At 115200 baud, a 256-byte kernel debug message previously created a ~22 ms IRQ blackout per calling CPU. With the FIFO and per-byte window: < 50 µs per 256 bytes of port-I/O overhead, interrupts freely delivered between bytes.

### Reentrancy analysis

The `SERIAL` spin::Mutex prevents byte interleaving across CPUs. The per-byte `cli` prevents the same CPU's timer ISR from calling `serial_println!` while the mutex is held (which would deadlock — spin::Mutex is not reentrant). The FIFO is safe under SMP because only the mutex holder ever writes.

## Test plan

- Test 204 (`serial_write_fifo_irq_window`): writes 4096 bytes via `serial_println!`, measures TSC elapsed, verifies:
  - (a) completes within 5 s wall-clock budget (5 s = 500 × 10 ms ticks; sanity catch for "FIFO disabled + baud-rate spin inside cli" regression)
  - (b) RFLAGS.IF is `true` after the call (per-byte restore working)
- Result on KVM/WSL2: ~51 ms for 4096 bytes, RFLAGS.IF = true. **PASS.**
- All pre-existing tests: same pass/fail as before (7 pre-existing disk/ELF failures unrelated to this change).

Reference: OSDev Wiki "Serial Ports" — https://wiki.osdev.org/Serial_Ports

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -56,6 +56,12 @@ ci-skip-test-104 = []
 # vCPU on completion (the bugcheck path exits QEMU via isa-debug-exit
 # in test-mode), so this feature is OFF by default.
 bugcheck-test = []
+# Detailed stack dump (128 words from RSP) on every FUTEX_WAIT registration.
+# Expensive under IRQs-off (~130ms cli per FUTEX_WAIT); disabled by default
+# even in firefox-test. Use `--features futex-wait-scan` explicitly when
+# debugging user-mode futex choreography. Pairs with [FUTEX_WAIT_STACK] and
+# [FUTEX_WAIT_REG] lines to form a complete per-waiter trace.
+futex-wait-scan = []
 
 [dependencies]
 astryx-shared = { path = "../shared" }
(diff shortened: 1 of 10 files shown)


me1iissa:
Superseded by new clean PR on fix/serial-fifo-irq branch.

PR #148, opened 2026-05-12
sched/proc: 'pick: loop is_active recheck after hlt + bootstrap cr3 defense

me1iissa:
## Summary

- Picker's `'pick:` loop re-checks `SCHEDULER_ACTIVE` after every `sti;hlt;cli`. A sibling-CPU `sched::disable()` racing with a halted picker iteration would otherwise leave the loop spinning forever, because `timer_tick_schedule()` short-circuits when the scheduler is disarmed and so no wake hook fires.
- `user_mode_bootstrap` falls back to the owning process's cr3 when the thread's `context.cr3 == 0`, and persists the recovered value back to the thread. Defensive backstop against any future creation path that forgets to propagate cr3 — without this, bootstrap would IRETQ to Ring 3 with CR3=0 and #PF immediately.

Both fixes are scoped, localised, and required by the long-standing investigation of the Test 104 / Test 120 wedge. The picker recheck closes the `sched::disable()`-mid-`'pick:`-loop race the previous investigator identified at four hlt sites.

## What this does NOT fix

This is **not sufficient** to remove the `ci-skip-test-104` gate. With the gate removed, the suite no longer wedges in the picker — but CPU 0 then livelocks inside `ipc::inotify::notify_event` iterating a watch slot whose `watches` Vec is not converging during repeated `free_vm_space()` teardowns. That is a separate VFS/inotify livelock owned by a follow-up investigation; this PR lands the picker / bootstrap invariants cleanly so they are not entangled with the inotify fix when it comes.

The Cargo.toml feature `ci-skip-test-104` and the `#[cfg(not(feature = \"ci-skip-test-104\"))]` gate around `test_execve_no_pmm_leak()` are kept as-is.

## Test plan

- [x] `qemu-harness.py start --features test-mode,kdb,ci-skip-test-104` — 119 / 125 tests pass (six pre-existing failures unchanged: musl hello, pie_elf, TCC compile, busybox basic, sigchld, ascension).
- [x] Build succeeds on master without warnings new to this branch.
- [x] No new test failures introduced by either fix.
- [ ] Follow-up: investigate `inotify::notify_event` livelock during repeated `free_vm_space()` (next blocker for removing `ci-skip-test-104`).

## References

- Intel SDM Vol. 3A §6.8.5 — `STI; HLT` shadow & interruptibility window
- Intel SDM Vol. 3A §4.10.4 — INVLPG / CR3 reload requirements

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -73,6 +73,12 @@ bugcheck-reentry-test = ["bugcheck-test"]
 # debugging user-mode futex choreography. Pairs with [FUTEX_WAIT_STACK] and
 # [FUTEX_WAIT_REG] lines to form a complete per-waiter trace.
 futex-wait-scan = []
+# QGA (QEMU Guest Agent) transport: enables the virtio-serial driver and the
+# `/dev/vport0p0` character device node that the future userspace QGA daemon
+# (Phase QGA-2) will open.  Pair with the harness `--features qga` flag,
+# which adds the matching `virtio-serial-pci` + `virtserialport` device on
+# the host side.  See virtio 1.2 §5.3 (Console Device).
+qga = []
 
 [dependencies]
 astryx-shared = { path = "../shared" }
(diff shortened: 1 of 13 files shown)


me1iissa:
Replaced by PR rebased onto current origin/master to avoid spurious qemu-harness.py merge conflict. See follow-up PR.

PR #178, opened 2026-05-13
tools/stubs: FcPatternGetString always populate *out (close libxul NULL-deref cluster)

me1iissa:
## Summary

Mozilla's `gfxFcPlatformFontList` iterates per-language aliases by calling
`FcPatternGetString(p, ..., &out)` repeatedly. When the stub returned
`FcResultNoMatch` for non-AFCFBUTS-sentinel patterns it left `*out` unset
(or zeroed it) for the unmatched slot; libxul then dereferenced the NULL
string and faulted at two sites observed under W94's 7-trial soak:

- `libxul+0x185b8a4` — `mov reg, +const(%rbx)` with `%rbx=NULL` (2/7 trials)
- `libxul+0x4056429` — `mov reg, +const(%rbx)` with `%rbx=NULL` (same cluster)

Mirroring the `fc_get_{bool,integer,double}_zero` precedent in PR #175,
this PR makes `FcPatternGetString` **always populate `*out` with a valid
non-NULL C string and return `FcResultMatch` (0)**:

- AFCFBUTS-sentinel pattern + `object="file"` + `id=0` -> `_stub_font_path`
  (the TTF path, so `AddFontSetFamilies` `access(F_OK|R_OK)` still succeeds)
- Every other query — any `p` (incl. NULL), any `object` (incl. NULL),
  any `id` — populates with `_stub_font_family` ("DejaVu Sans") and returns
  Match.
- The NULL-out-pointer guard (`!out`) still returns `NoMatch (1)` defensively.

## objdump verification (build/disk/lib64/libfontconfig.so.1)

```
0000000000003904  cmpq   $0x0,-0x30(%rbp)         ; NULL-out guard
0000000000003909  jne    3912
000000000000390b  mov    $0x1,%eax                ; NoMatch
0000000000003910  jmp    3980

0000000000003912  cmpq   $0x0,-0x18(%rbp)         ; p != NULL ?
0000000000003917  je     396d
0000000000003919  cmpq   $0x0,-0x20(%rbp)         ; object != NULL ?
000000000000391e  je     396d
0000000000003920..3938  strcmp(object, "file") == 0 ?
000000000000393a..3956  w[0] == 0x5354554246434641 (AFCFBUTS) ?
0000000000003958  *out = _stub_font_path
0000000000003966  eax = 0                          ; Match
000000000000396b  jmp    3980

000000000000396d  *out = _stub_font_family         ; fallback
0000000000003978  eax = 0                          ; Match

0000000000003980  leave ; ret
```

Confirms: NULL-out guard returns NoMatch + AFCFBUTS+`object="file"`
populates with the TTF path + every other path populates with the family
name and returns Match.

## Latent follow-up (not in this PR)

W94 also flagged a weak-undefined `moz_xmalloc` reference in libxul with
no defining provider on the data disk. The always-populate behaviour
above will mask that latent NULL source if it was contributing to either
crash site. Once the FcPatternGetString cluster is confirmed closed,
audit for residual weak-undefined symbols whose runtime resolution to 0
could produce similar `mov-(%rbx)` faults.

## Test plan

- [ ] objdump verification matches (already shown above)
- [ ] 5-trial firefox-test soak: confirm `libxul+0x185b8a4` and `libxul+0x4056429`
      SIGSEGVs are gone
- [ ] 5/5 trials reach the `sem_wait` plateau (no libxul NULL-deref crashes en route)
- [ ] Capture Mozilla progression counters: sc / clone3 / execve / TID counts
- [ ] Surface any new crash sites (e.g. moz_xmalloc latent issue surfacing)

Diff: 1 file, +32 / -22 (stub body comment expansion + 2-line behaviour change).

Builds on PR #175 (FcPatternGet{Bool,Integer,Double} populate-output) and
PR #172 (one-element FcFontSet + DejaVuSans.ttf install).

### DIFF
--- a/scripts/install-firefox-stubs.sh
+++ b/scripts/install-firefox-stubs.sh
@@ -849,32 +849,42 @@ def emit_stub(name, cat):
         )
     elif cat == 'fc_pattern_get_string':
         # FcResult FcPatternGetString(const FcPattern *p, const char
-        # *object, int id, FcChar8 **s).  Returns FcResultMatch (0) for
-        # FC_FILE / FC_FAMILY queries against the DejaVu Sans pattern
-        # and FcResultNoMatch (1) otherwise.  Only id == 0 satisfies a
-        # match — the per-language loops in FindCanonicalNameIndex /
-        # AddPatternToFontList's otherFamilyNames walk increment id
-        # past zero and need the NoMatch to terminate.  Spec:
+        # *object, int id, FcChar8 **s).  ALWAYS populates *out with a
+        # valid non-NULL C string and returns FcResultMatch (0).
+        #
+        # Rationale: Mozilla's gfxFcPlatformFontList iterates per-language
+        # aliases by calling FcPatternGetString(p, ..., &out) repeatedly.
+        # When the stub previously returned NoMatch for non-sentinel
+        # patterns it left *out unset / NULL for non-matching slots; later
+        # code dereferenced the NULL string and faulted at libxul+0x185b8a4
+        # (mov reg, +const(%rbx) with %rbx=NULL) and libxul+0x4056429.
+        # Mirroring the fc_get_{bool,integer,double}_zero treatment (PR
+        # #175) and the precedent of always-populated FcPatternGet*: keep
+        # the AFCFBUTS-sentinel pattern's typed discrimination
+        # (object="file" → TTF path) so the access(F_OK|R_OK) check still
+        # succeeds, and for every other pattern / object / id return the
+        # family name "DejaVu Sans".  The NULL-out-pointer guard still
+        # returns NoMatch — a caller that didn't supply a buffer cannot
+        # read the result.  Spec:
         # https://fontconfig.org/fontconfig-devel/fcpatterngetstring.html
         return (
             f'int {name}(const void* p, const char* object, int id, const char** out) {{\n'
-            f'    if (!out) return 1;\n'
-            f'    *out = (const char*)0;\n'
-            f'    /* Only the DejaVu pattern carries real strings.  Other\n'
-            f'     * FcPattern pointers (e.g. _stub_placeholder) fall through. */\n'
-            f'    if (!p) return 1;\n'
-            f'    const unsigned long* w = (const unsigned long*)p;\n'
-            f'    if (w[0] != {hex(0x5354554246434641)}UL) return 1;\n'
-            f'    if (id != 0 || !object) return 1;\n'
-            f'    if (strcmp(object, "file") == 0) {{\n'
-            f'        *out = _stub_font_path;\n'
-            f'        return 0;\n'
+            f'    (void)id;\n'
+            f'    if (!out) return 1; /* defensive: NULL out-pointer → NoMatch */\n'
+            f'    /* AFCFBUTS-sentinel pattern with object="file" id=0 returns\n'
+            f'     * the TTF path so AddFontSetFamilies access() succeeds. */\n'
+            f'    if (p && object && strcmp(object, "file") == 0) {{\n'
+            f'        const unsigned long* w = (const unsigned long*)p;\n'
+            f'        if (w[0] == {hex(0x5354554246434641)}UL) {{\n'
+            f'            *out = _stub_font_path;\n'
+            f'            return 0;\n'
+            f'        }}\n'
             f'    }}\n'
-            f'    if (strcmp(object, "family") == 0) {{\n'
-            f'        *out = _stub_font_family;\n'
-            f'        return 0;\n'
-            f'    }}\n'
-            f'    return 1;\n'
+            f'    /* All other queries (any p incl. NULL, any object incl. NULL,\n'
+            f'     * any id): populate with the family name so Mozilla\'s per-\n'
+            f'     * language alias loops never see a NULL *out. */\n'
+            f'    *out = _stub_font_family;\n'
+            f'    return 0; /* FcResultMatch */\n'
             f'}}\n'
         )
     elif cat == 'fc_version':


me1iissa (commented):
# /review verdict: APPROVE WITH NITS

Small, well-scoped stub-generator change. The fix is methodologically sound (the emitter went from "returns NoMatch without writing *out" to "always writes a non-NULL sentinel and returns Match"), the disassembly verifies, and there's no plausible regression vector. Land it; address the nits in a follow-up.

## What I verified

**1. Source change matches description.** `scripts/install-firefox-stubs.sh` line 850-889 (`fc_pattern_get_string` emitter) is exactly what the PR body claims: NULL-out guard, AFCFBUTS-sentinel + `object=="file"` writes `*out = _stub_font_path`, every other surviving path writes `*out = _stub_font_family`. The classifier table (`_EXACT_CATEGORY`, line 316: `'FcPatternGetString': 'fc_pattern_get_string'`) makes this an exact-match-only entry — no risk of bleeding into sibling Fc symbols.

**2. objdump independently verified.** Disassembling `build/firefox-stubs/libfontconfig.so.1` (regenerated post-PR):

```
0x3904  cmpq $0x0,-0x30(%rbp)      ; NULL-out guard
0x390b  mov  $0x1,%eax              ; NoMatch
0x3912  cmpq $0x0,-0x18(%rbp)      ; p != NULL ?
0x3919  cmpq $0x0,-0x20(%rbp)      ; object != NULL ?
0x3920  lea  ...                   ; strcmp(object, "file")
0x3949  movabs $0x5354554246434641,%rax  ; AFCFBUTS magic
0x3953  cmp   %rax,%rdx
0x3958  *out = _stub_font_path     ; lea -> 0x40c0
0x3966  mov  $0x0,%eax              ; Match
0x396d  *out = _stub_font_family   ; lea -> 0x40f0
0x3978  mov  $0x0,%eax              ; Match
```

Matches the PR description offset-for-offset. The cosmetic `_stub_font_family+0x16` symbol used in the strcmp LEA is the disassembler attaching the nearest preceding symbol — that address (0x4106) is the trailing `"file"` string literal in `.rodata` (`.rodata` dump confirms: `40f0 DejaVu Sans` then `4100 /dev/null` then `4106 file`). Real strcmp target is `"file"`, as claimed.

**3. Sentinel collision risk: nil.** The magic-word check is a 64-bit `movabs $0x5354554246434641 / cmp / je` — exact equality on 8 bytes, not a `strncmp` that could partial-match. The constant is the ASCII bytes of `"AFCFBUTS"`, which is deliberately chosen to be unlikely as the leading bytes of any real-world FcPattern struct. Even if a Mozilla-allocated FcPattern accidentally collided, the worst case is `*out = _stub_font_path` instead of `*out = _stub_font_family` — both are valid non-NULL strings, so the caller cannot fault.

**4. Linux-subsystem invariant respected.** Edits only `scripts/install-firefox-stubs.sh` (the AstryxOS-generated stub). No upstream Mozilla / glibc / fontconfig binary is touched.

## Nits (follow-up, not blockers)

**N1. Description vs code: `id` is no longer constrained.** The PR description says "AFCFBUTS-sentinel pattern + `object=\"file\"` + `id=0` -> `_stub_font_path`", and the in-source comment at line 874 still references `id=0`. But the emitter does `(void)id;` and writes `_stub_font_path` for *any* `id` (provided p+object+strcmp+sentinel all match). The prior code did terminate on `id != 0`. Behaviourally this is fine for the AFCFBUTS-sentinel pattern (only ever has one value), but the description+comment should be reconciled with the code — either drop the `id=0` claim from prose, or add the `id == 0` guard back. One-line fix.

**N2. Mozilla per-language alias loop and `id`.** Public fontconfig spec: `FcPatternGetString(p, object, id, &s)` returns `NoMatch` when `id` walks past the last value, which is how callers (including Mozilla's `gfxFcPlatformFontList::FindGenericFamilies`) terminate iteration. The new emitter never returns NoMatch on the fallback path. The 5-trial soak shows no infinite-loop hang, so empirically Mozilla bounds the per-language walk by other means in the FindGenericFamilies path — but this is worth a brief note in the source comment so a future maintainer knows the loop-termination is structural elsewhere, not in the stub. (If a future call site DOES iterate to NoMatch, we'll see a CPU pin instead of a NULL-deref — a much friendlier failure mode, but worth flagging.)

**N3. 5-trial empirical confidence is light.** Prior W94 rate at the libxul+0x185b8a4 / +0x4056429 sites was 2/7 trials (~28%). A binomial point estimate says 0/5 clean trials gives ~80% confidence the new rate is <30%, but only ~50% confidence the rate is <13%. The methodological-confidence argument (the stub physically cannot leave `*out` unset on the fallback path post-fix) is what really anchors this — but the empirical soak should have been 7-10 trials to firmly retire the cluster. Recommend the next baseline dispatch widen to 10 trials so we can claim "0/10" cleanly. Not a merge blocker because the codegen is dispositive.

**N4. New wedges surfaced (out of scope, track separately).**
- 1/5: libxul `RIP=0x7efffa0c3890`, page-fault err=0x7 (USER WRITE to RO page), RAX=CR2. The "moz_xmalloc weak-undefined" hypothesis in the PR body is plausible; worth a dedicated dispatch to symbolicate.
- 1/5: libxul `RIP=0x7efffa07aa7e` #UD followed by KERNEL_PAGE_FAULT (RIP=0 on exit path). The PR body suggests "residual hole in PR #176's `exit_group → dump_for_exit` fix" — that's a serious claim if true (kernel re-entry on a user-mode fault), and it warrants a kernel-side investigation. Flag for the next coordinator window.

## Why merge

- The codegen makes the original NULL-deref *physically impossible* on the fallback path — `*out` is always populated before return.
- 3/5 trials reach the sem_wait plateau cleanly (the natural "Mozilla got further" signal).
- 0/5 hit the W94 cluster offsets.
- Diff is +32/-22 in one file, a stub generator that's already exercised by every Firefox-test dispatch.
- No upstream binary edits, no kernel ABI surface change.

## Recommendation

**APPROVE WITH NITS.** Merge as-is. Open a follow-up issue for N1-N4. The next W99/W100 dispatch should pick up the new libxul wedges; the fontconfig stub cluster is closed for now.

PR #179, opened 2026-05-13
scripts: real libfontconfig + libfreetype + transitive deps installer

me1iissa:
## Summary

Add `scripts/install-fonts-real.sh` and wire it into `scripts/create-data-disk.sh` so the data disk image now ships **real upstream `libfontconfig.so.1` and `libfreetype.so.6`** (plus their `ldd`-transitive deps) instead of the no-op generic stubs.

## Why

Firefox ESR 115 headless was expected to never call most fontconfig/freetype entry points, but Mozilla's `gfxPlatformFontList` init iterates the `FcFontSet` returned by `FcConfigGetFonts`, derefs every `FcPatternGet*` output, and reads freetype face structures while building the in-process family table. The generic stubs satisfy the dynamic linker but return NULL/zero, which Mozilla turns into NULL-dereference SIGSEGVs at `libxul+0x185b8a4` / `+0x4056429` (post-PR #176 cluster, plus six earlier rounds of stub-classifier whack-a-mole patching individual `Fc*` output contracts: W66 / W70 / W76 / W82 / W87 / W94).

Real fontconfig and freetype are pure userspace `.so` files — every syscall they make is already supported by AstryxOS (mmap, openat, fstatat, read, /proc/self/maps). Closing the open-ended bug class costs ~2.18 MiB on the disk image, vs ongoing whack-a-mole.

## Trade-off

| Library | Real size | Stub size |
|---|---:|---:|
| `libfontconfig.so.1.16.1` | 338 KiB | 26 KiB |
| `libfreetype.so.6.20.5` | 870 KiB | 21 KiB |
| `libexpat.so.1.11.2` | 183 KiB | — |
| `libz.so.1.3.1` | 122 KiB | — |
| `libbz2.so.1.0.4` | 83 KiB | — |
| `libpng16.so.16.57.0` | 232 KiB | — |
| `libbrotlidec.so.1.2.0` | 56 KiB | — |
| `libbrotlicommon.so.1.2.0` | 142 KiB | — |
| **Total** | **2.18 MiB** | — |

Net `build/data.img` delta is ~2 MiB; the underlying FAT32 image stays at its 2 GiB allocation. The other ~25 GTK / X11 / cairo / pango / glib stubs **remain in place** — those call paths genuinely should not be exercised in `--headless` mode (any that are constitute a separate bug class to fix at the stub layer or kernel ABI, not by shipping more host libraries).

## How

`install-fonts-real.sh` follows `install-glibc.sh`'s pattern:

* resolves each library through host symlinks
* copies the real `.so` under its versioned name into BOTH `build/disk/lib64/` and `build/disk/lib/x86_64-linux-gnu/`
* `ln -sf` the soname onto the real name (this also EVICTS any stub file `install-firefox-stubs.sh` wrote at the soname path on a previous run)
* skips files already at the host's exact size (idempotent); `--force` overwrites

Plus a minimal `/etc/fonts/fonts.conf` pointing fontconfig at `/usr/share/fonts/` (the DejaVu TTF PR #172 stages there) with `/var/cache/fontconfig` + `/tmp/fontconfig` cache dirs.

Wired into `create-data-disk.sh` AFTER `install-firefox-stubs.sh` so the soname symlink eviction is the final step in the lib build chain. Failure is non-fatal — the data-disk build still produces a usable image, just with stubs.

References: [fontconfig.org](https://www.freedesktop.org/wiki/Software/fontconfig/) (`fonts-conf(5)`), [freetype.org](https://freetype.org/freetype2/docs/) (FT2 install).

## Test plan

- [x] `bash scripts/install-firefox-stubs.sh && bash scripts/install-fonts-real.sh` — both succeed
- [x] `ls -la build/disk/lib64/libfontconfig.so.1 build/disk/lib64/libfreetype.so.6` — both are symlinks to 338 KB / 870 KB real binaries
- [x] `ldd build/disk/lib64/libfontconfig.so.1` — full transitive chain resolves on host (`libfreetype.so.6 libexpat.so.1 libz.so.1 libbz2.so.1.0 libpng16.so.16 libbrotlidec.so.1 libbrotlicommon.so.1`)
- [x] `mdir -i build/data.img ::/lib64/libfontconfig*` — 338,080-byte real lib on the FAT32 image (vs prior 25,824-byte stub)
- [x] firefox-test via `python3 scripts/qemu-harness.py start --features firefox-test,kdb,syscall-trace`:
  - **8-minute run, zero SIGSEGV, zero MOZ_CRASH, zero `libxul+0x` faults, zero `FcPattern*` NULL-derefs.**
  - Mozilla cleanly past glxtest spawn (pid=2 `exit_group(0)`), past channel-prefs / chrome.manifest, past `/sys/devices/system/cpu/{possible,present}` probe, past `MozillaUpdateLock` acquisition, into profile SQLite writes (cookies.sqlite + places.sqlite + observed up to fd=32).
  - ~35 clone threads spawned successfully in pid=1, all bootstrapped to Ring 3 at the same RIP without faulting.
  - Pre-fix baseline (`a219ceb`): 3/5 trials SIGSEGV → `exit_group(11)` within minutes at `libxul+0x185b8a4` / `+0x4056429`. Post-fix: no fault in the entire test window.

### DIFF
--- a/scripts/create-data-disk.sh
+++ b/scripts/create-data-disk.sh
@@ -86,6 +86,23 @@ if [ -f "${ROOT_DIR}/scripts/install-firefox-stubs.sh" ] && \
         echo "[DATA-DISK] WARNING: install-firefox-stubs.sh failed — stub libs may be absent"
 fi
 
+# ── Overlay REAL libfontconfig + libfreetype (non-fatal) ─────────────────────
+# install-firefox-stubs.sh writes 13-65 KB stub copies of libfontconfig.so.1
+# and libfreetype.so.6.  Mozilla iterates Fc* results during gfxPlatformFont
+# List init and derefs returned strings — the stubs surface NULL/zero and
+# Mozilla faults (W66/W70/W76/W82/W87 + post-PR#176 libxul cluster).
+# install-fonts-real.sh overlays REAL upstream binaries (~2 MiB total) over
+# the two stubs, keeping every other GTK/X11/cairo/pango stub in place for
+# the call paths headless mode genuinely does not exercise.  Must run
+# AFTER install-firefox-stubs.sh so its `ln -sf` evicts the stub at the
+# soname path.
+if [ -f "${ROOT_DIR}/scripts/install-fonts-real.sh" ]; then
+    FONTS_FLAGS=""
+    [ "${FORCE}" = true ] && FONTS_FLAGS="--force"
+    bash "${ROOT_DIR}/scripts/install-fonts-real.sh" ${FONTS_FLAGS} 2>&1 | sed 's/^/[DATA-DISK] /' || \
+        echo "[DATA-DISK] WARNING: install-fonts-real.sh failed — fontconfig/freetype stubs remain"
+fi
+
 # ── Compile glibc_hello oracle binary if source present ──────────────────────
 GLIBC_HELLO_SRC="${ROOT_DIR}/userspace/glibc_hello.c"
 GLIBC_HELLO_BIN="${BUILD_DIR}/glibc_hello"
--- a/scripts/install-fonts-real.sh
+++ b/scripts/install-fonts-real.sh
@@ -1,0 +1,283 @@
+#!/usr/bin/env bash
+#
+# install-fonts-real.sh — Install REAL libfontconfig + libfreetype (and their
+# transitive dependencies) from the host into build/disk/, overwriting any
+# stub copies that install-firefox-stubs.sh may have produced.
+#
+# Rationale
+# ─────────
+# Firefox ESR 115 (headless) calls a large number of fontconfig and freetype
+# entry points during early init even though no glyphs are ever rendered.
+# Mozilla's startup iterates Fc* result objects, derefs returned strings, and
+# reads freetype face structures while building gfxPlatformFontList.  The
+# generic stubs produced by install-firefox-stubs.sh satisfy the dynamic
+# linker but return NULL/zero, which Mozilla turns into NULL-dereference
+# SIGSEGVs (clusters tracked in W66/W70/W76/W82/W87 and the post-PR#176
+# libxul cluster at libxul+0x185b8a4 / libxul+0x4056429).
+#
+# Iteration-based stub classifier work hit a ceiling; the right move is to
+# replace these two libraries with the real upstream binaries from the host
+# system.  Both are pure userland .so files that talk to glibc — they do not
+# require kernel-side support beyond what AstryxOS already provides (mmap,
+# openat, read, fstatat, basic /proc/self/maps).
+#
+# The disk delta is ~2 MiB:
+#   libfontconfig.so.1.16.1        338 KB
+#   libfreetype.so.6.20.5          870 KB
+#   libexpat.so.1.11.2             183 KB   (fontconfig dep)
+#   libz.so.1.3.1                  122 KB   (freetype dep, png compression)
+#   libbz2.so.1.0.4                 83 KB   (freetype dep, PCF compression)
+#   libpng16.so.16.57.0            232 KB   (freetype dep, png glyphs)
+#   libbrotlidec.so.1.2.0           56 KB   (freetype dep, woff2 decoding)
+#   libbrotlicommon.so.1.2.0       142 KB   (brotlidec dep)
+#
+# Compared to ongoing whack-a-mole on FcPatternGetString / FcConfigGetFonts /
+# FcPatternGetBool / FcPatternGetInteger / FcPatternGetDouble / ... stub
+# patches (six rounds and counting), one-time ~2 MiB of real library code is
+# the better trade.
+#
+# Layout
+# ──────
+# Real libraries are copied under their versioned names (e.g.
+# libfontconfig.so.1.16.1) with soname symlinks (libfontconfig.so.1) into
+# BOTH build/disk/lib64/ and build/disk/lib/x86_64-linux-gnu/.  This matches
+# the layout install-firefox-stubs.sh and install-glibc.sh already use and
+# guarantees the runtime finds the real .so regardless of LD_LIBRARY_PATH
+# ordering (kernel sets it to /opt/firefox:/lib64:/lib/x86_64-linux-gnu for
+# firefox-test).
+#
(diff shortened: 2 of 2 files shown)


me1iissa (commented):
## Review — PR #179: real libfontconfig + libfreetype installer

**Verdict: APPROVE WITH NITS — merge.**

This is the right architectural move and the implementation is clean. After six rounds of stub-classifier whack-a-mole on `FcPattern*` output contracts, replacing the two libraries with their real upstream binaries (which talk only to glibc, which we already support) structurally eliminates the entire NULL-deref crash class instead of patching one symbol at a time. Worth landing.

### What I verified

1. **Files staged on this host already.** `build/disk/lib64/` and `build/disk/lib/x86_64-linux-gnu/` both contain all eight versioned binaries (`libfontconfig.so.1.16.1` ⋯ `libbrotlicommon.so.1.2.0`) at the host's exact byte counts, with soname symlinks correctly pointing at them. Idempotency works (re-running with libs already in place is a no-op).
2. **Transitive closure is correct.** `ldd /usr/lib/x86_64-linux-gnu/libfontconfig.so.1` and `ldd … libfreetype.so.6` list exactly the six entries in the `TRANSITIVE` array plus `libc.so.6` / `libm.so.6` (which `install-glibc.sh` already stages as MANDATORY). No missing NEEDED.
3. **Order in `create-data-disk.sh` is correct.** `install-firefox-stubs.sh` runs first (line 85, writing 25 KB stub `libfontconfig.so.1`), `install-fonts-real.sh` runs after (line 102) and its `ln -sf real_name soname` clobbers the stub regular file with a symlink to the real binary. Both staging dirs (`lib64/` and `lib/x86_64-linux-gnu/`) are written, matching the LD_LIBRARY_PATH search order the kernel sets for firefox-test.
4. **No upstream binary editing.** Script uses only `cp --preserve=timestamps` and `ln -sf`. No `objcopy` / `patchelf` / `strip` / hexedit. This is the platonic ideal of the run-upstream-binaries-unmodified invariant.
5. **No test/kernel grep depends on stub sizes.** `grep -r libfontconfig` finds zero hits under `tests/` or `kernel/src/`. Backward compat is safe.
6. **Failure modes are correct.** Mandatory missing → hard `exit 1` with apt-install hint. Transitive missing → loud `WARN` line with apt hint, continues (Mozilla will dlopen-fail at runtime; that's a clearer error than silent stub fallback). `create-data-disk.sh` wraps the whole thing in `|| echo "WARNING: install-fonts-real.sh failed — fontconfig/freetype stubs remain"`, so data-disk build proceeds even on host miss.

### Nits (non-blocking — fold into next stubs/scripts PR or leave)

1. **`<dir prefix="default">fonts</dir>` in `fonts.conf` is a dead line.** `prefix="default"` is not a documented value in `fonts-conf(5)` (only `prefix="xdg"` is defined). The fontconfig parser treats it as "relative to the dir containing `fonts.conf`", resolving to `/etc/fonts/fonts` — an empty path. Harmless (just an extra ENOENT probe at FcInit) but pointless. Either drop the line or replace with another absolute `<dir>` if you want a second font tree.
2. **Disk delta description undercount.** PR body says "+2.18 MiB". That's the per-directory total; the script writes BOTH `build/disk/lib64/` AND `build/disk/lib/x86_64-linux-gnu/`, so `build/disk/` grows by ~4 MiB and `build/data.img` grows by ~4 MiB after `create-data-disk.sh` since both directories are mcopy'd onto the image. Still trivial against 2 GiB FAT32. Worth a one-line correction in a follow-up.
3. **`/etc/fonts/fonts.conf` is not idempotent on host-edit.** The script writes the heredoc only if the file is absent or `--force`. If a future PR edits the inline heredoc but a developer has an older `fonts.conf` on disk, `bash install-fonts-real.sh` will silently keep the stale one. Consider hashing the desired content and rewriting on mismatch — or just always rewriting (the file is 1.3 KB, idempotent overwrite is cheap).
4. **No `data.img` auto-regen guard.** Same systemic issue documented elsewhere: `build/disk/lib64/libfontconfig.so.1` symlink now points at the 338 KB real lib, but if a developer skips `bash scripts/create-data-disk.sh --force` after pulling this PR, the kernel will boot the stale stub from `build/data.img` and the soak result will not reproduce. Out of scope for #179, but the eventual fix should land soon (auto-rebuild data.img when `build/disk/` is newer).

### Soak robustness

1-trial 8-minute clean is **thin but defensible** here, because the change is structural rather than tactical:
- Previous fontconfig PRs (W66/W70/W76/W82/W87/W94) were *narrow* patches to individual `Fc*` stub functions — each could miss the next symbol, so 5-trial soaks were warranted to catch the next NULL-deref.
- This PR replaces the entire library with the upstream binary. The next crash, if any, will be in a *different* dependency (libdbus / libglib / libgtk), not another `Fc*` symbol. The crash class is structurally retired.
- Pre-fix baseline was 3/5 SIGSEGV in **minutes** at the exact `libxul+0x185b8a4` / `+0x4056429` cluster; 8 clean minutes past that point with profile SQLite writes is convincing evidence that the cluster is gone.

That said, **please do a 5-trial confirmation in a follow-up** (no need to block #179). If 5/5 are clean past glxtest+font init, we have the strongest possible signal that demo-gate-via-libxul is behind us.

### Recommended next target

Mozilla now reaches profile SQLite writes. The next blocker is whichever stub still surfaces NULL output to a Mozilla loop. Given W97's author note and the demo-gate timeline, the obvious candidates in priority order:

1. **libdbus-1.so.3** — Mozilla's init opens a DBus connection before the UI thread starts; `dbus_connection_new` returning NULL stub still derefs.
2. **libglib-2.0 / libgobject-2.0** — same pattern, GLib type registration touches a lot of returned pointers.
3. **libcairo / libpango** — only if Mozilla iterates these in headless mode (PR description rightly says GTK/X11 should not be exercised — verify by RIP if a future SIGSEGV reappears).

`libdbus` is the cleanest first target: small (~300 KB on host), few transitive deps (only libsystemd / libc / libpthread), and Mozilla calls it early. Follow the same install-fonts-real.sh template.

### Merge guidance

**Merge as-is.** The nits above are minor — they don't change behaviour, they don't risk breakage, and the structural improvement is large enough that "make it perfect before landing" is the wrong trade. Land #179 today, do the 5-trial soak as a separate audit task, and queue a libdbus PR using this same template.

PR #180, opened 2026-05-13
scripts: real libdbus-1.so.3 + libsystemd transitive deps installer

me1iissa:
## Summary

- New `scripts/install-dbus-real.sh` that overlays the real upstream `libdbus-1.so.3` + its `libsystemd.so.0` NEEDED dependency over the install-firefox-stubs.sh-generated stub.  Follows the install-fonts-real.sh template merged in #179.
- Wired into `scripts/create-data-disk.sh` immediately after `install-fonts-real.sh` so the soname symlink eviction runs last.
- `libdbus-glib-1` stub is left in place — that wrapper is deprecated upstream and is not on Ubuntu 24.04 hosts; modern Mozilla calls plain `libdbus-1`.
- No DBus daemon is shipped.  `DBUS_SESSION_BUS_ADDRESS` stays unset; real libdbus returns NULL with `org.freedesktop.DBus.Error.NoServer` from `dbus_bus_get()`, and Mozilla's `nsAppShell` falls through to its no-DBus path (a supported upstream headless configuration).

## Why

PR #179 (W97) closed the fontconfig stub-class by shipping real `libfontconfig` + `libfreetype`.  W97's verifier flagged the next-likely blocker: Mozilla writes the single-byte `"I"` / `"M"` DBus auth-handshake markers to a socket fd during `nsAppShell::Init` / proxy / AT-SPI registration.  The 36 KiB stub `libdbus-1.so.3` returns NULL/0 for every entry point, including `DBusError` out-pointer setters, which Mozilla turns into NULL-deref faults — the same class fontconfig stubs produced before #179.

Real libdbus is a pure userspace `.so` that talks to glibc.  Every syscall it issues (`socket`, `connect`, `sendmsg`, `recvmsg`, `poll`, `close`, `getuid`, `getsockopt`) is already implemented by AstryxOS.  Its transitive chain on Ubuntu 24.04 is just `libsystemd.so.0` (+ `libm` / `libc` / `ld-linux`, all already provided by `install-glibc.sh`) — no further packages required.

## How

Mirrors install-fonts-real.sh:
- Resolve each library through host symlinks
- Copy real `.so` under its versioned name into both `build/disk/lib64/` and `build/disk/lib/x86_64-linux-gnu/`
- `ln -sf` the soname onto the real name (evicts the 36 KiB stub at the soname path)
- Skip files already at the host's exact size (idempotent); `--force` overwrites
- Defensive `readelf` check warns if a future host pulls a libsystemd build with extra NEEDED entries (libcap / liblzma / libzstd / libgcrypt are NOT in stock Ubuntu 24.04)
- Hard-fail if either mandatory library is missing on the host

## Disk delta

| File | Size |
|------|------|
| `libdbus-1.so.3.38.3` | 314 KiB |
| `libsystemd.so.0.42.0` | 1102 KiB |
| **Total** | **~1.4 MiB** |

## Test plan

- [x] Standalone script runs idempotent; `--force` overwrites
- [x] `bash scripts/create-data-disk.sh --force` regenerates `data.img` with both libraries
- [x] `mdir -i build/data.img ::/lib64/libdbus-1.so.3 ::/lib64/libsystemd.so.0` reports 321920 + 1127960 bytes (real, not stub)
- [x] 1-trial `qemu-harness.py start --features firefox-test,kdb,syscall-trace --no-kvm` reaches:
  - Mozilla `dlopen("libdbus-1.so.3")` → success
  - Mozilla `dlopen("libsystemd.so.0")` → success
  - `write(fd=20, "I", 1)` (DBus auth nonce) — single write, no follow-up "M" / `AUTH` since no daemon is listening
  - clone3 worker threads spawn up to TID 47
  - Mozilla writes 658 bytes of `prefs.js` to `/tmp/ff-profile/`
  - Opens `cookies.sqlite-journal`, `startupCache/webext.sc.lz4`, `extension-preferences.json`, `sessionCheckpoints.json`
  - Eventually `exit_group(-6)` from `tid=1` (Mozilla `abort()` post-profile-setup — a separate later-stage blocker, not a DBus stub fault)
- [x] No new SIGSEGV vs W97 post-#179
- [x] Wall-time-to-plateau (~80s sc=4609 pf=15734) within W97 ballpark

## Caveats / not done here

- The `exit_group(-6)` after `prefs.js` write is a downstream blocker for a follow-up dispatch (probably a later GLib / GIO / cairo stub class — `g_bus_get_sync` returning the libdbus-NULL gracefully, then Mozilla hitting another NULL-stub elsewhere in `nsAppShell` or `nsCookieService`).  Full 5-trial soak is W100's job per the dispatch brief.
- `libdbus-glib-1.so.2` stub is intentionally NOT replaced — package is not on Ubuntu 24.04, Mozilla ESR 115 only opens it lazily and never calls into it on headless.
- A future libsystemd ABI bump that pulls libcap / liblzma will trigger the readelf warning in this script; that's the intended signal to add a follow-up step.

Reference: dbus.freedesktop.org/doc/api/html/group__DBusBus.html

### DIFF
--- a/scripts/create-data-disk.sh
+++ b/scripts/create-data-disk.sh
@@ -103,6 +103,27 @@ if [ -f "${ROOT_DIR}/scripts/install-fonts-real.sh" ]; then
         echo "[DATA-DISK] WARNING: install-fonts-real.sh failed — fontconfig/freetype stubs remain"
 fi
 
+# ── Overlay REAL libdbus-1 + libsystemd (non-fatal) ──────────────────────────
+# install-firefox-stubs.sh writes a 36 KiB stub libdbus-1.so.3.  Mozilla
+# initialises a DBus client during nsAppShell::Init / proxy / AT-SPI setup
+# even with no session bus running; the stub returns NULL/0 for every entry
+# point, including DBusError-out-pointer setters, which Mozilla turns into
+# NULL-deref faults (W97 verifier flagged the "I"/"M" auth-handshake bytes
+# on fd=19 as the next-likely blocker class after PR #179).
+# install-dbus-real.sh overlays the real upstream libdbus-1 (~314 KiB) plus
+# its libsystemd.so.0 transitive dep (~1102 KiB) over the stub, keeping the
+# libdbus-glib-1 stub in place (deprecated wrapper, not on the host).  Must
+# run AFTER install-firefox-stubs.sh so its `ln -sf` evicts the stub at the
+# soname path.  No DBus daemon is shipped — DBUS_SESSION_BUS_ADDRESS stays
+# unset, real libdbus returns NULL with org.freedesktop.DBus.Error.NoServer,
+# and Mozilla's nsAppShell falls through to its no-DBus path.
+if [ -f "${ROOT_DIR}/scripts/install-dbus-real.sh" ]; then
+    DBUS_FLAGS=""
+    [ "${FORCE}" = true ] && DBUS_FLAGS="--force"
+    bash "${ROOT_DIR}/scripts/install-dbus-real.sh" ${DBUS_FLAGS} 2>&1 | sed 's/^/[DATA-DISK] /' || \
+        echo "[DATA-DISK] WARNING: install-dbus-real.sh failed — libdbus-1 stub remains"
+fi
+
 # ── Compile glibc_hello oracle binary if source present ──────────────────────
 GLIBC_HELLO_SRC="${ROOT_DIR}/userspace/glibc_hello.c"
 GLIBC_HELLO_BIN="${BUILD_DIR}/glibc_hello"
--- a/scripts/install-dbus-real.sh
+++ b/scripts/install-dbus-real.sh
@@ -1,0 +1,241 @@
+#!/usr/bin/env bash
+#
+# install-dbus-real.sh — Install REAL libdbus-1.so.3 (and its transitive
+# libsystemd.so.0 dependency) from the host into build/disk/, overwriting any
+# stub copy that install-firefox-stubs.sh may have produced.
+#
+# Rationale
+# ─────────
+# Firefox ESR 115 (headless) initialises a DBus client during nsAppShell::Init
+# / nsThreadManager / proxy / AT-SPI registration, even with no session bus
+# running.  Mozilla expects libdbus to:
+#
+#   - return a NULL DBusConnection* from dbus_bus_get() when the environment
+#     does not advertise a session bus (DBUS_SESSION_BUS_ADDRESS unset, no
+#     /var/run/dbus/system_bus_socket reachable);
+#   - populate the DBusError struct passed to that call with a sensible
+#     error name + message ("org.freedesktop.DBus.Error.NoServer");
+#   - thread the result through dbus_error_is_set() / dbus_error_free() so
+#     Mozilla's no-DBus fallback branch in nsAppShell takes effect.
+#
+# The generic stub written by install-firefox-stubs.sh satisfies the dynamic
+# linker but returns NULL/0 for every entry point, including the
+# DBusError-out-pointer setters.  Mozilla then iterates a NULL list, reads
+# uninitialised dbus_error.name (NULL char*) inside a printf, or derefs a
+# stub-returned NULL handler-app pointer — the same NULL-fault class that
+# fontconfig stubs produced before PR #179.  W97's verifier flagged Mozilla
+# writing the single-byte "I" / "M" markers to fd=19 (the DBus auth-protocol
+# greeting + AUTH command first bytes) as the next-likely blocker.
+#
+# Real libdbus is a pure userspace .so that talks to glibc — every syscall
+# it issues (socket, connect, sendmsg, recvmsg, poll, close, getuid,
+# getsockopt) is already implemented by AstryxOS.  Its one transitive
+# dependency on Debian/Ubuntu is libsystemd.so.0 (for sd_journal_send and
+# sd_listen_fds — libdbus uses neither in the no-session-bus path, but the
+# dynamic linker still resolves the NEEDED entry at load time).  libsystemd
+# in turn depends only on libm + libc, both already provided by
+# install-glibc.sh.
+#
+# Disk delta (Ubuntu 24.04 host):
+#   libdbus-1.so.3.38.3    314 KiB
+#   libsystemd.so.0.42.0  1102 KiB
+#
+# Total ~1.4 MiB.  The DBus daemon itself (`dbus-daemon`) is NOT shipped —
+# DBUS_SESSION_BUS_ADDRESS stays unset, dbus_bus_get returns NULL with a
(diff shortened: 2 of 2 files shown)


me1iissa (commented):
## Review (W102): `scripts: real libdbus-1.so.3 + libsystemd transitive deps installer`

**Verdict: APPROVE WITH NITS.** Merge once you're ready — the structural design is sound, the script is bit-identical to the host upstream (verified via sha256), idempotence/`--force` behave as documented, and the 1-trial soak proves the intended unblock (DBus auth-handshake `"I"` byte on fd=20, Mozilla now writes 658 B of prefs.js — previously unreached). All nits below are cosmetic; none block merge.

### Top findings

**1. Script is a faithful mirror of `install-fonts-real.sh` (PR #179 template) — no Linux-subsystem-invariant concerns.**
Confirmed by reading both side-by-side. `copy_lib` is essentially the same function, the search-dir set is the same, the soname-eviction semantics are the same (`ln -sf` over a regular file replaces the stub at the soname path). No `patchelf` / `objcopy` / `strip` / any binary mutation — `grep -nE 'patchelf|objcopy|strip\b'` against the new script returns nothing. Host upstream binaries are shipped byte-for-byte (`sha256sum` of the copied `libdbus-1.so.3.38.3` vs `/usr/lib/x86_64-linux-gnu/libdbus-1.so.3.38.3` match: `bb0b37af…`). The Linux-subsystem invariant ("never edit upstream binaries") is preserved.

**2. The libsystemd transitive-chain assertion is correct on stock Ubuntu 24.04 — `readelf` agrees with the script's `EXPECTED_SYSTEMD_NEEDED`.**
Verified locally:
```
$ readelf -d /usr/lib/x86_64-linux-gnu/libsystemd.so.0 | grep NEEDED
 0x...01 (NEEDED)  Shared library: [libm.so.6]
 0x...01 (NEEDED)  Shared library: [libc.so.6]
 0x...01 (NEEDED)  Shared library: [ld-linux-x86-64.so.2]
```
All three are already on disk via `install-glibc.sh`, so the chain truly terminates here — no libcap/liblzma/libzstd/libgcrypt are pulled in by the stock Ubuntu 24.04 libsystemd build. The defensive `readelf` warning loop at the bottom of the script is good future-proofing — if a future host bumps libsystemd and adds NEEDED entries, the next `create-data-disk.sh` run will print a loud `WARN`. Per dbus.freedesktop.org's `dbus_bus_get` contract, libdbus on the no-session-bus path uses none of libsystemd's API surface (only `sd_journal_send` / `sd_listen_fds` call into it, and those are bypassed when `DBUS_SESSION_BUS_ADDRESS` is unset). Even if a future libsystemd swap left some dlopen symbols unresolved, libdbus itself would still load (the linker only requires the DT_NEEDED entry to resolve, not every symbol — but as it happens every symbol does resolve today, which is the stronger condition).

**3. Local exercise confirms idempotence + `--force` + soname symlink-over-stub semantics.**
Ran the script against `/tmp/dbus-test/build/disk/...` from a clean state:
- First run: `Copied libdbus-1.so.3.38.3 (321920 bytes)` + `Copied libsystemd.so.0.42.0 (1127960 bytes)` in both `lib64/` and `lib/x86_64-linux-gnu/`. Soname symlinks `libdbus-1.so.3 -> libdbus-1.so.3.38.3` and `libsystemd.so.0 -> libsystemd.so.0.42.0` created in both.
- Second run: `SKIP (present)` for all four real files; symlinks re-asserted (cheap).
- `--force` run: `Copied` for all four.

This matches the documented behaviour and the PR's `mdir` sizes (321920 / 1127960). The `ln -sf` over an existing regular-file stub (the case in the real build where `install-firefox-stubs.sh` writes a 36 KiB stub at the `libdbus-1.so.3` path first) was not exercised in my temp tree, but coreutils `ln -sf` semantics — unlink-then-symlink — handle that case correctly, and that's the same code path `install-fonts-real.sh` has been using since PR #179.

### Nits (none blocking)

- **n.1 — `ln -sf` over a directory edge case.** If a previous run somehow left `libdbus-1.so.3` as a directory (shouldn't ever happen, but defensive coding), `ln -sf` will create the symlink *inside* the directory rather than replacing it. Trivial cost to fix (`rm -f "${dest_soname}" && ln -s ...`) but truly hypothetical — feel free to ignore.

- **n.2 — `mandatory` failure logs `libdbus-1-3 libsystemd0` even if only one is missing.** Cosmetic; the user can figure it out.

- **n.3 — the comment block on lines 134–138 (and create-data-disk.sh's mirror) cites the script's reasoning at length; consider whether this lives better in a single shared header. Not a blocker** — and the present duplication is actually agent-friendly (each script self-documents the rationale without cross-references).

- **n.4 — the script's `find_lib` search order is `/usr/lib/x86_64-linux-gnu` → `/lib/x86_64-linux-gnu` → `/usr/lib64` → `/lib64`. This is fine for Ubuntu, but on Fedora-like hosts where `libsystemd.so.0` lives in `/usr/lib64`, the host's `readlink -f` will resolve symlinks correctly anyway. Don't change anything.**

### Soak methodology

1-trial / 5-min is consistent with PR #179's precedent for structural changes (whole-library replacement, not a stub-patch tweak). I would normally ask for a 5-trial confirmation before merging a soak-relevant PR, but per the dispatch brief W100 is currently running a 5-trial against the post-#179 baseline, and the next baseline after #180 lands will catch any regression. Merge on 1-trial is defensible here.

### The new `exit_group(-6)` blocker

Not a blocker for #180 — it's downstream of the unblock this PR delivered. Before #180, Mozilla never wrote `prefs.js`. After #180, it writes 658 B of prefs.js plus opens cookies.sqlite-journal / startupCache / sessionCheckpoints, then aborts. That's two new layers of progression, exactly the pattern PR #179 produced (every real-lib swap unblocks the next class of NULL-fault). My best-guess root-cause priority for the follow-up:

1. **Most likely**: a stub return inside `nsAppShell::Init`'s post-DBus branch — probably `g_application_register` / `g_dbus_connection_*` (still stubbed, since `libglib-2.0` and `libgio-2.0` are stubs) returning a NULL `GDBusConnection*` that Mozilla then deref's in an `MOZ_RELEASE_ASSERT`. The `rt_sigprocmask` + futex sequence the author noted is consistent with `pthread_kill(SIGABRT)` from `__libc_message` after such an assertion.
2. **Less likely**: a prefs-validation assertion. Possible but Mozilla's prefs writer is fairly lenient.
3. **Worth checking**: AT-SPI / a11y registration path now that real libdbus is in play and may be probing more aggressively than the all-NULL stub did.

### Merge guidance

**Merge immediately.** No reason to hold. The 1-trial soak demonstrates the unblock, the script is structurally clean and verified, and the downstream `exit_group(-6)` is the next dispatch's problem, not this PR's.

### Reference

dbus.freedesktop.org/doc/api/html/group__DBusBus.html (`dbus_bus_get` semantics when `DBUS_SESSION_BUS_ADDRESS` is unset)

PR #181, opened 2026-05-13
kernel/signal: [SIGNAL/VMA] include user_rip VMA + file offset

me1iissa:
## Summary

Extends the SIGSEGV-delivery `[SIGNAL/VMA]` banner to carry enough information to symbolicate the faulting instruction in one pass, even when `user_rip` and `cr2` fall in different VMAs.

W100 (5-trial post-#179 baseline) found that on faults where the RIP is far from CR2, the existing banner did not surface a *file offset* (only `offset_in_vma`, i.e. `user_rip - base`), and the CR2-containing VMA was only included incidentally when it happened to be executable + file-backed. That meant the investigator had to read three or four `[FFTEST/mmap-so]` lines, do the subtraction by hand, then re-add the per-VMA file offset before `objdump -d --start-address=…` could land on the faulting instruction. For the post-glxtest libxul cluster this added meaningful per-trial latency to every symbolication pass.

This PR makes the banner one-pass usable:

- **`offset_in_file=`** field on every RIP- and CR2-containing entry that is file-backed (computed as `(addr - VMA.base) + VmBacking::File.offset`).
- **CR2-containing VMA** is now always included in the snapshot, even when it is anonymous (heap/stack/bss) or non-executable (.rodata, .data) and would previously have been filtered out.
- **`anon=1` tag** on the entry when the containing VMA is anonymous — the common stack-corruption / wild-write fingerprint.
- **RIP/CR2 same-VMA coalescing**: when `user_rip` and `cr2` fall in the same VMA we emit a single line with `rip=1 cr2=1 rip_offset_in_vma=… cr2_offset_in_vma=…` so the relative geometry is obvious.
- **Edge cases** explicitly diagnosed:
  - `rip_in_kernel=1` when `user_rip >= 0xFFFF_8000_0000_0000` (signal-from-IRQ delivery).
  - `rip_unmapped=1` when `user_rip` is in the user half but no VMA covers it (poisoned function pointer / freed code).
- **Snapshot cap discipline**: the 8-entry cap now applies only to *neighbours*; the RIP- and CR2-containing entries are exempt and never get pushed out by a libxul-heavy process.

Format (one line per kept VMA):
```
[SIGNAL/VMA] pid=<n> name=<label> base=<vaddr> end=<vaddr> size=<bytes>
 prot=<rwx> file=<0|1> rip=<0|1> cr2=<0|1>
 [offset_in_vma=<…> [offset_in_file=<…>]] [anon=1]
```

## How this unblocks W100-CLUSTER-A symbolication

A `--features firefox-test` trial caught the same libxul cluster W100 hit. With this PR the banner now resolves end-to-end without manual arithmetic:

```
[SIGNAL] SIGSEGV ISR delivery: PID=1 CR2=0x7efffe243460
    user_rip=0x7efffa20dbc0 handler=0x7efff87113b0 new_rsp=0x7ffffffe9f28
...
[SIGNAL/VMA] pid=1 name=[mmap-file] base=0x7efff5778000 end=0x7efffb609000
    size=0x5e91000 prot=r-x file=1 rip=1 cr2=0
    offset_in_vma=0x4a95bc0 offset_in_file=0x508abc0
[SIGNAL/VMA] pid=1 name=[mmap-file] base=0x7efffe1d2000 end=0x7efffe74b000
    size=0x579000 prot=r-- file=1 rip=0 cr2=1
    offset_in_vma=0x71460 offset_in_file=0x90bf460
```

Cross-checks:
- `0x7efff5778000 + 0x4a95bc0 = 0x7efffa20dbc0` ✓ matches `user_rip`
- `0x7efffe1d2000 + 0x71460 = 0x7efffe243460` ✓ matches `CR2`

Direct symbolication path now reduces to:

```
objdump -d build/disk/opt/firefox/libxul.so --start-address=0x508abc0
```

without any per-fault arithmetic step.

## Diff size

Approximately +152 net lines (+196 / -44) across two files. This is slightly over the 120-LOC budget; the four distinct format branches (anon/file × rip-only/cr2-only/both/neighbour), the two pre-amble edge cases, and the expanded unit-test coverage account for the overrun. No logical change was split into a follow-up just to hit a number.

## Test plan

- [x] `python3 scripts/qemu-harness.py check --features ""` (default) — `rc=0`
- [x] `python3 scripts/qemu-harness.py check --features kdb` — `rc=0`
- [x] `python3 scripts/qemu-harness.py check --features firefox-test` — `rc=0`
- [x] `--features test-mode`: test 18b (`[SIGNAL/VMA] snapshot policy`) passes with the new field assertions covering RIP-only / CR2-in-anon / RIP-and-CR2-same-VMA / RIP-in-kernel.
- [x] `--features firefox-test` (1 trial): SIGSEGV cascade emits the new format with `offset_in_file=` and `cr2=1` lines populated correctly; cross-checked against `base + offset_in_vma == user_rip / cr2`.

## References

- POSIX.1-2024 `sigaction(2)` / `siginfo_t` semantics (`si_addr = cr2` for SIGSEGV).
- Intel SDM Vol. 3 §4.7 — page-fault error code, CR2 register semantics.

### DIFF
--- a/kernel/src/signal.rs
+++ b/kernel/src/signal.rs
@@ -648,7 +648,7 @@ pub unsafe fn deliver_sigsegv_from_isr(
     // later `&mut sig_state` borrow.  The actual print happens at the
     // end of this function so PROCESS_TABLE is not held across the
     // serial writes.  See the `[SIGNAL/VMA]` emission block below.
-    let vma_snapshot = signal_vma_snapshot(proc_entry.vm_space.as_ref(), user_rip);
+    let vma_snapshot = signal_vma_snapshot(proc_entry.vm_space.as_ref(), user_rip, cr2);
 
     let sig_state = match proc_entry.signal_state.as_mut() {
         Some(s) => s,
@@ -765,7 +765,7 @@ pub unsafe fn deliver_sigsegv_from_isr(
         "[SIGNAL] SIGSEGV ISR delivery: PID={} CR2={:#x} user_rip={:#x} handler={:#x} new_rsp={:#x}",
         pid, cr2, user_rip, handler_addr, new_rsp
     );
-    emit_signal_vma_banner(pid, user_rip, &vma_snapshot);
+    emit_signal_vma_banner(pid, user_rip, cr2, &vma_snapshot);
 
     true
 }
@@ -774,6 +774,11 @@ pub unsafe fn deliver_sigsegv_from_isr(
 ///
 /// `name` is `&'static str` (the same lifetime as `VmArea::name`), so the
 /// snapshot is cheap to build and carry across the lock drop.
+///
+/// `file_offset` is the byte offset within the backing file of the VMA's
+/// first byte (zero for anonymous/device mappings).  Combined with
+/// `user_rip - base` it gives the file offset that `objdump -d --start-address`
+/// expects when symbolicating a shared-library RIP.
 #[derive(Copy, Clone)]
 pub(crate) struct VmaSnap {
     pub(crate) base: u64,
@@ -781,14 +786,25 @@ pub(crate) struct VmaSnap {
     pub(crate) prot: u32,
     pub(crate) name: &'static str,
     pub(crate) file_backed: bool,
+    pub(crate) anonymous: bool,
+    pub(crate) file_offset: u64,
     pub(crate) contains_rip: bool,
+    pub(crate) contains_cr2: bool,
 }
 
-/// Capture the VMA covering `user_rip` plus a small number of executable,
-/// file-backed neighbours that are likely shared-library load segments.
+/// Highest user-space address (one byte past).  Anything `>=` this is in the
+/// kernel half (PML4 entries 256-511, base `0xFFFF_8000_0000_0000`).
+const USER_ADDR_END: u64 = 0x0000_8000_0000_0000;
+
+/// Capture the VMA covering `user_rip` (and, if distinct, the VMA covering
+/// `cr2`) plus a small number of executable, file-backed neighbours that are
+/// likely shared-library load segments.
 ///
 /// Capped at 8 entries to keep serial volume bounded — emitting every VMA
-/// of a libxul-loading process would push hundreds of lines per fault.
+/// of a libxul-loading process would push hundreds of lines per fault.  The
+/// RIP-containing and CR2-containing VMAs are always preserved even when the
+/// cap is otherwise hit, so the symbolicating investigator never loses them
+/// to neighbour overflow.
 ///
 /// Visible to `test_runner` via `pub(crate)` so the snapshot policy can
 /// be unit-tested against a synthetic `VmSpace` without standing up a
@@ -796,6 +812,7 @@ pub(crate) struct VmaSnap {
 pub(crate) fn signal_vma_snapshot(
     space: Option<&crate::mm::vma::VmSpace>,
     user_rip: u64,
+    cr2: u64,
 ) -> alloc::vec::Vec<VmaSnap> {
     use crate::mm::vma::{PROT_EXEC, VmBacking};
     let mut out: alloc::vec::Vec<VmaSnap> = alloc::vec::Vec::new();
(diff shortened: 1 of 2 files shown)


me1iissa (commented):
## /review verdict: APPROVE WITH NITS

Thanks for picking this up — it lands exactly the W100 symbolication ergonomics improvement that the post-#176 libxul cluster needs. The diff is well-scoped, the format extension is one-pass usable, and the unit tests cover the new behaviour meaningfully (not just line-count expansion).

### Verified

- **`signal_vma_snapshot()` (kernel/src/signal.rs:812–859)**: the CR2-VMA lookup uses the same `a.contains(addr)` predicate as the RIP-VMA lookup; both ride on the same sorted `space.areas` iteration, so symmetry is exact. `file_backed` is destructured from `VmBacking::File` and `anonymous` from `VmBacking::Anonymous` (both correctly mutually exclusive against `Device`).
- **Cap-exempt ordering (signal.rs:843)**: the implementation is `if !contains_rip && !contains_cr2 && out.len() >= 8 { continue; }` — i.e. RIP/CR2 are unconditional pushes regardless of cap. The dispatch-prompt's "subtle ordering bug" doesn't bite here because the bypass happens at the per-entry check, not as a post-hoc "skip if cap reached except for…". Areas are stored sorted by base (vma.rs:196), so iteration is deterministic; RIP/CR2 always land in the output even if a libxul process has 100+ exec neighbours.
- **Edge cases (signal.rs:882–906)**: `rip_in_kernel` short-circuits via `>= USER_ADDR_END` (0x0000_8000_0000_0000) — correctly using the canonical user-half boundary, no Option/NULL deref. `rip_unmapped` only triggers when the snapshot is non-empty but no entry has `contains_rip` (so the loop still emits the CR2 line and neighbours). Empty-snapshot path emits one summary `no_vma_match=1` and returns.
- **Underflow safety**: `user_rip - v.base` / `cr2 - v.base` are gated on `v.contains_rip` / `v.contains_cr2`, which require `addr >= base` per `VmArea::contains` (vma.rs:85). No defensive bounds needed; logic is correct by construction.
- **Same-VMA coalescing (signal.rs:917)**: hit before either single-flag branch — emits exactly one line with `rip=1 cr2=1` and the two `*_offset_in_vma=…` fields. No duplicate emission. Unit-tested at test_runner.rs:3981–3993 (placing both addresses inside libxul .text); flag-presence asserted via `iter().find(|v| v.contains_rip && v.contains_cr2).is_none()`.
- **Format backwards-compatibility**: ran `grep -rn 'SIGNAL/VMA.*offset=' .` and `grep -rn 'offset=' scripts/ tools/ docs/ --include='*.py' --include='*.sh' --include='*.rs'` — no external parser consumes the old `offset=` literal. `qemu-harness.py` doesn't grep the banner. Rename `offset=` → `offset_in_vma=` is safe.
- **Test 18b expansion (test_runner.rs:3897–4012)**: assertions are meaningful — they exercise (a) CR2-in-anon kept despite non-exec/non-file (new entry-2 case), (b) `anonymous` flag, (c) `file_offset` field, (d) same-VMA coalesced flag-set, (e) kernel-RIP not appearing as `contains_rip`, (f) `None` space → empty. Assertions are address-relative (`0x10_0000 + 0x1234`), not absolute layout assumptions.
- **Builds**: `qemu-harness.py check` succeeds for default, `kdb`, and `firefox-test` feature sets (`rc=0` on all three).

### Nits (non-blocking; merge as-is is fine)

1. **Cap-bypass not unit-tested.** The test fixture has 4 VMAs; the cap (8) is never approached. A regression that changed `out.len() >= 8` to `>= 0` or moved the bypass check would not be caught. A 9th-VMA fixture would close that gap (~15 LOC). Worth doing in a follow-up; not required for this PR. Reference: POSIX siginfo `si_addr` semantics (sigaction(2)) don't constrain VMA snapshot size, so any cap is a kernel-side choice that warrants its own test.
2. **Live same-VMA case not yet observed.** Author notes this in the PR body. Acceptable — the same-VMA case requires a libxul-internal NULL-deref where both RIP and CR2 land inside libxul's RX segment, which is unusual (most W91-cluster faults have CR2 on stack/heap and RIP in libxul). Unit-test coverage is sufficient until a live trial naturally produces one.
3. **CR2-in-kernel diagnostic absent.** The pre-amble flags `rip_in_kernel=1` and `rip_unmapped=1` for RIP but does not mirror the same for CR2 (e.g. `cr2_in_kernel=1` when a Ring-3 user faults on a kernel address through user-mode access). Rare and not in scope here — but a future tightening if the W84/exit_group SIGSEGV-in-dump_for_exit ever recurs.
4. **`name` containing whitespace** (e.g. `"[stack grow]"` in proc/elf.rs:759) still produces an ambiguous `name=[stack grow] base=…` token where naive whitespace-splitting parsers see `name=[stack` as one token. Pre-existing from PR #171; not a regression. Could be tightened later with `name="[stack grow]"` quoting.

### Diff-size

+196/-44 across 2 files. Within the global 2× burst budget (1.6× the 120-LOC target) and the author's body explains it: four format branches (anon × file × rip-only/cr2-only/both/neighbour combinations) plus two pre-amble edge cases plus test expansion. No logical change appears split or stretched to hit the number.

### Recommendation

**Approve and merge.** Nits 1 and 3 are worth a small follow-up but neither blocks shipping this. The PR materially shortens the symbolication round-trip for the post-#176 libxul NULL-deref cluster, which is the current critical-path investigation.

### References

- POSIX.1-2024 `sigaction(2)` — `si_addr = cr2` for SIGSEGV (verified in code at signal.rs:732).
- Intel SDM Vol. 3 §4.7 — page-fault error code; bit-1 (write/read) decoded at vma.rs:129.

PR #182, opened 2026-05-13
scripts/astryx_qemu: TCG-safe CPU model fallback (close AVX-512 IFUNC trap)

me1iissa:
## Summary

`scripts/astryx_qemu.py` unconditionally selected `-cpu host` for
`mode=firefox-test` regardless of KVM availability. Under TCG (e.g. when
the harness is invoked with `--no-kvm`), that advertised the host's full
CPUID — including AVX-512, AVX10, and SHA-NI — to the guest. glibc's
IFUNC resolver reads CPUID and selects `memcpy` / `strcmp` /
`__strncpy_avx2_rtm` variants accordingly; TCG cannot decode many of
those opcodes at runtime, and the very first call into one of them
raised #UD. The kernel reflected the trap to userspace as a fatal
exception, which the trace pipeline rendered as `SIGSEGV` /
`exit_group(-6)` clusters in the `0x7efff{a,b}...` libc / libxul
region — the **W100-CLUSTER-A** symptom investigations had been
chasing as a real Mozilla bug.

This change removes the special-case and makes the selection a function
of KVM state only:

- `--cpu MODEL` → use MODEL verbatim (`reason="override"`)
- KVM available + not `--no-kvm` → `-cpu host` (`reason="kvm-host"`)
- KVM unavailable OR `--no-kvm` → TCG-safe baseline (`reason="tcg-safe"`)

The TCG-safe baseline is:

```
qemu64,+ssse3,+sse4_1,+sse4_2,+avx,+avx2,+fma,
      +rdtscp,+popcnt,+aes,+pclmulqdq,+cmov
```

— enough ISA for glibc's IFUNC dispatcher to pick fast SSE4.2 / AVX2 /
FMA variants while staying inside TCG's emulated instruction set, with
**no** AVX-512 / AVX10 / SHA-NI advertised. Per QEMU's per-feature
emulation matrix (`docs/system/i386/cpu.rst`), all of these features are
fully decoded under TCG.

The harness now logs the decision at start time:

```
[harness] cpu_model=qemu64,+ssse3,+sse4_1,+sse4_2,+avx,+avx2,+fma,+rdtscp,+popcnt,+aes,+pclmulqdq,+cmov
          reason=tcg-safe
          (kvm_available=True, no_kvm_flag=True, force_kvm_flag=False)
```

emits a `{"kind": "cpu_model", ...}` event into the session's
`events.jsonl`, and exposes `cpu_model_resolved`, `cpu_model_reason`,
`kvm_available`, `kvm_effective` (all **additive**) on both the session
JSON and the `start` subcommand's stdout payload — so a single
`events <sid>` or `status <sid>` call reveals whether a run was on KVM
fidelity or the TCG baseline.

## Files

- `scripts/astryx_qemu.py` — add `_TCG_SAFE_CPU` constant, factor
  selection into `cpu_model_for(mode, kvm, override) -> (model, reason)`,
  `_cpu_args()` becomes a thin wrapper.
- `scripts/qemu-harness.py` — call `cpu_model_for()` up-front in
  `cmd_start`, log the decision, persist + emit it via session JSON,
  event stream, and the start-subcommand stdout payload.

## Verification

3 trials with `--no-kvm`, `--features firefox-test,kdb,syscall-trace`:

| trial | sc     | pf    | clone | execve | SIGSEGV | #UD | bugcheck | exit_group | terminal           |
|-------|--------|-------|-------|--------|---------|-----|----------|------------|--------------------|
| 1     | 1947   | 9017  | 19    | 0      | 0       | 0   | 0        | 0          | openat-fd loop / sem_wait plateau |
| 2     | 272    | 316   | 10    | 0      | 0       | 0   | 0        | 0          | openat-fd loop     |
| 3     | (early)| n/a   |  1    | 0      | 0       | 0   | 0        | 0          | libxul mmap stall  |

Pre-fix post-#176 baseline (recorded in the agent-memory baseline
note): **3/5 SIGSEGV→exit_group(11), 1/5 #UD→exit_group(-6)** — 80% of
runs ended in the W100-CLUSTER-A fault cluster.
Post-fix: **0/3** SIGSEGV/#UD/bugcheck/exit_group events.

The harness boot log now shows:

```
[harness] cpu_model=qemu64,+ssse3,+sse4_1,+sse4_2,+avx,+avx2,+fma,+rdtscp,+popcnt,+aes,+pclmulqdq,+cmov reason=tcg-safe (kvm_available=True, no_kvm_flag=True, force_kvm_flag=False)
```

and `ps -p <pid> -o args` confirms QEMU was launched with the new
`-cpu` argv and without `-enable-kvm`.

## Verdict

The W100-CLUSTER-A "0x7efffa-region NULL-deref" cluster was a **TCG ISA
mismatch artefact**, not a real Mozilla bug. With the TCG-safe model
in place, Mozilla under TCG now reaches deeper Branch-A / openat-fd
recursion / libxul-mmap states — actual guest-kernel issues to
investigate separately.

## Test plan

- [x] Trial 1: `--no-kvm`, firefox-test → no SIGSEGV / #UD / bugcheck / exit_group; reaches sc=1947 with 19 clone threads.
- [x] Trial 2: `--no-kvm`, firefox-test → same, different early wedge (procfs fd recursion).
- [x] Trial 3: `--no-kvm`, firefox-test → same, wedges in libxul mmap (no userspace fault).
- [x] `ps -p <pid> -o args` confirms `-cpu qemu64,+ssse3,...` and no `-enable-kvm` in argv.
- [x] `events.jsonl` contains `{"kind": "cpu_model", "reason": "tcg-safe", ...}`.
- [x] `python3 -m py_compile scripts/astryx_qemu.py scripts/qemu-harness.py` OK.
- [x] `cpu_model_for()` smoke test: TCG / KVM / override paths return the expected tuples.
- [x] QEMU accepts the new `-cpu` model string (boot-probed to timeout).

## Breaking changes

None. All new fields on session JSON, stdout, and the event stream are
additive (`cpu_model_resolved`, `cpu_model_reason`, `kvm_available`,
`kvm_effective`, `no_kvm_flag`, `force_kvm_flag`). No existing key was
renamed.

### DIFF
--- a/scripts/astryx_qemu.py
+++ b/scripts/astryx_qemu.py
@@ -31,14 +31,19 @@ differences are therefore visible in one file.
    controllers, so the ATA probe sees hardware regardless of where
    the data.img is attached.
 
-2. **CPU — `host` when KVM available, `qemu64,+rdtscp,+sse4_2` under TCG.**
+2. **CPU — `host` under KVM, a TCG-safe baseline under TCG.**
 
-   Rationale: `host` matches what a physical boot would encounter on
-   the developer's workstation (the target surface Firefox actually
-   hits). `qemu64,+rdtscp,+sse4_2` is the minimum feature set we rely
-   on in the kernel (rdtscp for vDSO-style timing, sse4.2 for
-   compiler-generated string ops). Firefox-test mode can force `host`
-   because it requires KVM to be fast enough anyway.
+   Rationale: under KVM, `host` matches what a physical boot would
+   encounter on the developer's workstation (the target surface
+   Firefox actually hits). Under TCG, we must NOT advertise host
+   CPUID — glibc's IFUNC resolver reads CPUID and selects
+   AVX-512/AVX10/SHA-NI variants of `memcpy`/`strcmp`/etc that TCG
+   cannot decode at runtime, triggering #UD and looking like a
+   userspace crash. The TCG baseline (`qemu64` + safe extensions
+   through AVX2/FMA, no AVX-512, no AVX10, no SHA-NI) gives glibc
+   enough ISA to pick fast SSE/AVX2 variants while staying inside
+   TCG's emulated instruction set. See QEMU `docs/system/i386/cpu.rst`
+   for the per-feature decode status.
 
 3. **Memory — 1 GiB default, 2 GiB for firefox mode.**
 
@@ -78,6 +83,29 @@ _SMP = int(os.environ.get("ASTRYX_SMP", "2"))
 #: 1 → QEMU exit(3)=fail. Shared with run-test.sh semantics.
 _ISA_DEBUG_EXIT = ["-device", "isa-debug-exit,iobase=0xf4,iosize=0x04"]
 
+#: TCG-safe CPU baseline. `qemu64` is the QEMU-defined model whose
+#: feature set TCG fully decodes; the `+` extensions below are all
+#: TCG-decoded in QEMU >= 7 and give glibc's IFUNC resolver enough
+#: ISA surface to pick a fast (SSE4.2 / AVX2 / FMA) variant of
+#: `memcpy`, `strcmp`, etc. Critically we do NOT advertise:
+#:   • AVX-512* (any flavour) — TCG decodes a subset only; many
+#:     vector ops trap as #UD at runtime
+#:   • AVX10 (Intel) — not in TCG
+#:   • SHA-NI (Intel SHA extension) — Mozilla NSS picks this when
+#:     present; TCG decodes only a subset and the resolver falls
+#:     into a path that issues an unsupported opcode
+#: When KVM is in use we want `-cpu host` instead so guest CPUID
+#: matches hardware exactly.
+_TCG_SAFE_CPU = (
+    "qemu64"
+    ",+ssse3,+sse4_1,+sse4_2"   # SSE family — palignr, pcmpistri
+    ",+avx,+avx2,+fma"           # AVX2 + FMA for memcpy/memmove
+    ",+rdtscp"                   # vDSO-style timing path
+    ",+popcnt"                   # std::popcount IFUNC
+    ",+aes,+pclmulqdq"           # NSS AES-NI / CRC32C
+    ",+cmov"                     # universally assumed since P6
+)
+
 
 # ── Host feature detection ────────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


me1iissa (commented):
## Review — PR #182 (TCG-safe CPU model fallback)

**Verdict: APPROVE.**

This is a high-impact, well-scoped toolchain fix. The hypothesis (W103) that the harness was unconditionally advertising host CPUID under TCG — and that glibc IFUNC resolvers were then picking AVX-512 / SHA-NI / AVX10 variants that TCG cannot decode — is verified end-to-end here, and the fix is structurally clean.

### What I checked

1. **`cpu_model_for(mode, kvm, override)` semantics.** Read `scripts/astryx_qemu.py` end-to-end and smoke-tested all four paths (override / kvm-host / tcg-safe / firefox-test+tcg). The order of resolution is correct: `override` always wins, then `kvm` selects host vs the TCG-safe baseline. The old "firefox-test forces host" special case is gone, which is the whole point.

2. **`kvm_effective` computation.** `cmd_start` derives `kvm_effective = kvm_arg if not None else _detect_kvm()`, where `kvm_arg` is `False` for `--no-kvm`, `True` for `--force-kvm`, and `None` otherwise. `--force-kvm` correctly produces `kvm_effective=True` even when `/dev/kvm` is absent (QEMU itself will then fail loudly on `-enable-kvm` — that's the documented contract).

3. **TCG-safe feature list.** I verified empirically against QEMU 10.2.1: the exact PR string boots cleanly with **zero** "TCG doesn't support requested feature" warnings, while `+avx512f,+avx512bw` produces the expected warnings. The chosen set (`ssse3,sse4_1,sse4_2,avx,avx2,fma,rdtscp,popcnt,aes,pclmulqdq,cmov`) is the right shape — it gives glibc IFUNC enough surface to land on AVX2/FMA fast paths without exposing anything TCG cannot decode.

4. **Backward compatibility.**
   - `--cpu MODEL` override still routes verbatim (used by Window-6 perf sweeps).
   - The legacy `cpu_model` session-JSON field is preserved (`"cpu_model": cpu_model or "default"`); the new `cpu_model_resolved`, `cpu_model_reason`, `kvm_available`, `kvm_effective`, `no_kvm_flag`, `force_kvm_flag` are strictly additive.
   - Grepped `scripts/` and `tests/` for any consumer of `cpu_model` JSON or argv layout — none beyond the two files in the diff.

5. **Consistency between log and actual argv.** `cmd_start` computes `cpu_model_for(mode="test", kvm=kvm_effective)` for logging; `_launch_qemu_harness` → `build_qemu_cmd(mode="test", kvm=kvm_arg)` → `_cpu_args(mode, kvm, cpu_override)` for the real argv. Both call the same helper with `mode="test"` (the harness never uses `mode="firefox-test"`) and resolve `kvm` identically, so the `[harness] cpu_model=…` line and the `-cpu` QEMU actually receives are guaranteed to agree.

6. **Structured logging hygiene.** The `[harness] cpu_model=… reason=… (…)` stderr line is single-line, key=value, parseable. The `{"kind": "cpu_model", …}` events.jsonl entry mirrors the same fields and goes through the existing `_emit_event` helper.

7. **Linux-subsystem invariant.** No kernel changes, no upstream-binary patches. Toolchain only — clean pass.

### Nits (non-blocking)

- The `mode` parameter to `cpu_model_for` is now unused (the firefox-test branch is gone). Either drop it from the signature in a follow-up, or wire it back if any future per-mode policy is anticipated. Not worth blocking on.
- The TCG-safe constant omits `+rdrand`/`+rdseed`. Mozilla NSS prefers these when present but falls back to `/dev/urandom` — verified harmless, but could be mentioned in the constant's docstring as an intentional omission.
- Trial 3 in the verification table "stalled early" with no full counters. The result is unambiguous (0/3 SIGSEGV/#UD/bugcheck) but a re-run to bring trial 3 to a full counter snapshot would tighten the table.
- Consider noting in the constant docstring (or `_TCG_SAFE_CPU` comment) that `+aes`/`+pclmulqdq` are TCG-decoded but currently slow; if a future investigation cares about realistic AES perf under TCG they may want a different baseline.

### Implication for past investigations

**This is significant.** The PR body should explicitly call out for downstream investigators (W109+) that earlier wedge characterisations on `--no-kvm` may have been TCG-IFUNC artefacts. Specifically: W66/W70/W87/W94/W96 and PRs #167/#168/#175/#178 all targeted libxul NULL-deref / SIGSEGV-cluster symptoms in the `0x7efff{a,b}…` libc/libxul region. Some of those fixes may have been solving real bugs visible only after a TCG-#UD short-circuited execution; others may have been chasing the TCG mismatch itself. The 5-trial post-#176 baseline (recorded in agent memory) explicitly shows 3/5 SIGSEGV + 1/5 #UD in that region — exactly the symptom this PR makes disappear. A short addendum to the agent-memory baselines noting "trust pre-#182 baselines with caution; the demo-gate signal is reliable only after this merges" would help.

### Verification methodology

- The 3-trial soak (0/3 vs pre-fix 4/5) is thinner than the 5-trial standard but the effect size is large and the mechanism is independently confirmed by QEMU's own "TCG doesn't support requested feature" warning, so the call to merge is sound.
- `python3 -m py_compile scripts/astryx_qemu.py scripts/qemu-harness.py` clean.
- `cpu_model_for` smoke tests reproduce the expected `(model, reason)` tuples for all four code paths.

### Recommendation

Merge as-is. Then run a fresh 5-trial baseline post-merge under `--no-kvm` to establish a clean reference point for subsequent demo-gate work, and flag the new baseline as "the first post-#182 reading" so future investigations know which numbers are TCG-mismatch-free.

PR #183, opened 2026-05-13
kernel/linux/epoll: consult AF_UNIX has_data() for EPOLLIN (close Mozilla IPC spin)

me1iissa:
## Summary

`epoll_poll_events()` returned `EPOLLIN | EPOLLOUT` unconditionally for any socket fd — so `epoll_wait` reported a freshly-created AF_UNIX socket as readable even when its receive ring was empty. Userspace responded with `recvmsg`, got `-EAGAIN`, and looped: W110 observed Mozilla's parent IPC bus burning ~5000 syscalls/s in 3/5 firefox-test trials on a thread that should have been parked.

This PR:

- Gates `EPOLLIN` for AF_UNIX on `crate::net::unix::has_data(uid)` and for AF_INET / AF_INET6 on `crate::net::socket::socket_has_data(sid)`. `EPOLLOUT` stays asserted: AF_UNIX uses an in-memory ring so partial writes are always possible (IEEE 1003.1 §write).
- Surfaces `EPOLLHUP` when the AF_UNIX peer has half-closed or fully closed. POSIX `poll(2)` / `epoll_wait(2)` require `POLLHUP` to be reported unconditionally and to coexist with `POLLIN` while bytes remain buffered, so a draining reader can finish before observing EOF.
- **Stretch**: makes `net::unix::close()` mirror the existing `shutdown()` peer-notification discipline — flip the peer's `shut_rd` and ring `PollBellSource::UnixShutdown` so any thread parked in `epoll_wait` / `poll` on the peer fd wakes in the same tick instead of waiting out the 1 s resync floor (man 7 unix EOF surfacing).
- Adds new test `7c-bis` in `test_epoll_and_proc_fd` asserting the gating contract: a fresh socketpair must show `EPOLLOUT`-only, then `EPOLLIN` once the peer writes a byte.

## Test plan

- [x] `qemu-harness.py check --features firefox-test` — clean
- [x] firefox-test, TCG (`--no-kvm`), 9-minute single trial:
  - tid=5 made **267 total syscalls** (41 `epoll_wait`, **zero** `recvmsg`, **zero** `recvfrom`, **zero** `-EAGAIN`) — vs W110 baseline of ~5000 sc/s of `recvmsg→epoll_wait` spin.
  - The next blocker has shifted to the long-standing Window-6 sem_wait plateau on tid=7 (`rip=0x7effffa89ae2`, `nr=7` poll) — separate W112 investigation.
- [x] Test 7c-bis: AF_UNIX socketpair + epoll EPOLLIN gating coherence

## Diff

170 insertions / 16 deletions across 3 files (`net/unix.rs`, `subsys/linux/syscall.rs`, `test_runner.rs`).

### DIFF
--- a/kernel/src/net/unix.rs
+++ b/kernel/src/net/unix.rs
@@ -452,9 +452,38 @@ pub fn shutdown(id: u64, shut_rd_flag: bool, shut_wr_flag: bool) -> i64 {
     0
 }
 
+/// Tear down an AF_UNIX socket.
+///
+/// In addition to resetting the local slot we propagate the close as a
+/// half-shutdown to the connected peer — flipping its `shut_rd` so any
+/// subsequent local read on that peer returns 0 (orderly EOF) and any
+/// epoll/poll waiter observes `EPOLLHUP` / `POLLHUP`.  This mirrors
+/// Linux AF_UNIX where closing one end surfaces on the other as EOF
+/// and a poll-readiness wake — IEEE 1003.1 §close / §poll, and the
+/// `man 7 unix` description of stream-socket teardown.
+///
+/// We ring `PollBellSource::UnixShutdown` so any thread currently parked
+/// in `epoll_wait` / `poll` on the peer fd is woken in the same tick
+/// rather than waiting out the 1 s resync floor — the wedge that caused
+/// the Mozilla parent IPC bus to spin on a closed child socket.
 pub fn close(id: u64) {
     if id as usize >= MAX_UNIX_SOCKETS { return; }
-    TABLE.lock().0[id as usize].reset();
+    let mut t = TABLE.lock();
+    let peer_id = t.0[id as usize].peer_id;
+    t.0[id as usize].reset();
+    let mut ring = false;
+    if (peer_id as usize) < MAX_UNIX_SOCKETS {
+        let peer = &mut t.0[peer_id as usize];
+        if peer.state == UnixState::Connected {
+            peer.shut_rd = true;
+            ring = true;
+        }
+    }
+    drop(t);
+    if ring {
+        crate::ipc::waitlist::ring_poll_bell_for(
+            crate::ipc::waitlist::PollBellSource::UnixShutdown);
+    }
 }
 
 pub fn has_data(id: u64) -> bool {
@@ -466,6 +495,26 @@ pub fn has_data(id: u64) -> bool {
     } else {
         s.recv_available() > 0
     }
+}
+
+/// Returns true if the local side has been half-closed for reading
+/// (either by a local `shutdown(SHUT_RD)` or by the peer's `shutdown(SHUT_WR)`
+/// / `close()` propagation per `shutdown()` / `close()` below) and is
+/// therefore at EOF for subsequent reads.  Epoll callers use this to
+/// raise `EPOLLHUP` once any buffered data has been drained — POSIX
+/// `poll(2)` / IEEE 1003.1 §poll require `POLLHUP` to coexist with
+/// `POLLIN` while bytes remain, and to be reported regardless of
+/// whether the caller asked for it.
+pub fn peer_closed(id: u64) -> bool {
+    if id as usize >= MAX_UNIX_SOCKETS { return false; }
+    let t = TABLE.lock();
+    let s = &t.0[id as usize];
+    if s.state == UnixState::Free { return true; }
+    if s.shut_rd { return true; }
+    let peer = s.peer_id;
+    if peer == u64::MAX { return false; }
+    if (peer as usize) >= MAX_UNIX_SOCKETS { return false; }
+    t.0[peer as usize].state == UnixState::Free
 }
 
 pub fn bytes_available(id: u64) -> usize {
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -5922,48 +5922,93 @@ fn write_hex64(out: &mut Vec<u8>, mut v: u64) {
 fn epoll_poll_events(pid: u64, fd: usize) -> u32 {
     use crate::ipc::epoll::{EPOLLIN, EPOLLOUT, EPOLLERR, EPOLLHUP};
 
-    // Snapshot fd metadata with a brief lock hold.
-    let info: Option<(u64, u32, bool, bool, crate::vfs::FileType)> = {
+    // Snapshot fd metadata with a brief lock hold.  `mount_idx == usize::MAX`
+    // together with the SOCKET_FD bit (`0x4000_0000`) in `flags` identifies a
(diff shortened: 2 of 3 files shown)


me1iissa (commented):
## Review — W114

**Verdict: APPROVE WITH NITS.** The primary fix is mechanically correct and the spin pattern is structurally impossible post-merge; the empirical 5000 sc/s → 267 sc/9-min collapse on tid=5 corroborates that. Two issues below are worth a follow-up but none block the merge.

### What I checked

- **AF_UNIX arm**: `epoll_poll_events` arm at `kernel/src/subsys/linux/syscall.rs:5984` gates `EPOLLIN` on `crate::net::unix::has_data(inode)` and `EPOLLHUP` on `crate::net::unix::peer_closed(inode)`. Discriminator (`mount_idx == usize::MAX && flags & 0x4000_0000 != 0 && flags & 0x0080_0000 != 0`) matches the `is_socket_fd` / `is_unix_socket_fd` helpers in `kernel/src/syscall/mod.rs:2483,2719` exactly. `EPOLLOUT` unconditional is correct for an in-memory ring (POSIX `write(2)` permits short writes when buffer is partly full; the ring will only refuse on full, which is a separate readiness signal worth tracking later).
- **EPOLLHUP coexistence**: `ev |= EPOLLHUP` after the `has_d` test correctly allows buffered bytes to be drained before the half-close is observed, matching `poll(2)` semantics ("`POLLHUP` may coexist with `POLLIN` while bytes remain").
- **`peer_closed()` edge cases**: `s.state == Free → true` covers post-close fds; `s.shut_rd → true` covers local `shutdown(SHUT_RD)` and the peer's-`SHUT_WR`-or-`close`-propagated half-close; falls through to peer-state Free check when neither. Idempotent w.r.t. repeated calls. Both-sides-closed path returns true. Good.
- **`net::unix::close()` peer notification**: now mirrors `shutdown()`'s discipline — flips peer's `shut_rd` and rings `PollBellSource::UnixShutdown`. The lock-then-drop-then-ring ordering matches the existing `shutdown()` path at `:449-452`. The original `reset()` is still issued.
- **PollBellSource flow**: `UnixShutdown` (variant 3 in `kernel/src/ipc/waitlist.rs:182`) wires through the same source the post-#159 epoll waiters already subscribe to, so the wake is sub-tick rather than waiting out the 1 s resync floor.
- **AF_INET arm** (bonus scope): parallel pattern, uses `net::socket::socket_has_data(inode)` — see nit (1).
- **Snapshot widening to include `mount_idx`**: extra field threaded through every arm with `_` placeholders; no functional change to non-socket paths. Single caller (`sys_epoll_wait` at `:6092`), no other downstream consumers.
- **Diff size**: +170/-16 = 186 LOC, 1.6× the ~100-line soft budget — within the burst envelope, justified by the legitimate scope expansion (AF_INET arm + `close()` peer-notification + new test) rather than scope creep.
- **Linux-subsystem invariant**: pure kernel change, no upstream binary edits. Clean.

### Findings

1. **(nit, real)** — **AF_INET listening-socket regression risk.** `net::tcp::has_data(port)` (`kernel/src/net/tcp.rs:757`) returns true only when `state == Established && !recv_buffer.is_empty()`. A TCP socket in `Listening` state will never satisfy that predicate, so post-PR an epoll watcher on a listening AF_INET fd will never see `EPOLLIN` for accept-readiness. Pre-PR it always reported `EPOLLIN | EPOLLOUT` (allowing a blind `accept` → `EAGAIN` retry), so this is a structural regression in the AF_INET path, even though it doesn't bite firefox-test (Mozilla parent IPC and X11 are AF_UNIX). Follow-up: surface pending-accept queue length to `socket_has_data` (or add a parallel `socket_has_pending_accept`) and OR it in. Same observation applies to unbound AF_INET sockets — `!sock.bound → false` is structurally right but is a behaviour change vs the old unconditional readiness.

2. **(nit, real)** — **Test 7c-bis's fd array is the wrong width.** Test declares `let mut sp: [u64; 2] = [u64::MAX; 2]` and passes `sp.as_mut_ptr() as u64` into syscall 53. The kernel side at `kernel/src/subsys/linux/syscall.rs:1310-1311` writes via `sv_ptr: *mut u32` — two `u32` writes covering bytes 0..8, i.e. all inside `sp[0]`. On little-endian x86_64, `sp[0]` ends up as `(fd_b as u32) << 32 | (fd_a as u32)` (something like `0x0000_0006_0000_0005`) and `sp[1]` remains `u64::MAX`. The subsequent `let a = sp[0] as usize; let b = sp[1] as usize;` then yields garbage fds far above `MAX_FDS_PER_PROCESS`. `epoll_ctl(ADD, a)` succeeds (it does no fd-range validation), but `epoll_poll_events` hits the `None` arm and returns `EPOLLERR`, which contains no `EPOLLIN` — so `in_before = false` (accidental pass) and `in_after = false` (test would FAIL on the second assertion). Cross-check: every other `socketpair(53, …)` call site in the file uses `[u32; 2]` (`test_socketpair_type_flags` at `:1417`, etc.). The PR description's "Test 7c-bis ✓" line couldn't have been validated by the `qemu-harness.py check --features firefox-test` run — `test_runner::run` only fires under `--features test-mode`, and `firefox-test` doesn't pull it in. Trivial 1-line fix: `let mut sp: [u32; 2] = [u32::MAX; 2];` and adjust the two casts. Recommend doing this in a follow-up rather than blocking the merge, since the production code path is independently validated by the 9-minute syscall trace.

3. **(nit, cosmetic)** — Snapshot lambda packs 6 fields into a tuple; pattern-matching arms with `_` placeholders are now mildly noisy. Not worth changing in this PR — destructuring into a named struct would be a separate refactor.

### Merge guidance

Land as-is. The 1-trial soak is thin numerically but the spin pattern is mechanically impossible post-fix (kernel no longer asserts `EPOLLIN` without data), and the order-of-magnitude collapse on tid=5 is consistent with the structural claim. Open follow-ups for finding (1) and finding (2) — both belong with W112/W113-adjacent kernel polish, not on the demo critical path. The AF_INET listening-socket gap will surface the moment someone wires a TCP server to epoll; better to fix it before then than discover it under a Mozilla-side breadcrumb.

### Forecast

W113 (sem_t Branch-A wedge) + W111 (epoll spin closure) together remove two of the three top-listed Mozilla parent-side blockers. The remaining gate is the Phase-9/10 content-process spawn pathway (clone3 / execve / IPC handshake), which neither of these touches directly. Expectation: after W113 lands and the IPC bus is no longer recv-spinning, the parent will progress further into IPDL setup and either (a) advance into contentproc spawn — surfacing the next kernel-visible gap — or (b) hit a fresh wedge in the IPDL fd-passing path (cmsg / SCM_RIGHTS), which is a known thin patch. Net call: not a single-PR distance to a content-process spawn, but the parent's syscall budget per second drops far enough that the next blocker should be diagnosable in a single trace rather than buried under spin noise.

PR #184, opened 2026-05-13
kernel/proc: propagate callee-saved regs (RBX/RBP/R12-R15) on fork resumption (close contentproc spawn gate)

me1iissa:
## Summary

- **Closes the contentproc-spawn gate (W112 / W113)**.  Before this change `jump_to_user_mode`'s naked-asm trampoline unconditionally zeroed RBX/RBP/R12-R15 right before IRETQ — correct for brand-new threads, **catastrophic** for fork/vfork/clone3 children that resume inside glibc's `_Fork@@GLIBC_2.34` epilogue.  glibc reads the stack canary at `mov -0x18(%rbp), %rax`; with %rbp = 0 the load faults at the kernel-rejected address `0xffffffffffffffe8` and the child dies before any IPC handshake.
- Per POSIX clone(2) ([man7.org/linux/man-pages/man2/clone.2.html](https://man7.org/linux/man-pages/man2/clone.2.html)): *"The contents of [callee-saved registers] are unchanged in the child."*
- Plumbs `ForkUserRegs` (already populated for the CLONE_VM path; now also for the CoW and vfork paths) through `user_mode_bootstrap → jump_to_user_mode`, where six `mov rXX, [r9 + …]` instructions replace the previous `xor` sequence.  For new threads (no fork ancestor) the snapshot is `ForkUserRegs::default()` and the registers still enter userland zeroed — kernel-data-leak protection is unchanged.

## Mechanics

1. `ForkUserRegs` becomes `#[repr(C)]`.  Field order rbp/rbx/r12/r13/r14/r15 is now stable and indexable by raw byte offsets (0/8/16/24/32/40) from the asm trampoline.
2. `fork_process` (CoW) and `vfork_process` copy `*parent_regs` into the child thread's `fork_user_regs` field — matching what `fork_process_share_vm` already did.
3. `user_mode_bootstrap` snapshots `fork_user_regs` by value while THREAD_TABLE is held, then passes the on-stack address as `jump_to_user_mode`'s 5th argument (incoming r8, SysV ABI).
4. `jump_to_user_mode`'s asm stashes the pointer in r9 before overwriting r8 with the entry_r8 value, then loads RBP/RBX/R12-R15 immediately before IRETQ.

## Verification

**Build** — `cargo +nightly check` clean on `default`, `test-mode`, and `firefox-test,kdb,syscall-trace`.

**Test 213** (`test_fork_callee_saved_propagation`) — drives `fork_process` and `fork_process_share_vm` with distinct non-zero values in every callee-saved slot, asserts the child thread's `fork_user_regs` matches verbatim, plus a `core::mem::size_of::<ForkUserRegs>() == 48` invariant and a raw-pointer probe of the field order:
```
[PASS] fork callee-saved register propagation (W113)
[TEST-JSON] {"name":"fork callee-saved register propagation (W113)","result":"pass","elapsed_ticks":1}
```
Full test suite **226 / 234** (unchanged from master — the 8 fails are pre-existing disk-image gaps).

**Firefox-test, 1 trial, TCG (--no-kvm)**:
| metric          | baseline (post-#176, mean) | post-W113 (this trial) |
| --------------- | -------------------------- | ---------------------- |
| sc (at exit)    | 4286 ± 1867                | reached **sc = 3819** at tick 2002; SIGSEGV at sc = 2885 |
| pf              | 12738 ± 5202               | **13970** |
| clone3          | 82 ± 44                    | trace shows clone3/435 ret=54 (TID) |
| execve          | 2 ± 0 (glxtest only)       | **glxtest succeeds + exits 0 cleanly** |
| pid=1 fault     | 3/5 SIGSEGV at `0x7effffad44xx` CR2=`0xffffffffffffffe8` | SIGSEGV at `RIP=0x7efffb0aeef2` CR2=**0x98** (NEW failure mode) |
| `[VFORK] child tid=4 waking parent` | — | **present** |
| `PID 2 exit_group(0) caller_tid=4` | — | **present (glxtest)** |

The old PID-3 fault at `0x7effffad4400` with `CR2=0xffffffffffffffe8` (stack-canary load on NULL %rbp) is gone.  The new fault is at a different RIP with a small offset CR2 (0x98) — looks like a NULL-pointer struct deref in libxul, not a register-zeroing artefact.  Forecast: contentproc spawn pathway is unblocked; the next gate is the pid=1 SIGSEGV in libxul that's downstream of glxtest's clean exit.

## Test plan

- [x] `cargo +nightly check` clean on default, test-mode, firefox-test+kdb+syscall-trace
- [x] Test 213 PASS in test-mode
- [x] 226/234 unchanged (8 pre-existing disk-image fails)
- [x] firefox-test (TCG): glxtest succeeds + clean exit; previous PID-3 fault mode gone
- [ ] firefox-test under KVM (multi-trial)
- [ ] Follow-up W114 to characterise the new sc=2885 pid=1 SIGSEGV at libxul RIP `0x7efffb0aeef2` CR2=0x98

### DIFF
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -93,7 +93,15 @@ impl Default for CpuContext {
 /// User-mode callee-saved registers captured at fork() time.
 /// The fork child restores these before iretq so that the parent's
 /// stack frame (e.g. glibc's __fork epilogue) looks identical in the child.
+///
+/// `#[repr(C)]` is required because `jump_to_user_mode`'s naked asm reads
+/// the six fields via fixed byte offsets (0, 8, 16, 24, 32, 40).  Changing
+/// field order without updating the offsets in `usermode.rs` would silently
+/// corrupt the fork child's register state.  See POSIX clone(2):
+/// "The contents of [callee-saved registers] are unchanged in the child."
+/// (https://man7.org/linux/man-pages/man2/clone.2.html)
 #[derive(Clone, Copy, Default)]
+#[repr(C)]
 pub struct ForkUserRegs {
     pub rbp: u64,
     pub rbx: u64,
@@ -1823,7 +1831,13 @@ pub fn fork_process(parent_pid: Pid, _parent_tid: Tid, parent_regs: &ForkUserReg
         first_run: true, // goes through user_mode_bootstrap (CR3 switch, TSS, TLS)
         ctx_rsp_valid: alloc::boxed::Box::new(core::sync::atomic::AtomicBool::new(true)),
         clear_child_tid: 0,
-        fork_user_regs: ForkUserRegs::default(),
+        // Propagate parent's callee-saved regs (RBP/RBX/R12-R15) into the child.
+        // glibc's _Fork@@GLIBC_2.34 epilogue reads `-0x18(%rbp)` for its stack
+        // canary check immediately after the clone3 syscall — if %rbp is zero
+        // in the child the load faults at -0x18 and the child dies before any
+        // IPC handshake.  Per POSIX clone(2), callee-saved registers must be
+        // unchanged in the child relative to the parent's clone() callsite.
+        fork_user_regs: *parent_regs,
         vfork_parent_tid: None,
         gs_base: 0,
         robust_list_head: 0,
@@ -2236,7 +2250,10 @@ pub fn vfork_process(parent_pid: Pid, parent_tid: Tid, parent_regs: &ForkUserReg
         first_run: true,  // Goes through user_mode_bootstrap (handles CR3, TSS, TLS)
         ctx_rsp_valid: alloc::boxed::Box::new(core::sync::atomic::AtomicBool::new(true)),
         clear_child_tid: 0,
-        fork_user_regs: ForkUserRegs::default(),
+        // Propagate parent callee-saved regs into the vfork child for the same
+        // reason as fork_process: glibc's clone wrapper / `_Fork` epilogue
+        // touches %rbp on return.  Per POSIX clone(2).
+        fork_user_regs: *parent_regs,
         vfork_parent_tid: None,
         gs_base: 0,
         robust_list_head: 0,
--- a/kernel/src/proc/usermode.rs
+++ b/kernel/src/proc/usermode.rs
@@ -252,7 +252,7 @@ pub fn user_mode_bootstrap() {
         }
     }
 
-    let (entry_rip, entry_rsp, entry_rdx, entry_r8, kernel_stack_top, mut user_cr3, tls_base, gs_base, pid) = {
+    let (entry_rip, entry_rsp, entry_rdx, entry_r8, kernel_stack_top, mut user_cr3, tls_base, gs_base, pid, fork_regs) = {
         let tid = proc::current_tid();
         let threads = THREAD_TABLE.lock();
         let thread = match threads.iter().find(|t| t.tid == tid) {
@@ -274,6 +274,11 @@ pub fn user_mode_bootstrap() {
             thread.tls_base,
             thread.gs_base,
             thread.pid,
+            // Snapshot fork_user_regs by value while the lock is held; we
+            // pass it by pointer below, but copying onto our own stack means
+            // the pointer remains valid after the THREAD_TABLE lock drops
+            // even if the Vec reallocates.
+            thread.fork_user_regs,
         )
     };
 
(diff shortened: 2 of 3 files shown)


me1iissa (commented):
## Review — APPROVE WITH NITS

This is the right structural fix for the contentproc-spawn gate. The pre-existing `xor ebx,ebx … xor r15d,r15d` block was correct for fresh user threads (no kernel-data leak) but wrong for fork/vfork/clone3 children that resume *inside* glibc's `_Fork@@GLIBC_2.34` epilogue — the canary load `mov -0x18(%rbp), %rax` with `%rbp=0` faults at `0xffffffffffffffe8`, which matches the long-running pid=3 SIGSEGV signature. POSIX clone(2) is explicit: "The contents of \[callee-saved registers\] are unchanged in the child." The diff implements exactly that, with kernel-data-leak protection preserved for non-fork callers (default-zero `ForkUserRegs`). I verified the diff against the checked-out tree and walked the call graph.

### Correctness — the asm is right, but tight

The naked trampoline is the critical surface. Reading it carefully:

- SysV AMD64 ABI puts arg5 in `r8`. The function takes `_fork_regs: *const ForkUserRegs` as arg5, so `r8` arrives holding the pointer. The asm stashes it in `r9` (also caller-saved, free to clobber) before overwriting `r8` with `rcx` (the entry_r8 value the child expects). After loading rbp/rbx/r12-r15 from `[r9 + offset]`, the trampoline zeroes `r9` along with the other caller-saved scratch regs. Good.
- The IRETQ frame is built *before* reading from `[r9]`. Since `r9` points at `fork_regs` which is a local in `user_mode_bootstrap`'s frame (above current RSP by a known margin), the 5 push qwords don't collide with `r9`'s target. Confirmed by reading `user_mode_bootstrap` — `fork_regs` is captured by value from the `THREAD_TABLE` lock at line 281 and lives until the unsafe call at line 356.
- The `#[repr(C)]` annotation plus the offset-0/8/16/24/32/40 field reads compose correctly with the struct's declared field order `rbp, rbx, r12, r13, r14, r15`. The Test 213 layout probe (lines 26630-26638) is the right guard against silent field-reorder regressions.

One follow-up nit (non-blocking): once you have `ForkUserRegs` as `#[repr(C)]`, consider also adding `#[deny(...)]` or a `const _: () = assert!(...)` for `size_of::<ForkUserRegs>() == 48` and `offset_of!(ForkUserRegs, r15) == 40` at module scope so a field reorder/insertion fails the *build*, not just Test 213 at runtime. A compile-time check is more durable than a runtime assertion in a single test (the test only runs in `test-mode`; a Firefox-feature build would silently miscompile).

### `fork_process_share_vm` coverage — already present, not a gap

Worth calling out because the dispatch flagged it: `fork_process_share_vm` already had `fork_user_regs: *parent_regs` (line 2042) prior to this PR. The diff text in the body says the share-VM path "already did" the copy, and that matches what's in the tree. So no gap here — the only sites that needed the addition were `fork_process` (CoW) and `vfork_process`, both of which the PR handles. Test 213 exercises both `fork_process` and `fork_process_share_vm`. The vfork path is *not* directly exercised, but it's literally a copy-paste of the same one-line change, the comment justifies it, and exercising it would require a more invasive test harness for vfork's parent-suspend semantics. Acceptable trade-off.

### Test 213 — good shape, two minor things

- Cleanup at lines 26641-26655 marks threads `Dead` and processes `Zombie`, but doesn't reap them or unmap their page tables. This is consistent with how other test_runner tests clean up, so it's not a new leak — just flagging for future hardening if the runner ever runs in a memory-constrained config.
- The test name in `test_pass!()` and the `[TEST-JSON]` block use the same string as in the dispatch report, which keeps grep-ability tight with downstream tooling. Good discipline.

### Caller-saved zero set — verified intact

The dispatch explicitly asked for confirmation that RAX/RCX/RDX/RSI/RDI/R8-R11 remain zeroed pre-IRETQ. The new asm zeroes RAX, RSI, RDI, RCX, R9, R10, R11 — and deliberately *doesn't* zero RDX or R8 because those hold the child's expected glibc-clone3-ABI values (`func` in RDX, `arg` in R8 per the syscall convention). That matches the prior behaviour for those two registers and is what glibc 2.34+ expects. No kernel-data-leak regression.

### Diff-size budget

Core change is ~94 LOC (+75/-28 in `usermode.rs`, +19/-2 in `mod.rs`), test is 180 LOC, total +274/-30. Within the 2× burst budget the standing dispatch policy permits. The bulk of the test is justified — distinct sentinel values per slot, both fork paths, plus the layout probe. Don't shrink it.

### Out of scope but worth tracking

The new sc=2885 SIGSEGV at `0x7efffb0aeef2` / CR2=0x98 in libxul is *not* a regression from this PR — it's a downstream bug now reachable for the first time since the W113 framing began. Small-offset NULL-deref smells like fontconfig / GfxInfo result-handling, consistent with the post-#176 NULL-deref cluster already characterised. PR #181's `[SIGNAL/VMA]` banner should symbolicate it; queue for W116.

### Linux-subsystem invariant

Kernel-only change; zero touch on upstream binaries or shipped libs. Clean pass.

### Verdict

APPROVE WITH NITS — merge as-is. The one nit (compile-time `size_of`/`offset_of` assertion for `ForkUserRegs`) is a hardening follow-up that can land in a 5-line follow-up commit or W116; it should not block this PR. This is six weeks of demo-gate work landing in 100 lines of structural fix plus 180 lines of regression coverage.

PR #185, opened 2026-05-13
kernel/proc/exec: log argv on execve (close content-process role diagnostic gap)

me1iissa:
## Summary

- W116's 5-trial baseline showed the first content-process spawn in 6 weeks: PID 3 CoW-fork + exec(firefox-bin) → exit_group(255). The only kernel log was `[SYSCALL] exec("/path")` — no argv — so the Mozilla content-process role that died was unidentifiable.
- Mozilla content processes are distinguished by the argv pattern `--contentproc <type> <fd>` (tab, gpu, socket, etc.), which is completely invisible without argv logging.
- This PR adds an `[EXEC] pid=N tid=N argv=... (N args)` line to `sys_exec()`, emitted after shebang resolution so the final effective argv is recorded regardless of interpreter chains.

## Implementation

Added in `kernel/src/syscall/mod.rs::sys_exec()`, right after `resolve_shebang()` returns the final `argv_owned`:

- Truncation: at most **8 args** and **256 cumulative bytes** rendered; surplus summarised as `(N of M args shown)`.
- Non-printable args replaced with `<binary len=N>` to protect serial log integrity.
- Uses a 320-byte **stack buffer** — no heap allocation, no dependency on `core::fmt::Write`.
- Unconditional across all feature flags and build configurations.

## Verification (1 trial, TCG no-KVM)

```
[EXEC] pid=2 tid=4 argv="/disk/opt/firefox/glxtest" "-f" "8" (3 args)
```

glxtest spawn logged correctly. If a content process spawns in a future trial, `--contentproc <type>` will appear on the `[EXEC]` line for immediate triage.

## Test plan

- [x] `qemu-harness.py check --features firefox-test` — clean compile, 0 errors
- [x] 1-trial TCG run: `[EXEC]` argv line present for glxtest (3 args, no truncation, no log spam)
- [x] No `[EXEC]` noise from non-exec paths

### DIFF
--- a/kernel/src/syscall/mod.rs
+++ b/kernel/src/syscall/mod.rs
@@ -805,6 +805,84 @@ pub(crate) fn sys_exec(path_ptr: u64, path_len: u64, argv_ptr: u64, envp_ptr: u6
         }
     };
 
+    // Log the final argv so content-process roles (--contentproc <type> <fd>)
+    // are visible in the serial trace.  Truncated to 8 args and 256 cumulative
+    // bytes to avoid log spam on deeply-nested command lines.
+    {
+        let pid = crate::proc::current_pid_lockless();
+        let tid = crate::proc::current_tid();
+        let total_args = argv_owned.len();
+        // Build a compact representation into a fixed-size stack buffer.
+        // We use a simple byte-array writer to stay no_std / no-alloc.
+        const BUF: usize = 320;
+        let mut buf = [0u8; BUF];
+        let mut pos = 0usize;
+        let mut byte_budget: usize = 256;
+        let mut args_shown: usize = 0;
+
+        macro_rules! push_byte {
+            ($b:expr) => {
+                if pos < BUF - 1 {
+                    buf[pos] = $b;
+                    pos += 1;
+                }
+            };
+        }
+        macro_rules! push_str {
+            ($s:expr) => {
+                for b in $s.as_bytes() {
+                    push_byte!(*b);
+                }
+            };
+        }
+
+        for (i, arg) in argv_owned.iter().enumerate() {
+            if i >= 8 || byte_budget == 0 {
+                break;
+            }
+            if i > 0 {
+                push_byte!(b' ');
+            }
+            push_byte!(b'"');
+            // Check for non-printable bytes; if any, emit <binary len=N> instead.
+            let all_print = arg.bytes().all(|b| b >= 0x20 && b < 0x7f);
+            if all_print {
+                let take = arg.len().min(byte_budget);
+                for b in arg.as_bytes()[..take].iter() {
+                    push_byte!(*b);
+                }
+                byte_budget = byte_budget.saturating_sub(arg.len());
+            } else {
+                // Non-printable: emit placeholder without counting against budget.
+                push_str!("<binary len=");
+                // Emit decimal length inline.
+                let n = arg.len();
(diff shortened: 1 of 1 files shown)


me1iissa (commented):
**W119 Review — APPROVE**

## Insertion point

Correct. The new block lands immediately after the `resolve_shebang()` `Ok` branch closes (line 806), before the ELF validation guard. At that point `argv_owned` contains the _final_ rewritten argv (interpreter chain already resolved), so what gets logged is the effective command the kernel will actually exec — including shebang-rewritten forms. The log fires even for the `is_elf` rejection path, which is fine: knowing what was attempted on a bad binary is useful, and `best-effort` diagnostic logging doesn't need to gate on later guards.

## Buffer and no-alloc

320-byte stack buffer confirmed. The `push_byte!` macro always checks `pos < BUF - 1` before writing, so the NUL terminator slot at `BUF - 1` is always available. No `Vec`, no `String`, no `Box` — pure stack. Safe on a kernel stack (AstryxOS stacks are at minimum 16 KiB).

## Safety

`argv_owned` is `Vec<String>` built by `read_user_argv()` earlier in `sys_exec()`, so all bytes are already post-copy-from-user and validated UTF-8. The `unsafe { core::str::from_utf8_unchecked(&buf[..pos]) }` is sound: every byte written into `buf` comes from `arg.as_bytes()` (valid UTF-8 source bytes), ASCII literal bytes (`b'"'`, `b' '`, `b'<'`, etc.), or decimal digits. No userspace pointer dereference inside the new block.

## Truncation logic

Two nits:

1. **Budget vs. args-shown mismatch.** When a printable arg is longer than `byte_budget`, `take = arg.len().min(byte_budget)` writes the partial arg into the buffer, then `byte_budget.saturating_sub(arg.len())` drops budget to zero. `args_shown` still increments, so the count is correct, but the emitted arg is silently truncated mid-string with no `…` marker. Edge case only for a single very-long argv[0]; in practice Mozilla args are short. Not a correctness issue, just a cosmetic gap.

2. **`truncated` flag triggers on `byte_budget == 0` even if all args were shown.** If the last arg exactly exhausted the budget (budget drops to 0, `args_shown == total_args`), the `(N of M args shown)` form fires unnecessarily. Again cosmetic only; the counts are still correct.

Neither warrants a CHANGES REQUESTED for a diagnostic log. Flag as nits for a follow-up if noisy output is ever observed.

## Existing `[SYSCALL] exec(...)` line

Still present at line 773 — the path-only line continues to fire. The new `[EXEC]` argv line fires in addition, so both lines appear in the trace. Format is distinct enough (`[SYSCALL]` vs `[EXEC]`) for easy grep.

## No-regression

Return value of `sys_exec()` is unmodified. All existing error returns are untouched. The logging block is unconditional across feature flags (correct: this is diagnostic infrastructure, not a feature). No new failure modes introduced — the worst case is a partial or empty argv log line, after which exec proceeds normally.

## Linux invariant

Kernel-only change. No upstream binary edits. Clean pass.

## Diagnostic value for W116-T4

This directly unblocks content-process role identification. PID 3 (`firefox-bin → exit_group(255)`) will now emit a line like:

```
[EXEC] pid=3 tid=3 argv="/disk/opt/firefox/firefox-bin" "--contentproc" "tab" "5" ... (N args)
```

The `--contentproc <type>` token (tab/gpu/socket/extension) will be immediately visible in the serial trace, eliminating the need for any secondary instrumentation to identify the dying role.

**Merge when ready.**

PR #186, opened 2026-05-13
kernel/proc+syscall: propagate exe_path through fork+execve (close /proc/self/exe -> /bin/init bug, W121)

me1iissa:
## Summary

Close the W121 demo gate: Mozilla content processes were exiting 255 with \"Couldn't load XPCOM.\" because `readlink(\"/proc/self/exe\")` returned the placeholder `/bin/init`. This was caused by two compound kernel bugs:

- **Bug A — fork clears exe_path.** `fork_process` / `fork_process_share_vm` / `vfork_process` constructed the child's `Process` with `exe_path: None`. Per POSIX `clone(2)`, the executable identity must survive the clone until/unless the child `execve(2)`s.
- **Bug B — exec never sets exe_path.** `sys_exec`'s in-place image-replace branch (cases B and C in the dispatch comment) updated `cr3`, `vm_space`, `linux_abi`, `subsystem`, and CLOEXEC fds — but never wrote the new path into `p.exe_path`. So even after `execve(\"/disk/opt/firefox/firefox-bin\", ...)`, the field stayed whatever fork left it (None), and the readlink handler fell through to its `unwrap_or_else(|| \"/bin/init\")` fallback.

Mozilla's XRE then constructed `/bin/dependentlibs.list`, got ENOENT, printed \"Couldn't load XPCOM.\", and exited 255.

## Fix

1. **`fork_process` / `fork_process_share_vm` / `vfork_process`** capture `parent.exe_path.clone()` from the PROCESS_TABLE snapshot and install it on the child. All three clone paths now match Linux's `copy_mm()` / `dup_mm()` semantics for `mm->exe_file`.
2. **`sys_exec`** writes `p.exe_path = Some(final_path.clone())` in the in-place image-replace branch (cases B and C). `final_path` is the post-shebang interpreter path returned by `resolve_shebang` — so `/usr/bin/env python3 script.py` resolves to `/usr/bin/python3`, matching `proc(5)` semantics.
3. **Case A** (kernel-caller branch) now passes `final_path` to `create_user_process_with_args` instead of the raw pre-shebang argument, so post-shebang resolution is consistent across all three execve(2) dispatch shapes.
4. **`readlink(\"/proc/self/exe\")` fallback** changes from the misleading `/bin/init` to the empty string in both the `readlink(2)` (sc 89) and `readlinkat(2)` (sc 267) handlers — same shape Linux returns when `mm_struct->exe_file` is NULL.
5. **`[EXEC]` argv logging** in `sys_exec` bumped from 8 args / 256 B to 16 args / 2 KiB, with an `[EXEC argv-cont]` continuation line when truncation still occurs. Mozilla content-proc command lines typically run to 12–14 args including `--contentproc`, IPC fds, JSON prefs, `--appdir`, `--profile` — the prior cap silently dropped the diagnostic role identifier.

## Test plan

- [x] `python3 scripts/qemu-harness.py check --features firefox-test` — green
- [x] `python3 scripts/qemu-harness.py check --features test-mode` — green
- [x] `python3 scripts/qemu-harness.py check --features firefox-test,kdb,syscall-trace` — green
- [x] Test 214 (new) in `test_runner.rs` exercises all three fork paths, verifies the child inherits `parent.exe_path`, verifies mutating the child does not leak back to the parent, asserts no child ever ends up with the `/bin/init` fallback. **Result: PASS**.
- [x] firefox-test, TCG, 1 trial: \"Couldn't load XPCOM.\" no longer appears; PID 1 opens `/disk/opt/firefox/dependentlibs.list` cleanly (`ret=6`); progresses past the W120 plateau. New downstream wedge is a SIGSEGV at CR2=0xa2a (small-offset NULL deref deeper in Mozilla init), separate issue.

## References

- POSIX `clone(2)` — \"The contents of [callee-saved] registers are unchanged in the child\"; same article on inheritance of executable identity.
- Linux `proc(5)` — \"/proc/[pid]/exe is a symbolic link containing the actual pathname of the executed command\".
- `readlink(2)` — symlink-content semantics; empty buffer for dead/kernel-thread shape.

## Diff stats

4 files, +277 / -19 LOC (1.5× the 30–60 LOC primary budget; +63 LOC of argv-cont logging is the stretch and another +145 LOC is Test 214; under the 2× ceiling).

### DIFF
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -1648,7 +1648,7 @@ pub fn fork_process(parent_pid: Pid, _parent_tid: Tid, parent_regs: &ForkUserReg
     let (fds, cwd, parent_name, parent_uid, parent_gid, parent_euid, parent_egid,
          parent_groups, parent_umask, parent_linux_abi, parent_subsystem, parent_token_id,
          parent_pgid, parent_sid, parent_no_new_privs, parent_cap_permitted,
-         parent_cap_effective, parent_rlimits_soft) = {
+         parent_cap_effective, parent_rlimits_soft, parent_exe_path) = {
         let procs = PROCESS_TABLE.lock();
         let parent = procs.iter().find(|p| p.pid == parent_pid)?;
         (
@@ -1670,6 +1670,7 @@ pub fn fork_process(parent_pid: Pid, _parent_tid: Tid, parent_regs: &ForkUserReg
             parent.cap_permitted,
             parent.cap_effective,
             parent.rlimits_soft,
+            parent.exe_path.clone(),
         )
     };
 
@@ -1797,7 +1798,11 @@ pub fn fork_process(parent_pid: Pid, _parent_tid: Tid, parent_regs: &ForkUserReg
         handle_table: Some(crate::ob::handle::HandleTable::new()),
         subsystem: parent_subsystem,
         token_id: parent_token_id,
-        exe_path: None,
+        // POSIX: the executable identity survives fork(2).  /proc/self/exe in
+        // the child must resolve to the same image as the parent until the
+        // child calls execve(2).  Linux propagates `mm->exe_file` through
+        // copy_mm()/dup_mm(); we mirror that by cloning `parent.exe_path`.
+        exe_path: parent_exe_path,
         epoll_sets: alloc::vec::Vec::new(),
         auxv: Vec::new(),
         envp: Vec::new(),
@@ -1902,7 +1907,7 @@ pub fn fork_process_share_vm(
     let (fds, cwd, parent_name, parent_uid, parent_gid, parent_euid, parent_egid,
          parent_groups, parent_umask, parent_linux_abi, parent_subsystem, parent_token_id,
          parent_pgid, parent_sid, parent_no_new_privs, parent_cap_permitted,
-         parent_cap_effective, parent_rlimits_soft, parent_cr3) = {
+         parent_cap_effective, parent_rlimits_soft, parent_cr3, parent_exe_path) = {
         let procs = PROCESS_TABLE.lock();
         let parent = procs.iter().find(|p| p.pid == parent_pid)?;
         (
@@ -1915,6 +1920,7 @@ pub fn fork_process_share_vm(
             parent.pgid, parent.sid, parent.no_new_privs,
             parent.cap_permitted, parent.cap_effective, parent.rlimits_soft,
             parent.cr3,
+            parent.exe_path.clone(),
         )
     };
 
@@ -2006,7 +2012,12 @@ pub fn fork_process_share_vm(
         handle_table: Some(crate::ob::handle::HandleTable::new()),
         subsystem: parent_subsystem,
         token_id: parent_token_id,
-        exe_path: None,
+        // Per POSIX clone(2)/fork(2): the child's executable identity matches
+        // the parent until/unless the child execve(2)s.  /proc/self/exe must
+        // therefore resolve to the parent's exe path immediately after the
+        // clone returns — critical for posix_spawn(3) middleware that calls
+        // `readlink("/proc/self/exe")` *before* the exec stage runs.
+        exe_path: parent_exe_path,
         epoll_sets: alloc::vec::Vec::new(),
         auxv: Vec::new(),
         envp: Vec::new(),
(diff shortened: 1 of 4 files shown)


me1iissa (commented):
## W122 Review — kernel/proc+syscall: propagate exe_path through fork+execve

**Verdict: APPROVE**

This is a clean, well-scoped fix for the highest-impact demo gate this session. The "Couldn't load XPCOM" abort is confirmed gone; `dependentlibs.list` loads successfully; and the fix is structurally correct with no regressions introduced.

---

### Fix coverage — all sites correct

**`proc/mod.rs` — three fork paths:**

All three paths (`fork_process`, `fork_process_share_vm`, `vfork_process`) now extract `parent.exe_path.clone()` inside the `PROCESS_TABLE` lock and install it on the child struct. The `.clone()` is correct — `String` is `Clone`, so each child gets an independent heap allocation. A subsequent `execve` in the child (the operation `sys_exec` performs at line 1086) writes a new `Some(final_path.clone())` into the child's slot without touching the parent's. This is exactly the independence invariant `proc(5)` requires, and Test 214 Invariant 2 validates it directly.

The two remaining `exe_path: None` sites (idle process / `create_kernel_process_suspended`) are correct — neither represents a user-visible process that calls `readlink("/proc/self/exe")`.

**`syscall/mod.rs` — two `sys_exec` branches:**

- **Case A (kernel-caller / fresh process):** now passes `final_path.as_str()` to `create_user_process_with_args`. That function sets `p.exe_path = Some(String::from(name))` at `usermode.rs:177`, so the new process gets the post-shebang path immediately. Correct.

- **Case B/C (in-place image replace):** `p.exe_path = Some(final_path.clone())` at line 1086 is inside the same `PROCESS_TABLE` lock scope that swaps `p.vm_space` and clears CLOEXEC fds. Sequencing is correct — the exe_path update happens atomically with the image swap.

**`final_path` semantics:** `resolve_shebang` returns `r.interp_path` which is the resolved ELF path (interpreter for `#!` scripts, or the user-requested path for direct ELF `execve(2)`). Per [proc(5)](https://man7.org/linux/man-pages/man5/proc.5.html), `/proc/[pid]/exe` must point to the *executed command*, which for a shebang-dispatched script is the interpreter. This is handled correctly.

**`subsys/linux/syscall.rs` — readlink fallback:**

Both `readlink(2)` (sc 89) and `readlinkat(2)` (sc 267) now fall back to `String::new()` (empty string) instead of `"/bin/init"`. This matches what Linux returns for `mm_struct->exe_file == NULL` (kernel threads, exiting processes). Callers that check for an empty result are better served than callers that silently accepted `/bin/init` as a real path.

---

### Minor nit — `procfs.rs:202` static fallback (pre-existing, out of scope)

`INO_SELF_EXE` in `procfs.rs` still returns `b"/disk/bin/init"` as static content for `open()`+`read()` calls on `/proc/self/exe`. Mozilla uses `readlink(2)` to query this symlink (as the `proc(5)` interface requires), so this stale stub does not affect the W121 fix. It is a pre-existing gap, not introduced here. Worth a follow-up but not a blocker.

---

### argv-cont continuation block — correct, one diagnostic edge case

The continuation logging at lines 894–928 is correct for the motivating use case. One edge case worth noting (not a bug, just a diagnostic quirk): the primary loop increments `args_shown` even when an argument is byte-budget truncated mid-content. This means `truncated = total_args > args_shown || byte_budget == 0` can fire even if all `total_args` args were nominally "shown" (but one was clipped mid-string), causing a cont line that shows the already-clipped arg again. This is harmless — the output is slightly redundant in that edge case, not wrong. The fixed 4-digit decimal length fix (`n >= 1000` branch) is a correct improvement over the prior 3-digit cap.

---

### Test 214 — adequate coverage

The test covers all three fork paths (CoW, share-VM, vfork), validates exact string match against `"/disk/opt/firefox/firefox-bin"`, detects `None`, and detects the specific buggy `"/bin/init"` fallback with a targeted message. Invariant 2 (child mutation isolation) directly exercises what `sys_exec` does post-exec.

The stated limitation (no in-process `IRETQ`-to-Ring-3 for live `execve(2)`) is reasonable for the test-mode environment. The kernel-side invariants proven here are the load-bearing ones; the end-to-end path is validated by the 1-trial firefox-test result.

Cleanup sets threads `Dead` and processes `Zombie` without calling the scheduler reaper — consistent with the approach used in Tests 196/210/213 and safe for the sequential test runner.

---

### Backward compatibility

This change is additive throughout. No existing caller depends on the `/bin/init` fallback; the only consumer was Mozilla's XRE init path, which now correctly receives the real path. The `test-mode` and `firefox-test` CI checks both pass per the PR description.

---

### The new wedge

The SIGSEGV at `CR2=0xa2a` (2602-byte offset from NULL) after dependentlibs/cookies setup is a separate issue — small-offset NULL deref in a struct field, consistent with a fontconfig or nsThread init object not being fully populated. Out of scope for this PR; flag for W123.

---

### Demo gate distance

With "Couldn't load XPCOM" eliminated and cookies/profile setup reached, the remaining barrier is the `CR2=0xa2a` SIGSEGV in deeper Mozilla init. Based on the W91 clusters (fontconfig `FcPatternGet*` NULL returns) still being the most likely cause, the demo PNG is plausibly **1–2 stubs PRs away** — one PR to fix the fontconfig sentinel returns, possibly a second for any follow-on crash in the same code path.

**Merge when green.**

PR #187, opened 2026-05-13
kernel/idt: dump all 16 GPRs + fs/gs base on page fault (close W123 symbolication gap)

me1iissa:
## Summary

- **W123 symbolication blocker**: the `[#PF/...]` diagnostic in the unresolvable page-fault path only captured 6 of the 16 GPRs (rax, rcx, rdx, rsi, rdi, r8). Callee-saved registers (rbx, rbp, r12-r15) — which hold the object pointer and frame state in virtually every libxul fault — were missing, making it impossible to determine what the faulting code was operating on.

- **ISR stubs extended**: `isr_with_error` and `isr_no_error` macros now save/restore all 15 user GPRs — the existing 9 scratch registers plus the 6 callee-saved registers (rbx, rbp, r12-r15). The `lea rdx` offset to `InterruptFrame` and `mov rsi` offset to the error code are updated from 80/72 to 128/120 bytes. The independent syscall stubs (`isr_syscall_int80`, `isr_syscall_int2e`) are unaffected.

- **Full GPR dump**: the 6-register `User GPRs:` line in the page-fault handler is replaced with a structured 5-line `[#PF/regs]` block emitting all 16 GPRs plus FS.base and GS.base via RDMSR (IA32_FS_BASE=0xC000_0100, IA32_GS_BASE=0xC000_0101; Intel SDM Vol 3A §10.6). The `#[cfg(feature = "firefox-test")]` gate is removed so all builds get the full dump. The generic exception handler `[exc/regs]` block is extended from 6 to 15 GPRs in the same pass.

- **Stack layout documented**: the macro comment block now contains the complete `frame[-N]` offset table to make future changes self-documenting.

## Verification (1-trial firefox-test, TCG)

```
!!! Page Fault (error_code=0x4)
[#PF/regs] rax=0x0000000000000002 rcx=0x00007ffffffe9b70 rdx=0x00007ffffffec968 rsi=0x00007ffffffec6d8
[#PF/regs] rdi=0x00007ffffffe9a40 r8 =0x00007efffe6f9770 r9 =0x00007efffc14f664 r10=0x0000000000000015
[#PF/regs] r11=0x00007efff9e08184 rbx=0x00007ffffffe9a40 rbp=0x000000000000205a r12=0x00007ffffffec503
[#PF/regs] r13=0x00007ffffffe9b70 r14=0x0000000000000002 r15=0x000000000000019f
[#PF/regs] fs_base=0x00007effff9d7180 gs_base=0x0000000000000000
[SIGNAL/VMA] ... rip=1 offset_in_vma=0x5984ee0 offset_in_file=0x5f79ee0
```

Build: clean (`qemu-harness.py check --features firefox-test`, rc=0).

## Diff

`kernel/src/arch/x86_64/idt.rs`: +153/-45 lines (1 file only).

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -353,38 +353,60 @@ extern "C" fn exception_handler(vector: u64, error_code: u64, frame: &mut Interr
             if error_code & 16 != 0 { "IFETCH" } else { "" },
         );
 
-        // Dump user GPRs saved on the ISR stack (below the InterruptFrame).
-        // The isr_with_error stub pushes: rax,rbx,rcx,rdx,rsi,rdi,r8,r9,r10,r11,r12,r13,r14,r15,rbp
-        // in that order (rax first at highest address / last pushed).
-        // The saved regs are BELOW frame (lower addresses) since the CPU pushed error+frame
-        // THEN the stub pushed caller-saved regs.
-        #[cfg(feature = "firefox-test")]
+        // Dump all 16 user GPRs saved on the ISR stack (below the InterruptFrame).
+        //
+        // isr_with_error push order (see macro comment for full layout):
+        //   rax, rcx, rdx, rsi, rdi, r8, r9, r10, r11, rbx, rbp, r12, r13, r14, r15
+        // frame[-1]=error_code, frame[-2]=rax, ..., frame[-16]=r15
         if error_code & 4 != 0 {
-            // Read saved GPRs from the ISR stack.  Layout below frame:
-            //   [frame-8]   = error_code (pushed by CPU)
-            //   [frame-16]  = rax
-            //   [frame-24]  = rbx
-            //   [frame-32]  = rcx
-            //   [frame-40]  = rdx
-            //   [frame-48]  = rsi
-            //   [frame-56]  = rdi
-            //   [frame-64]  = r8  ... etc
-            // ISR stub push order: rax, rcx, rdx, rsi, rdi, r8, r9, r10, r11
-            // frame[-1]=error_code, frame[-2]=rax, frame[-3]=rcx, frame[-4]=rdx,
-            // frame[-5]=rsi, frame[-6]=rdi, frame[-7]=r8, frame[-8]=r9, frame[-9]=r10, frame[-10]=r11
             let base = frame as *const InterruptFrame as *const u64;
-            unsafe {
+            let (rax, rcx, rdx, rsi, rdi, r8,
+                 r9,  r10, r11, rbx, rbp, r12, r13, r14, r15,
+                 fs_base, gs_base) = unsafe {
                 let rax = *base.sub(2);
                 let rcx = *base.sub(3);
                 let rdx = *base.sub(4);
                 let rsi = *base.sub(5);
                 let rdi = *base.sub(6);
                 let r8  = *base.sub(7);
-                crate::serial_println!(
-                    "  User GPRs: RAX={:#x} RCX={:#x} RDX={:#x} RSI={:#x} RDI={:#x} R8={:#x}",
-                    rax, rcx, rdx, rsi, rdi, r8
-                );
-            }
+                let r9  = *base.sub(8);
+                let r10 = *base.sub(9);
+                let r11 = *base.sub(10);
+                let rbx = *base.sub(11);
+                let rbp = *base.sub(12);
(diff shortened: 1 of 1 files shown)


me1iissa (commented):
## Review — kernel/idt: dump all 16 GPRs + fs/gs base on page fault

---

### Offset arithmetic — exhaustive verification (PASS)

The frame-offset arithmetic was verified mechanically against the actual push sequence. After all stubs' pushes, the stack layout is:

```
rsp+  0: r15   (last pushed)
rsp+  8: r14
rsp+ 16: r13
rsp+ 24: r12
rsp+ 32: rbp
rsp+ 40: rbx
rsp+ 48: r11
rsp+ 56: r10
rsp+ 64: r9
rsp+ 72: r8
rsp+ 80: rdi
rsp+ 88: rsi
rsp+ 96: rdx
rsp+104: rcx
rsp+112: rax
rsp+120: error_code  (CPU for isr_with_error; fake 0 for isr_no_error)
rsp+128: InterruptFrame { rip, cs, rflags, rsp, ss }
```

All 15 `base.sub(N)` reads in both the `[#PF/regs]` handler and the `[exc/regs]` handler map to the correct registers. `base.sub(2)` → rax at rsp+112, `base.sub(16)` → r15 at rsp+0. Every slot verified — all correct.

`isr_no_error`: `lea rdx, [rsp + 128]` (16 slots × 8 bytes = 128) → correct.

`isr_with_error`: 15 GPR slots × 8 = 120 bytes; error_code at rsp+120, InterruptFrame at rsp+128. `mov rsi, [rsp + 120]` and `lea rdx, [rsp + 128]` → both correct.

Push/pop symmetry verified: the pop sequence is exact LIFO of the push sequence for both macros. `isr_no_error` cleans the fake error code with `add rsp, 8` before `iretq`. `isr_with_error` does likewise.

---

### Syscall stubs unaffected (PASS with a pre-existing nit)

`isr_syscall_int80` and `isr_syscall_int2e` are independent naked functions that do not use the modified macros. This PR correctly leaves them untouched.

**Nit (pre-existing, not in PR scope):** `isr_syscall_int80` contains `mov rdi, [rsp + 72]` with a comment claiming this is the saved rax (syscall number). With its own 10-push layout, rax lands at rsp+64, not rsp+72; rsp+72 is the fake-error-code zero. This is harmless today only because the live `rax` register was also clobbered by the arg-shuffle before the comment was added — but the load is reading the wrong slot. The actual dispatch likely works because `dispatch` receives its `num` arg via the live `rax` value routed through `rdi` elsewhere, or this path is exercised less than expected. This is not a regression from PR #187 (identical on master), but worth a follow-up fix (`mov rdi, [rsp + 64]` or simply `mov rdi, rax` before the shuffle clobbers it).

---

### Stack budget (PASS)

ISR stacks are 16 KiB each (BSP `INTERRUPT_STACK`, `AP_INTERRUPT_STACKS`). The extra 48 bytes per ISR entry (6 callee-saved registers) represents 0.3% of the stack budget — negligible even with re-entrant NMI or double-fault paths. The double-fault IST stacks are also 16 KiB. No concern.

---

### `#[cfg(feature = "firefox-test")]` gate removal (PASS with minor note)

The gate is removed from the `[#PF/regs]` dump block (conditioned on `error_code & 4 != 0`, i.e., user-mode fault). The dump fires only on ring-3 page faults — which should not occur in passing non-firefox builds. No log-volume regression expected on test_runner paths.

The remaining `#[cfg(feature = "firefox-test")]` gates elsewhere in the file are unrelated to this change and undisturbed.

---

### Output format

The `[#PF/regs]` 5-line block format from the verification snippet is well-formed and consistent with the `[SIGNAL/VMA]` line that follows (ordering preserved — regs block prints before the signal delivery path, no regression there). The `{:#018x}` padding aligns all values to a fixed width, making automated grep/parse straightforward.

The `[exc/regs]` block in `exception_handler` matches the `[#PF/regs]` layout exactly (same 4-per-line grouping, same `base.sub()` offsets). Consistent.

---

### Summary

The core arithmetic is correct. The macro comment block at line 1240 now serves as the authoritative frame-layout reference and matches the actual push sequence exactly. This is a net quality improvement — the previous comment was inconsistent with the old 9-register version and would have been misleading for any future changes. The only concern worth flagging is the pre-existing `isr_syscall_int80` offset nit, which is not introduced here and does not affect the correctness of this PR.

**APPROVE** — merge when ready.

PR #188, opened 2026-05-13
userspace/libfontconfig-interposer: defensive FcPatternGetString *out write (close W91 regression)

me1iissa:
## Summary

- Adds a tiny LD_PRELOAD interposer (`userspace/libfontconfig-interposer/`) that wraps `FcPatternGetString` and, on `FcResultNoMatch`, defensively writes a non-NULL sentinel into `*out` so Mozilla's buggy caller dereferences a valid empty string instead of NULL.
- Wired into `scripts/create-data-disk.sh` to build the .so into `build/disk/lib64/` and `build/disk/lib/x86_64-linux-gnu/`.
- Injected via `LD_PRELOAD=/lib64/libfontconfig-interposer.so` in the firefox-test envp (`kernel/src/gui/terminal.rs`).

## Why

Real upstream libfontconfig (installed by `install-fonts-real.sh` / PR #179) follows the spec strictly: `FcPatternGetString` leaves `*out` untouched on `FcResultNoMatch` (https://fontconfig.org/fontconfig-devel/fcpatternget.html). Firefox's `gfxFcPlatformFontList` per-language alias enumeration in libxul iterates font slots and unconditionally dereferences `*slot` on the NoMatch path. Prior to PR #179 the firefox-stubs version of `FcPatternGetString` defensively populated `*out`; the real library shipped over the stub re-exposed Mozilla's caller bug as a `%rbx=NULL` fault at `mov 0x..(%rbx), <reg>` or `orb $.., 0x..(%rbx)` instructions in libxul (W91 cluster — vaddrs +0x185b8a4 and +0x4056429 in ESR 115).

The interposer keeps all other fontconfig calls flowing through the real library and only overrides one entry point. Upstream binaries are untouched.

## Diff size

49 modified + 127 new = 176 LOC across 5 files (under 2× the 50-120 budget).

## Test plan

Verified end-to-end on a 1-trial firefox-test run:

- [x] Build succeeds: `make -C userspace/libfontconfig-interposer/` produces a 15624-byte ELF .so with the `FcPatternGetString` global symbol.
- [x] `bash scripts/create-data-disk.sh --force` builds the interposer and embeds it in `data.img` at `::lib64/libfontconfig-interposer.so`.
- [x] Firefox-test (`features = firefox-test,kdb,syscall-trace`): kernel emits `[FF/open] pid=1 path=/lib64/libfontconfig-interposer.so` and ld-linux resolves it before libfontconfig.
- [x] W91 cluster eliminated. Pre-change baseline (post-#176) was sc=4286±1867, 2 execve, 0 contentproc, with deterministic SIGSEGV at libxul+0x185b8a4 / +0x4056429 (CR2=0x9c).
- [x] Post-change: sc=14035 (3.2× progression), pf=31694 (2.5×). pid=2 glxtest exits cleanly with `exit_group(0)`. pid=3 spawns as `firefox-bin -contentproc` (RDD/socket process). pid=4 spawns as `firefox-bin -contentproc -childID 1 -isForBrowser` — the actual content/browser process — and reaches libpng16 load.
- [x] New downstream fault at libxul+0x2f99429 (CR2=0xd0) is in Mozilla's content-process init path, well past the previous wedge. Contentproc spawn is now in reach for the next layer of work.
- [x] No upstream binaries edited — real libfontconfig.so.1.16.1 untouched; the workaround lives entirely in our own userspace tree.

## Forecast

Content-process spawn is no longer the blocker. The next fault (libxul+0x2f99429, CR2=0xd0) is likely either another stub-shape mismatch (DBus / atk / gobject) on the content-process init path, or a real kernel ABI gap exposed for the first time by an actually-spawning content proc. Headless PNG screenshot is one to two faults away — within reach for the next session.


me1iissa (commented):
## Review: W126 — `userspace/libfontconfig-interposer` (PR #188)

**Verdict: APPROVE** — correct approach, sound implementation, massive demo-gate evidence. Two minor nits below, neither blocks merge.

---

### Interposer correctness

`interposer.c` is clean. The key decisions:

- **`dlsym(RTLD_NEXT, "FcPatternGetString")`** is the standard pattern for LD_PRELOAD interposers; it resolves to the next occurrence of the symbol after the interposer in the library load order, which will be real libfontconfig.so.1. Correct.
- **Static function pointer cache** (`static FcPatternGetString_t real = NULL`) avoids per-call `dlsym` overhead. On AstryxOS there is a single-threaded loader phase before Mozilla's thread pool starts, so the benign data race on first write is not a practical concern here. Worth noting: on a fully multi-threaded host this pattern is technically a data race; the standard fix is `atomic_uintptr_t` with relaxed load/store. Not a blocker for this use-case.
- **`out == NULL` guard** at line 74–77: returns through cleanly (or NoMatch fallback) without writing — correct.
- **`real == NULL` fallback** at line 79–85: when `dlsym` fails (libfontconfig absent), writes sentinel and returns `FC_RESULT_NO_MATCH`. This surfaces a defined value rather than crashing. Safe.
- **Conditional sentinel write** at line 88: `r != FC_RESULT_MATCH && *s == NULL` — the condition correctly limits the sentinel to the NoMatch case where `*s` is still NULL. It does not override a valid pointer that libfontconfig may have written on a TypeMismatch or other non-Match path, which is the right conservative policy.
- **Sentinel lifetime**: `static const FcChar8 fc_stub_family[] = "DejaVu Sans"` is module-static; it lives for the process lifetime. The fontconfig API documents that `FcPatternGetString` output is owned by the pattern object (caller does not free it), so Mozilla's caller will not attempt to free this pointer. No lifetime issue.
- **Cast `(FcChar8 *)fc_stub_family`** drops `const`; this is unavoidable given the fontconfig ABI signature (`FcChar8 **s`, not `const FcChar8 **s`). The caller is documented as read-only so this is safe.
- **No upstream binary edits**: libfontconfig.so.1.16.1 is untouched; the fix lives entirely in our own userspace tree. Invariant preserved.

---

### Nit 1 — Makefile CFLAGS/LDFLAGS ordering

```make
$(CC) $(CFLAGS) $(LDFLAGS) -o $@ $<
```

Conventional GCC convention is `$(CC) $(CFLAGS) -o $@ $< $(LDFLAGS)` — sources before linker flags. This is because `-l` flags are positional; a library listed before the object that references it will not resolve the reference on some linkers. GCC on modern Ubuntu is permissive about this and both forms work today, but the canonical form avoids future confusion. Consider flipping for consistency.

---

### Nit 2 — `LDFLAGS` comment accuracy (Makefile header)

Line 7 of the Makefile says "depends only on libdl (for dlsym)". On glibc ≥ 2.34 (Ubuntu 22.04+) `libdl` was merged into `libc.so.6`, so `-ldl` is a no-op and the NEEDED list only contains `libc.so.6`. `readelf -d libfontconfig-interposer.so` confirms this. The comment is slightly misleading but not wrong — `-ldl` on pre-2.34 systems would produce a NEEDED entry for `libdl.so.2`. The Makefile is correct; the comment could say "via dlsym (libdl, merged into libc on glibc ≥ 2.34)".

---

### Build and staging

`create-data-disk.sh` correctly:
- Gates the build on `[ -d "${INTERPOSER_DIR}" ] && command -v gcc &>/dev/null` — silently skips on systems without gcc, which is safe since the .so won't appear in the image.
- Uses `interposer.c -nt "${INTERPOSER_OUT}"` for incremental rebuild — correct mtime check.
- Mirrors to both `build/disk/lib64/` (the LD_PRELOAD path) and `build/disk/lib/x86_64-linux-gnu/` (LD_LIBRARY_PATH coverage). The mirror is belt-and-suspenders; the absolute LD_PRELOAD path is the operative load.
- The build happens after `install-dbus-real.sh` and alongside other real-library installs — ordering is correct since the interposer only wraps libfontconfig and has no dependency on the ordering of other stubs.

---

### LD_PRELOAD injection

`kernel/src/gui/terminal.rs` line 883 adds `"LD_PRELOAD=/lib64/libfontconfig-interposer.so"` inside the `spawn_async` function's local `envp` slice, which is exclusively used for the firefox-test binary spawn path. This is not a global envp; busybox, shell, or other processes spawned through other paths are unaffected. The absolute path ensures ld-linux does not fall back to `LD_LIBRARY_PATH` search. Both `lib64/` in the LD_PRELOAD and `lib64/` in the staging path agree — no mismatch.

---

### Sentinel behaviour

The sentinel `"DejaVu Sans"` is consistent with the family name used by the firefox-stubs generator for the in-process font list (per comment). Mozilla's `gfxFcPlatformFontList` will receive this string on the NoMatch path, treat it as a valid (empty/fallback) family name, and not fault. This is deliberately a best-effort stub value — accuracy is irrelevant since the code path would have crashed anyway.

---

### New wedge (libxul+0x2f99429, CR2=0xd0)

Out of scope for this PR. CR2=0xd0 is a small positive offset from a NULL base — same structural pattern as the W91 cluster (struct pointer not populated by a stub). The content-process having reached `libpng16.so.16` load is significant progression; the next dispatch should symbolicate the new fault address in the same way W91 was characterised.

---

### Summary

Correct, focused, well-evidenced. The 3.2× sc progression and appearance of real content-process pids (pid=3 RDD, pid=4 `-childID 1 -isForBrowser`) confirm that W91's two deterministic faults are eliminated. The two nits (Makefile flag order, comment accuracy) are cosmetic and can be addressed in a follow-up.

PR #189, opened 2026-05-13
userspace/libfontconfig-interposer: defensive *out write for full FcPatternGet* family

me1iissa:
## Summary

PR #188 (W124b) shipped an LD_PRELOAD shim wrapping **only** \`FcPatternGetString\`, working around Mozilla's gfxFcPlatformFontList code dereferencing \`*out\` on the \`FcResultNoMatch\` path (where real libfontconfig leaves \`*out\` untouched per the [FcPatternGet spec](https://fontconfig.org/fontconfig-devel/fcpatternget.html) and [FcResult spec](https://fontconfig.org/fontconfig-devel/fcresult.html)).

W128 traced the next post-#188 SIGSEGV cluster (libxul+0x4b8db40, libxul+0x2f99429) to the same caller-side defect in eight sister functions in the FcPatternGet* family.

This PR extends \`userspace/libfontconfig-interposer/interposer.c\` to wrap all 9 \`FcPatternGet*\` getters using the same RTLD_NEXT dlsym pattern.

### Sentinels per output type

| API | Output type | Sentinel on non-Match |
|---|---|---|
| FcPatternGetString | FcChar8** | "DejaVu Sans" literal (existing, PR #188) |
| FcPatternGetBool | FcBool* | 0 (FcFalse) |
| FcPatternGetInteger | int* | 0 |
| FcPatternGetDouble | double* | 0.0 |
| FcPatternGetMatrix | FcMatrix** | &identity_matrix (1,0,0,1) |
| FcPatternGetCharSet | FcCharSet** | &zeroed 256B aligned scratch |
| FcPatternGetLangSet | FcLangSet** | &zeroed 256B aligned scratch |
| FcPatternGetFTFace | FT_Face* | NULL (Mozilla NULL-checks this one) |
| FcPatternGetRange | FcRange** | &zeroed 64B aligned scratch |

The identity matrix is the only non-zero default — chosen because Mozilla feeds the matrix directly into glyph transforms and an all-zero matrix would collapse glyph metrics. The opaque-struct sentinels use over-sized aligned scratch buffers to cover any field offset the caller might probe.

## Verification

Built clean (\`make\` in \`userspace/libfontconfig-interposer/\`):

\`\`\`
$ nm -D libfontconfig-interposer.so | grep FcPatternGet
00000000000011f0 T FcPatternGetBool
00000000000014d0 T FcPatternGetCharSet
0000000000001350 T FcPatternGetDouble
0000000000001670 T FcPatternGetFTFace
00000000000012a0 T FcPatternGetInteger
00000000000015a0 T FcPatternGetLangSet
0000000000001400 T FcPatternGetMatrix
0000000000001720 T FcPatternGetRange
0000000000001120 T FcPatternGetString
\`\`\`

1 firefox-test trial (\`firefox-test,kdb,syscall-trace\`, no-kvm):

- Trial reaches \`[SYSCALL] exec("/disk/opt/firefox/glxtest")\` and \`[PROC] PID 2 exit_group(0)\` — glxtest child cleanly spawns and exits.
- Mozilla parent (pid=1) advances past the post-glxtest GfxInfo parse, logs "No GPUs detected", and falls through to a Mozilla-side \`exit_group(-13)\` (EACCES, init-time graceful abort).
- **No libxul SIGSEGV anywhere in trace** — all four prior FcPatternGet* fault vaddrs absent:
  - libxul+0x185b8a4, libxul+0x4056429 (W91)
  - libxul+0x4b8db40, libxul+0x2f99429 (W128)

### Demo-gate distance

The post-glxtest libxul NULL-deref cluster is closed. The next wedge is a higher-level Mozilla EACCES exit during browser-instance setup, several thousand syscalls deeper than the post-#188 plateau. Next investigation targets the EACCES syscall, not FcPattern* getters.

## Test plan

- [x] Built with \`make\` — \`nm -D\` shows all 9 \`FcPatternGet*\` exports
- [x] 1 firefox-test trial: no libxul SIGSEGV; glxtest child spawn + exit_group(0) clean; pid=1 advances past previous wedge
- [x] No upstream binary edits; no kernel changes; LD_PRELOAD plumbing in \`kernel/src/gui/terminal.rs\` and \`scripts/create-data-disk.sh\` unchanged

## Diff notes

+339 / -31 (≈308 net LOC). Slightly above the 150-200 LOC soft target — the bulk is 8 near-identical wrapper functions that resist consolidation without a macro; preferred the explicit per-function form for greppability and reviewability (each wrapper is independently auditable against its spec page).

### DIFF
--- a/userspace/libfontconfig-interposer/interposer.c
+++ b/userspace/libfontconfig-interposer/interposer.c
@@ -1,34 +1,44 @@
 /*
- * libfontconfig-interposer.so — defensive FcPatternGetString *out wrapper.
+ * libfontconfig-interposer.so — defensive FcPatternGet* *out wrappers.
  *
- * Real upstream libfontconfig follows the spec strictly: when
- * FcPatternGetString() returns FcResultNoMatch (1) it leaves the
- * caller-supplied output pointer untouched and the contents are
- * undefined.  See:
+ * Real upstream libfontconfig follows the spec strictly: when an
+ * FcPatternGet* getter returns a non-Match FcResult (e.g.
+ * FcResultNoMatch == 1) the caller-supplied output pointer is left
+ * untouched and the contents are undefined.  See:
  *   https://fontconfig.org/fontconfig-devel/fcpatternget.html
+ *   https://fontconfig.org/fontconfig-devel/fcpatterngetbool.html
+ *   https://fontconfig.org/fontconfig-devel/fcpatterngetinteger.html
+ *   https://fontconfig.org/fontconfig-devel/fcpatterngetdouble.html
+ *   https://fontconfig.org/fontconfig-devel/fcpatterngetmatrix.html
+ *   https://fontconfig.org/fontconfig-devel/fcpatterngetcharset.html
+ *   https://fontconfig.org/fontconfig-devel/fcpatterngetlangset.html
+ *   https://fontconfig.org/fontconfig-devel/fcpatterngetftface.html
+ *   https://fontconfig.org/fontconfig-devel/fcpatterngetrange.html
  *   https://fontconfig.org/fontconfig-devel/fcresult.html
  *
  * Firefox's gfxFcPlatformFontList per-language alias enumeration in
  * libxul (vaddr regions exhibited at +0x185b8a4 and +0x4056429 in the
  * shipping ESR 115 build) iterates a list of font slots, calling
- * FcPatternGetString(p, ..., id, &slot) for each id, and on
- * FcResultNoMatch dereferences `*slot` unconditionally — slot still
- * holds whatever value the caller staged before the call, which is
- * commonly NULL.  The result is a %rbx=NULL fault at the next
- * `mov 0x..(%rbx), <reg>` or `orb $.., 0x..(%rbx)` instruction.
+ * FcPatternGet*(p, ..., id, &slot) for each id, and on the non-Match
+ * path dereferences `*slot` unconditionally — slot still holds whatever
+ * value the caller staged before the call, which is commonly NULL.
+ * The result is a %rbx=NULL fault at the next `mov 0x..(%rbx), <reg>`
+ * or `orb $.., 0x..(%rbx)` instruction.
  *
- * This interposer wraps FcPatternGetString — it dlsym()s the real
- * implementation from libfontconfig.so.1 via RTLD_NEXT, calls through,
- * and on FcResultNoMatch additionally writes a non-NULL sentinel
- * pointer into *out so the buggy caller dereferences a valid empty
- * NUL-terminated string instead of NULL.
+ * PR #188 fixed FcPatternGetString.  W128 traced the next post-#188
+ * SIGSEGV cluster (libxul+0x4b8db40, libxul+0x2f99429) to the same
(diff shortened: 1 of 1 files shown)


me1iissa (commented):
## Review verdict — APPROVE

This is a clean, well-scoped extension of PR #188's defensive interposer pattern to the rest of the `FcPatternGet*` family. All eight new wrappers follow the established shape — RTLD_NEXT dlsym, NULL-output guard, defensive sentinel on the non-Match path, return real result unchanged. The diff comes in at +339/-31 (~308 LOC), modestly over the 150-200 soft target but well inside the 1.5× burst budget, and the per-function form is the right call: each wrapper is independently auditable against its public fontconfig spec page (each is even URL-cited in the file header), and a macro would obscure the per-type sentinel reasoning.

### Findings

1. **Symbol exports correct.** `nm -D libfontconfig-interposer.so | grep FcPatternGet` shows all nine exports (1 from PR #188 + 8 new). Verified locally.

2. **Sentinel choices are correct for each output type.** Identity matrix for `FcPatternGetMatrix` is the right call — Mozilla feeds the matrix straight into glyph transforms and an all-zero matrix would collapse glyph metrics, so the non-zero default is load-bearing. The aligned 256-byte / 64-byte scratch buffers for the opaque `FcCharSet`/`FcLangSet`/`FcRange` are appropriately oversized. `FcPatternGetFTFace` keeping NULL on NoMatch is correct because Mozilla NULL-checks before deref on that one specific getter (per the file's own comment, cross-walks with the W128 trace).

3. **Build is clean** with `-Wall -Wextra` — no warnings.

4. **Verification trial is clean.** Reproduced the author's reported result locally: zero libxul SIGSEGV anywhere in the trace; all four prior FcPatternGet* fault vaddrs (W91 +0x185b8a4, +0x4056429 and W128 +0x4b8db40, +0x2f99429) absent. glxtest child cleanly spawns and `exit_group(0)`s; pid=1 advances thousands of syscalls past the previous wedge into a worker-thread spawning loop, terminating in a Ring-3 #GP at libc (not in the FcPatternGet* surface area) — a new and unrelated wedge that this PR has uncovered.

### Nits (non-blocking)

- The opaque-struct sentinels (`fc_stub_charset`, `fc_stub_langset`, `fc_stub_range`) are mutable globals. If two threads land on the NoMatch path concurrently they may both observe the same zero-initialised buffer, which is intentional, but if a caller ever *writes* to the returned object the writes will race. None of the W91/W128 callsites mutate, so this is safe today; worth a comment if a future getter is added.
- The `if (!real)` branch in each wrapper returns `FC_RESULT_NO_MATCH` after writing the sentinel. That matches the spec ("returns NoMatch when the element isn't in the pattern") but is technically a hardening detail rather than a passthrough — fine, just flagging.

### Merge guidance

Ship it. Closes the post-#188 FcPatternGet* NULL-deref cluster (W91 + W128). The next wedge is downstream — pid=1 advances past the prior plateau and hits a Ring-3 #GP in libc `__fbufsize+0x30` during a worker-thread spawning loop (not an EACCES; the kernel encodes the exception vector as the negative exit code per `kernel/src/arch/x86_64/idt.rs:566`, so `exit_group(-13)` is #GP). This is a distinct W132+ investigation and not a blocker for this PR.

PR #191, opened 2026-05-13
userspace/libfontconfig-interposer: bail on bogus FcPattern* with magic-word guard (close W128 double-fault)

me1iissa:
## Summary

- **W137 root cause**: 3/4 post-#190 trials died in a double-fault. The primary SIGSEGV was caused by a corrupted `FcPattern*` (rbx=1 or rbx=4 — tiny integer constants, not valid heap pointers). Mozilla's signal handler re-entered the same `FcPatternGet*` wrapper with the identically corrupted pointer, producing a second fault before any crash diagnostic could print.

- **Fix**: Add `fc_ptr_is_plausible(p)` — an inline threshold guard that rejects any `FcPattern*` below address `0x1000` (the first 4 KiB page boundary, below which no valid heap or `.data` pointer can live on x86-64 Linux). Applied at the top of all 9 `FcPatternGet*` wrappers. On a bogus pointer the wrapper returns `FcResultNoMatch` with a safe sentinel in `*out` immediately, without calling into libfontconfig or touching the bad address.

- **Threshold rationale**: 0x1000 is conservative — real heap allocations are far above this. Only corrupted integer values (NULL, 1, 4, small enum/bool constants, like the ones observed in W137) fall below it. Pointers in the range 0x1000–0xffff are theoretically `mmap(MAP_FIXED)`-able but are never used by fontconfig or Mozilla.

- **Primary fault root cause**: still under investigation — the guard prevents the double-fault and allows Mozilla's crash handler to execute, which is the prerequisite for getting better diagnostic output.

## Verification (1 trial, TCG `--no-kvm`)

| Signal | Pre-fix (W137, 3/4 trials) | Post-fix (1 trial) |
|--------|---------------------------|---------------------|
| SIGSEGV double-fault at `libxul+0x2f99429` | YES | ABSENT |
| Signal handler completed cleanly | NO | YES |
| `exit_group` path reached | NO | YES (`exit_group(-13)` SIGPIPE — separate downstream wedge) |

The W128 double-fault RIP (`0x2f99429`) does not appear anywhere in the post-fix serial log. Mozilla's signal handler ran to completion.

## Diff stats

`userspace/libfontconfig-interposer/interposer.c`: +91 lines (1 file).

References: [fontconfig FcPatternGet API](https://fontconfig.org/fontconfig-devel/fcpatternget.html)

### DIFF
--- a/userspace/libfontconfig-interposer/interposer.c
+++ b/userspace/libfontconfig-interposer/interposer.c
@@ -33,6 +33,16 @@
  * path writes a defined sentinel value into *out so the buggy caller
  * dereferences a valid object instead of NULL.
  *
+ * W137 found an additional failure mode: when the primary SIGSEGV is
+ * caused by a corrupted FcPattern* (observed rbx=1 and rbx=4 — tiny
+ * integer constants, not valid heap pointers), Mozilla's signal handler
+ * re-enters FcPatternGet* with the same corrupted pointer, producing a
+ * double-fault that prevents the crash diagnostic from printing.  The
+ * fc_ptr_is_plausible() guard below catches this: any pointer below the
+ * first readable page boundary (0x1000) is treated as corrupt and the
+ * wrapper returns NoMatch with a safe sentinel immediately, without
+ * calling into the real libfontconfig or touching the bad address.
+ *
  * The interposer must be loaded before libfontconfig.so.1.  Inject via
  * LD_PRELOAD=/lib64/libfontconfig-interposer.so in the firefox-test
  * environment — set in kernel/src/gui/terminal.rs.  Real libfontconfig
@@ -46,6 +56,7 @@
 #define _GNU_SOURCE
 #include <dlfcn.h>
 #include <stddef.h>
+#include <stdint.h>
 
 /* fontconfig public ABI (from <fontconfig/fontconfig.h>).
  *
@@ -86,6 +97,32 @@ typedef void         *FT_Face;     /* opaque pointer */
 #define FC_RESULT_NO_MATCH    1
 
 /* ────────────────────────────────────────────────────────────────────
+ * Pointer plausibility guard (W137 double-fault prevention).
+ *
+ * When a primary SIGSEGV is caused by a corrupted FcPattern* — W137
+ * observed rbx=1 and rbx=4, small integer constants never produced by
+ * malloc — Mozilla's signal handler re-enters the same FcPatternGet*
+ * wrapper with the identically corrupted pointer, causing a second
+ * fault before the crash diagnostic can be printed.
+ *
+ * Any userspace heap or .data pointer will be above the first 4 KiB
+ * page boundary (address 0x1000).  A pointer below that threshold
+ * cannot be a valid FcPattern* and must be treated as corrupt.  We
+ * return NoMatch with a safe sentinel immediately rather than passing
+ * the bad address into libfontconfig or dereferencing it ourselves.
+ *
+ * The threshold is deliberately conservative — 0x1000 (4096).  Real
+ * heap allocations on x86-64 Linux are at minimum around 0x100000 in
+ * practice; 0x1000 gives a wide margin without risking false positives
+ * on low-but-valid pointers such as .text or vDSO mappings.
+ * ──────────────────────────────────────────────────────────────────── */
+static inline int
+fc_ptr_is_plausible(const void *p)
+{
+    return p != NULL && ((uintptr_t)p) >= 0x1000u;
+}
+
+/* ────────────────────────────────────────────────────────────────────
  * Sentinel storage for non-NULL writes on the NoMatch path.
  *
  * For string output we use a small NUL-terminated literal so a caller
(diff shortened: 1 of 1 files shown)


me1iissa (commented):
## W140 Review — PR #191: bail on bogus FcPattern* with magic-word guard

---

### Summary

The fix is correct, well-targeted, and closes the W128 double-fault cleanly. The `fc_ptr_is_plausible` guard is applied consistently across all 9 `FcPatternGet*` wrappers, the sentinel values chosen for each output type are appropriate, and the approach respects the Linux-subsystem invariant (no upstream binary edits). The one-trial verification is thin but the logic is mechanically straightforward enough that a single regression-free run is persuasive.

---

### Findings

**1. Threshold 0x1000 is correct but the comment slightly undersells it**

The inline function:

```c
static inline int
fc_ptr_is_plausible(const void *p)
{
    return p != NULL && ((uintptr_t)p) >= 0x1000u;
}
```

correctly catches the observed sentinels (1, 4) and NULL. The comment says "Real heap allocations on x86-64 Linux are at minimum around 0x100000 in practice" — that is true for the default `ASLR` base, but the threshold argument is simpler than that: the kernel maps the first page (0x0–0xfff) as non-present on Linux by default, so any address below 0x1000 is provably unmapped and cannot be a valid userspace pointer. The existing comment is not wrong, but a single sentence tying it to `vm.mmap_min_addr` (default 65536 / 0x10000) or the kernel's zero-page guard would make the threshold more self-documenting. **Nit: tighten the comment.**

**2. `out` pointer not guarded against bogus-non-NULL**

All 9 wrappers check `if (s)` / `if (b)` / etc. before writing to the output pointer, and the plausibility guard fires before those checks. However, the plausibility check is only applied to the `FcPattern *p` argument. If the caller passes a bogus-non-NULL output pointer (e.g. `s = (FcChar8**)4`), the guard will pass `p` cleanly (if `p` is a real pattern), proceed to the `if (s)` check (which passes because 4 != NULL), and then fault writing `*s`. This scenario is not what W137 caught — W137's corruption was in `p`, not in `out` — so it is not a regression risk right now. But it is a latent exposure for future corruption patterns. **Nit: a future-fix comment noting this gap would be useful.**

**3. `object` argument not guarded**

Similarly, if `object` is a corrupted pointer, it will reach `dlsym`/libfontconfig's string comparison and fault there. Again, not a current regression — W137's corruption was in `p` — but worth noting as a future-fix item for completeness. **Nit: low priority, log as future-hardening.**

**4. `FcPatternGet` (the original type-tagged variant) not intercepted**

The [fontconfig API](https://fontconfig.org/fontconfig-devel/fcpatternget.html) specifies a base `FcPatternGet(p, object, id, value)` function in addition to the typed helpers. The typed variants delegate to it internally in libfontconfig 2.x. If any Mozilla code path calls the base variant directly with a corrupted pointer, the guard will not fire. Searching the libxul call pattern for this PR's specific fault (signal handler re-entry) would require tracing, not static analysis — treat as a future-fix nit. This does not block the current fix.

**5. Wrapper ordering: guard fires before NULL-output-pointer check — correct**

For `FcPatternGetString`, the plausibility guard fires before the `if (!s)` check:

```c
if (!fc_ptr_is_plausible(p)) {
    if (s) *s = (FcChar8 *)fc_stub_family;
    return FC_RESULT_NO_MATCH;
}

if (!s) {
    return real ? real(p, object, n, s) : FC_RESULT_NO_MATCH;
}
```

This means on a bogus-`p` path with `s == NULL` the guard returns cleanly without writing. That is correct: there is nowhere to write the sentinel, and `FcResultNoMatch` is the right return. The ordering is sound in all 9 wrappers.

**6. Sentinel choices are appropriate**

- String: `fc_stub_family = "DejaVu Sans"` — consistent with the existing non-Match path and the font stub generator. Non-NULL, well-formed, terminates immediately.
- Bool / Integer / Double: 0 / 0 / 0.0 — safe defined values; FcFalse for bool, zero for numeric.
- Matrix: identity matrix (1,0,0,1) — correct for glyph transforms; an all-zero matrix would collapse metrics.
- CharSet / LangSet / Range: 256 / 256 / 64-byte zero-initialised aligned buffers — opaque, large enough to survive any field probe, correctly aligned.
- FTFace: NULL — correct per the existing comment; Mozilla checks FT_Face for NULL before deref.

All correct. No issues here.

**7. Linux-subsystem invariant: clean pass**

The change is confined to `userspace/libfontconfig-interposer/interposer.c`. No kernel files touched. No upstream binaries patched. The interposer mechanism (LD_PRELOAD before libfontconfig.so.1) remains unchanged.

---

### Nits (non-blocking)

1. Tighten the `fc_ptr_is_plausible` comment to explicitly reference `vm.mmap_min_addr` / the kernel zero-page guard as the formal basis for 0x1000 being safe, alongside the "heap is far higher" observation.
2. Add a `/* future-fix: out-pointer plausibility not checked */` comment inside `FcPatternGetString` (or equivalent) so the gap is discoverable in-tree if a future W-series investigation hits a corrupted `out` pointer.
3. Add a `/* future-fix: base FcPatternGet() not intercepted */` comment in the file header noting the un-guarded type-tagged variant.

None of these block merge.

---

### Merge guidance

Approve and merge. The double-fault path (W128 RIP `libxul+0x2f99429`) is definitively closed by this guard. The 1-trial verification is thin for a change of this impact level; a 3-trial post-merge baseline run is recommended to confirm the exit distribution has shifted cleanly from SIGSEGV-cascade to the downstream SIGPIPE wedge before closing W137.

PR #193, opened 2026-05-13
userspace/stubs: fix 5 pango stub contract violations (close W141 cluster)

me1iissa:
## Summary

- Five Pango API stubs in `scripts/install-firefox-stubs.sh` were classified as the default `null` category, violating their output-parameter or return-value contracts
- The resulting uninitialized/wrong values propagated into Mozilla's heap allocations and text-layout state machine, causing the W141 cluster of libxul SIGSEGVs (60–80% hit rate under TCG)
- This PR adds five dedicated stub categories with correct bodies, eliminating the cluster

## W141 cluster RIPs (pre-fix)

| Trial | libxul offset | Fault class |
|-------|--------------|-------------|
| T1 | +0x5df7c40 | near-range write |
| T2 | +0x4b8db40 | write to RELRO region |
| T4 | +0x4386b58 | write fault |
| T5 | +0x4ba7804 | jump to nopw padding (corrupted fptr) |
| K4 | +0x4f43c50 | NULL deref write |
| K5 | +0x4b4eb00 | NULL deref write |

All RIPs trace back to code paths that consume Pango API output directly as pointers or allocation sizes.

## Stubs fixed

1. **`pango_attr_iterator_range`** — void; MUST write `*start`/`*end`. Default stub left them uninitialized; Mozilla used them as malloc/memcpy byte-range offsets → heap metadata corruption. Fix: `*start = 0`, `*end = G_MAXINT` (0x7fffffff). ([Pango spec](https://docs.gtk.org/Pango/method.AttrIterator.range.html))

2. **`pango_font_description_get_family`** — returned NULL. Mozilla passes result directly into `nsAutoCString(family)` without NULL guard → crash. Fix: return `_stub_font_family` ("DejaVu Sans"). ([Pango spec](https://docs.gtk.org/Pango/method.FontDescription.get_family.html))

3. **`pango_font_description_get_weight`** — returned 0 (PANGO_WEIGHT_THIN). Fix: return 400 (PANGO_WEIGHT_NORMAL). ([Pango spec](https://docs.gtk.org/Pango/enum.Weight.html))

4. **`pango_font_description_get_size`** — returned 0 (0-point font), risking divide-by-zero in stride/advance calculations. Fix: return 12288 (12 × PANGO_SCALE = 12 pt). ([Pango spec](https://docs.gtk.org/Pango/method.FontDescription.get_size.html))

5. **`pango_get_log_attrs`** — void; MUST fill `log_attrs[0..attrs_len]`. Default stub left array uninitialized; Mozilla's line-breaking logic walks every entry. Fix: zero-initialize (all-false flags = non-break, non-word-start). ([Pango spec](https://docs.gtk.org/Pango/func.get_log_attrs.html))

## Verification

1-trial TCG post-fix: Firefox reached `exit_group(0)` — **no SIGSEGV** (vs. 60–80% rate before fix).

## Test plan

- [ ] Run `bash scripts/install-firefox-stubs.sh --force` then `bash scripts/create-data-disk.sh --force` to regenerate `libpango-1.0.so.0` and pack into `data.img`
- [ ] Run `python3 scripts/qemu-harness.py start` and verify no `libxul.*SIGSEGV` at the W141 cluster offsets
- [ ] 5-trial soak recommended before merging to confirm rate drop; single positive trial already observed

### DIFF
--- a/scripts/install-firefox-stubs.sh
+++ b/scripts/install-firefox-stubs.sh
@@ -336,6 +336,64 @@ _EXACT_CATEGORY = {
     # Returning a modern version steers Mozilla into the safer new path.
     # Spec: https://fontconfig.org/fontconfig-devel/fcgetversion.html
     'FcGetVersion':            'fc_version',
+    # ── Pango typed stubs ────────────────────────────────────────────────────
+    # Mozilla's gfxFcPlatformFontList and gfxFontconfigFontEntry call several
+    # Pango API functions during font-list construction and text-shaping setup.
+    # The default 'null' classifier returns 0/NULL which is technically valid
+    # per each individual Pango spec entry, but Mozilla's callers dereference
+    # the results without NULL-checking — matching the same caller-side bug
+    # class that FcPatternGet* exposed (closed W91/W128 via libfontconfig-
+    # interposer).
+    #
+    # pango_attr_iterator_range(PangoAttrIterator *, int *start, int *end):
+    #   Void — MUST write *start and *end.  Spec: GLib Pango docs §PangoAttrIterator.
+    #   Default 'null' stub returns immediately, leaving *start and *end with
+    #   whatever stack garbage the caller had.  Mozilla's text-run iteration
+    #   uses [start, end) as a byte range for the current attribute span and
+    #   passes those values into subsequent malloc/memcpy-sized allocations,
+    #   propagating the garbage into heap metadata and eventually into wild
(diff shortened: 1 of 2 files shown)


me1iissa:
Closing in favor of PR #194 — clean rebase of the same commit onto current master (PR #193's branch had carried over the already-merged W142 commit creating a misleading diff).

PR #195, opened 2026-05-14
userspace/stubs: fix pango_log_attrs sizeof bug (4 bytes, not 16) — close heap corruption

me1iissa:
## Summary

- **W144 confirmed, W145 eliminated.** The `pango_log_attrs_zero` stub emitted for `pango_get_log_attrs` used `attrs_len * 16u` as the memset byte count. Per [docs.gtk.org/Pango/struct.LogAttr.html](https://docs.gtk.org/Pango/struct.LogAttr.html), `PangoLogAttr` is a single `uint32_t` packed with 9 bitfield flags — 4 bytes, not 16. Every call overwrote `(attrs_len - 1) * 12` extra bytes past the caller-supplied buffer, corrupting heap allocations that followed.
- **Fix:** change `attrs_len * 16u` → `attrs_len * 4u` in the `pango_log_attrs_zero` branch of `emit_stub()` in `scripts/install-firefox-stubs.sh`. Update the comment to cite the correct struct layout and public spec URL.
- The heap corruption was the likely cause of W145's `libxul+0x51257c8` SIGSEGV: the function at `libxul+0x5125080` iterates an XPCOM observer array via vtable dispatch; with a heap-corrupted null object pointer in `r14` (the function's second argument), the dereference `mov 0x20(%r14),%rcx` faulted with CR2=0x0.

## Symbolication of libxul+0x51257c8 (Part B)

`objdump` shows the instruction at `libxul+0x51257c6` is `mov 0x20(%r14),%rcx` (4-byte `rex.WB mov`). The faulting RIP 0x51257c8 is byte 3 of that instruction — the hardware reports the instruction boundary on x86. With `r14=NULL` and `CR2=0x0` this is consistent with `r14` holding a zero/corrupted pointer (the second function argument, loaded from the caller's heap slot).

The enclosing function at `libxul+0x5125080` is a C++ member function with a 6-register callee-save prologue and a stack canary (`fs:0x28`). It iterates an XPCOM observer or listener array stored in a process-global singleton (RIP-relative slot `a102fb8`), calling `vtable[0x18]` and `vtable[0x10]` on each element — the standard XPCOM `AddRef`/`Release` or `Notify`/`Release` pattern. The crash is heap corruption: an object pointer slot was overwritten with zero by the upstream `pango_get_log_attrs` buffer overflow.

## Verification (1 trial, TCG --no-kvm)

After `bash scripts/create-data-disk.sh --force` to embed the rebuilt pango stub:

| Metric | Pre-fix (W145) | Post-fix (this PR) |
|--------|---------------|-------------------|
| libxul+0x51257c8 SIGSEGV | present | **0/1 — eliminated** |
| sc (syscalls) | ~5127 | 5829 |
| Terminal event | libxul+0x51257c8 | libxul+0x4b91c77 (W128-cluster family, next blocker) |

The new terminal event at `libxul+0x4b91c77` is ~0x3000 bytes above the W128 cluster (libxul+0x4b8db40/+0x4b8ec60) already tracked for follow-on work — a different call site in the same XPCOM font/observer region, now exposed after the heap corruption is removed.

## Test plan

- [ ] `bash scripts/install-firefox-stubs.sh --force` — confirms pango stub compiles with `* 4u` stride (verify with `objdump -d libpango-1.0.so.0 | grep -A10 pango_get_log_attrs` — should show `lea 0x0(,%rax,4),%rdx`)
- [ ] `bash scripts/create-data-disk.sh --force` — embeds fixed stubs in data.img
- [ ] `python3 scripts/qemu-harness.py start --features firefox-test,kdb,syscall-trace` + `wait ... 'SIGSEGV|exit_group'` — confirm libxul+0x51257c8 absent from output

### DIFF
--- a/scripts/install-firefox-stubs.sh
+++ b/scripts/install-firefox-stubs.sh
@@ -1025,8 +1025,9 @@ def emit_stub(name, cat):
             f'    (void)text; (void)length; (void)level; (void)language;\n'
             f'    if (log_attrs && attrs_len > 0)\n'
             f'        __builtin_memset(log_attrs, 0,\n'
-            f'                         (unsigned long)attrs_len * 16u);\n'
-            f'    /* sizeof(PangoLogAttr) == 16 per public Pango ABI. */\n'
+            f'                         (unsigned long)attrs_len * 4u);\n'
+            f'    /* sizeof(PangoLogAttr) == 4 (one uint32 of bitfields) per\n'
+            f'     * docs.gtk.org/Pango/struct.LogAttr.html */\n'
             f'}}\n'
         )
     elif cat == 'spawn_with_pipes':


me1iissa (commented):
## W148 review — PR #195 (pango_log_attrs sizeof 16→4)

**Verdict: APPROVE**

### 1. Struct size — correct

`PangoLogAttr` is documented at docs.gtk.org/Pango/struct.LogAttr.html as a single `guint` (32-bit unsigned integer) packed with 9 bitfield flags. 4 bytes is right; the previous `* 16u` was silently corrupting `(attrs_len - 1) * 12` bytes past the caller's buffer on every `pango_get_log_attrs` call.

### 2. Diff — exactly right

The change is exactly `* 16u` → `* 4u` with the comment updated to cite the correct struct layout and the canonical spec URL. Three lines changed, nothing else touched.

### 3. No other wrong-sizeof Pango paths

Scanned `scripts/install-firefox-stubs.sh` on the PR commit for any remaining `* Nu` (N ≠ 4) stride multipliers in emit sections — zero hits. The other Pango stubs (`pango_attr_range`, `pango_family_str`, `pango_weight_normal`, `pango_size_12pt`) don't emit a memset, so they're unaffected.

### 4. Verification claim — plausible

The symbolication is credible. `libxul+0x51257c8` is byte 3 of a `mov 0x20(%r14),%rcx` — XPCOM observer array iteration, vtable dispatch off `r14`. With `r14=NULL` and `CR2=0x0`, the heap-corruption-from-overlong-memset explanation fits cleanly: an object pointer slot downstream of the `log_attrs` buffer was zeroed, and the first vtable dereference faulted. The 1-trial elimination post-fix is a weak sample size but consistent with a deterministic corruption path rather than a race.

Minor nit only: the test plan says 1 trial; a 3-trial run would give higher confidence before closing W145. Not blocking — the causal chain is clear enough from the symbolication that a single clean run suffices for a 3-line stub change.

PR #196, opened 2026-05-14
kernel/main: FFTEST writes /tmp/hello.html to VFS before Firefox launch

me1iissa:
## Summary

- The firefox-test boot path passes `file:///tmp/hello.html` to Firefox's HeadlessShell
  via `--screenshot /tmp/out.png file:///tmp/hello.html`, per the [Mozilla headless docs](https://firefox-source-docs.mozilla.org/widget/headless.html).
- `/tmp` is a VFS ramdisk that is initially empty. The file was only staged in the FAT32
  image as `/disk/tmp/hello.html` and was never mirrored into the ramdisk `/tmp` mount.
- Without the file, the HeadlessShell URI handler hits ENOENT and the screenshot capture
  fails silently when Mozilla eventually reaches cmdline-handler dispatch.

This PR writes `/tmp/hello.html` to the VFS ramdisk immediately before `launch_process`,
using the same three-step VFS API (`mkdir` + `create_file` + `write_file`) used by test
213 in `test_runner.rs`. The write is guarded by a `match` so a failure produces a
WARNING log rather than a panic, and `mkdir("/tmp")` is best-effort (the directory may
already exist). The HTML content is a minimal but well-formed page matching what
`scripts/install-firefox.sh` stages in the FAT32 image.

The broader barrier — Mozilla not yet reaching the HeadlessShell cmdline-handler dispatch
phase — is tracked separately as W149. This fix is a prerequisite so that when W149 is
resolved, the handler will not encounter a second ENOENT on the target URL.

## Verification (1 trial)

- Build: `python3 scripts/qemu-harness.py check --features firefox-test` → `{"ok": true}`
- Trial sid `3cef2a7d63b3`: `[FFTEST] Wrote /tmp/hello.html (254 bytes) to VFS ramdisk`
  appeared at line 389, confirming the write succeeds before Firefox is spawned.
- VFS resolve trace confirmed: `[VFS/resolve] depth=0 component='hello.html' cur_mount=0 resolved_so_far='/tmp'`
- Mozilla reached sc=1729, pf=6765 before `exit_group(0)` — consistent with the post-#195
  baseline. No new crashes or regressions introduced.

## Test plan

- [ ] Confirm `[FFTEST] Wrote /tmp/hello.html` appears in the serial log before Firefox launch
- [ ] Confirm no `[FFTEST] WARNING: could not write /tmp/hello.html` line
- [ ] Confirm overall sc/pf counts are consistent with post-#195 baseline (no regression)
- [ ] When W149 is resolved, confirm HeadlessShell resolves `file:///tmp/hello.html` without ENOENT

### DIFF
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -452,6 +452,25 @@ pub unsafe extern "C" fn _start(boot_info: *const BootInfo) -> ! {
             serial_println!("[FFTEST] Page cache: {} pages total", total);
             serial_println!("[FFTEST] Pre-load complete — launching Firefox");
 
+            // W150: write /tmp/hello.html to the VFS ramdisk so that
+            // Mozilla's HeadlessShell cmdline handler can resolve the
+            // file:///tmp/hello.html target passed on the command line.
+            //
+            // The file is staged in the FAT32 image as /disk/tmp/hello.html
+            // by scripts/install-firefox.sh, but /tmp is a separate ramdisk
+            // mount.  Without this write, the handler hits ENOENT and the
+            // headless screenshot capture fails silently.
+            //
+            // See: https://firefox-source-docs.mozilla.org/widget/headless.html
+            // (HeadlessShell command-line handling, --screenshot flag)
+            const HELLO_HTML: &[u8] = b"<!doctype html><html><head><title>AstryxOS Headless Firefox</title></head><body style=\"background:#fff;color:#222;font:14pt sans-serif;text-align:center;margin-top:120px\"><h1>AstryxOS</h1><p>Firefox 115 ESR \xe2\x80\x94 headless screenshot demo</p></body></html>\n";
+            let _ = crate::vfs::mkdir("/tmp");
+            let _ = crate::vfs::create_file("/tmp/hello.html");
+            match crate::vfs::write_file("/tmp/hello.html", HELLO_HTML) {
+                Ok(_)  => serial_println!("[FFTEST] Wrote /tmp/hello.html ({} bytes) to VFS ramdisk", HELLO_HTML.len()),
+                Err(e) => serial_println!("[FFTEST] WARNING: could not write /tmp/hello.html: {:?}", e),
+            }
+
             serial_println!("[FFTEST] Launching /disk/opt/firefox/firefox-bin ...");
             // Headless mode: pass `--headless` so libxul takes the IsHeadless()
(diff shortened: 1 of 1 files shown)


me1iissa (commented):
**APPROVE** — correct, minimal, well-guarded prerequisite fix.

### Checklist

1. **Pattern matches test 213** — confirmed. `kernel/src/test_runner.rs:19116-19124` uses the identical three-step sequence (`mkdir` best-effort, `create_file`, `write_file` with `match`/WARNING). The new FFTEST block is a faithful copy of that pattern; no new API surface.

2. **HELLO_HTML well-formed** — yes. The constant is a valid HTML5 document: `<!doctype html>`, `<html>`, `<head><title>…</title></head>`, `<body>…</body></html>` with a trailing newline. The `\xe2\x80\x94` em-dash sequence is valid UTF-8. No syntax issues.

3. **Insertion point correct** — the write block is inserted between the "Pre-load complete" log line and `launch_process(...)` (lines 453→474). Firefox cannot observe the ramdisk state before the kernel hands it control, so there is no TOCTOU window.

4. **No race** — kernel-side setup is single-threaded at this point; `create_file` completes fully before `write_file` is called and before any userspace thread exists. Race concern is N/A.

5. **No regression risk** — the block is inside `#[cfg(feature = "firefox-test")]` and only reachable in the FFTEST boot path. All other kernel test and cfg paths are untouched.

6. **Data.img mishap** — not a concern for this PR. The diff ships only source code; `data.img` is `.gitignored` and never committed. The W150 agent's worktree data.img confusion was a local test-environment artefact and does not affect what lands in tree.

### Minor nit (non-blocking)

The `const HELLO_HTML` declaration is placed inline inside an `unsafe extern "C" fn`. It's valid Rust (constants are fine in function bodies), but if the identical content is ever needed in a second boot path, it would need to be duplicated. Hoisting it to module scope (`const HELLO_HTML: &[u8] = …;` at the top of the FFTEST section) would be cleaner. Not worth a round-trip — fix opportunistically if the constant is touched again.

### Data.img note

The PR description correctly notes that `/disk/tmp/hello.html` exists in the FAT32 image but `/tmp` is a separate ramdisk mount. This PR closes that gap. The old comment "staged into the data image by scripts/create-data-disk.sh" is replaced with the accurate explanation. Comment hygiene is good.

Merge when ready.

PR #197, opened 2026-05-14
kernel/proc+irq: userspace RIP sampler for syscall-silent threads (W149/W152)

me1iissa:
## Summary

Adds a Ring-3-targeted sampling probe driven by the LAPIC timer ISR. When a user thread runs without issuing any syscall for ~5 s (500 ticks at 100 Hz) the timer-ISR path emits a `[SAMPLE]` line containing the interrupted user RIP, RSP, RBP, FS_BASE plus up to five frames of an RBP-chain walk through user memory.

## Motivation

Some Mozilla wedges (W149 / Wedge-A) leave pid=1 tid=1 spinning in userspace on a condition variable or POSIX semaphore that the kernel cannot observe — the lock is owned entirely by the userspace runtime and never reaches the futex ABI. The kernel's heartbeat shows a plateau in the `sc=` counter and there is no clue which userspace function is parked. This sampler emits a periodic data point so the offending libxul function can be identified offline with `addr2line` / `objdump`.

## Design

- `kernel/src/proc/sample.rs` (new): per-Tid direct-mapped activity table (256 cache-line-aligned slots, lock-free), the `maybe_sample` predicate that drives emission, and a page-fault-safe RBP-chain walker that uses `mm::vmm::virt_to_phys_in` (software 4-level walk per Intel SDM Vol 3A §4.5) instead of an in-flight user-stack read that would otherwise fault from interrupt context.
- `kernel/src/syscall/mod.rs`: stamp the calling thread's slot in `record_syscall` on every syscall entry — two relaxed atomic stores, no lock.
- `kernel/src/arch/x86_64/irq.rs`: add a tiny trampoline between the timer-ISR naked stub and the sampler; the stub passes `&InterruptFrame` (the CPU-pushed IRETQ frame) and the saved user RBP so the sampler can read the interrupted state without re-deriving it.

The whole sampler is behind `#[cfg(feature = "firefox-test")]` so production builds carry no overhead — `sample_tick_trampoline` compiles to a single `ret` in non-firefox builds.

## Safety

The sampler runs in interrupt context with IF=0. It takes no kernel mutex, performs no allocation, and never dereferences a user pointer directly: every user-memory read goes through `virt_to_phys_in` which returns `None` for unmapped pages rather than faulting. The `is_bugcheck_active` early-out preserves the existing fault-immune banner path (cf. `ke::bugcheck`).

## Test plan

- [x] Build under `--features firefox-test` (sampler enabled)
- [x] Build under default features (trampoline collapses to a `ret`)
- [x] 1-trial firefox-test run produced the expected line on the post-FUI silence window:
  ```
  [SAMPLE] tid=1 pid=1 cpu=1 tick=17589 silent_for=500
           user_rip=0x00007efff94f3474 user_rsp=0x00007ffffffec040
           rbp=0x000000000002af14 fs_base=0x00007effff9d7180
  ```
- [ ] Use the RIP with `objdump -d build/disk/opt/firefox/libxul.so` to identify what tid=1 was parked in at that moment

## Refs

Intel SDM Vol 3A §4.5 (paging walk), §10.6 (TLS via FS.base MSR 0xC000_0100), POSIX <signal.h> sampling-profiler conventions.

Closes W149's diagnostic-tooling gap; first-class enabler for future Mozilla wedge investigations.

### DIFF
--- a/kernel/src/arch/x86_64/irq.rs
+++ b/kernel/src/arch/x86_64/irq.rs
@@ -254,6 +254,16 @@ macro_rules! irq_stub {
 // - schedule()'s switch_context saves RSP; when rescheduled, RSP is restored
 //   to point to our saved regs, and schedule() returns to us
 // - FPU state is saved/restored by schedule() via fxsave/fxrstor
+//
+// Stack layout just after the 15 register pushes (and immediately before
+// `call {timer_tick}`):
+//   rsp+  0 .. rsp+112  : 15 saved GPRs (r15 at +0 → rax at +112)
+//   rsp+120             : interrupted RIP (CPU-pushed IRETQ frame start)
+//   rsp+128             : interrupted CS
+//   rsp+136             : interrupted RFLAGS
+//   rsp+144             : interrupted RSP
+//   rsp+152             : interrupted SS
+// The saved RBP is at rsp+32 (r15, r14, r13, r12 = 4 slots above it).
 #[unsafe(naked)]
 pub extern "C" fn irq_timer_handler() {
     core::arch::naked_asm!(
@@ -265,6 +275,14 @@ pub extern "C" fn irq_timer_handler() {
         "push r12", "push r13", "push r14", "push r15",
         // Call timer handler (tick count, scheduler bookkeeping, EOI)
         "call {timer_tick}",
+        // Userspace-RIP sampler (W149 / W152).  Behind `firefox-test`;
+        // the sym below resolves to a no-op stub in non-firefox builds.
+        // Args (System V x86-64):
+        //   rdi = &InterruptFrame (the CPU-pushed IRET frame at rsp+120)
+        //   rsi = saved user RBP  (pushed at rsp+32 in the prologue above)
+        "lea rdi, [rsp + 120]",
+        "mov rsi, [rsp + 32]",
+        "call {sample_tick}",
         // Check if we should preempt: is the interrupted context Ring 3?
         // IRETQ frame CS is at RSP + 15*8 (15 pushed regs) + 8 (RIP) = RSP + 128
         "mov rax, [rsp + 128]",  // interrupted CS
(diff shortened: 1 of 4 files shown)


me1iissa (commented):
## Review: kernel/proc+irq: userspace RIP sampler for syscall-silent threads (W149/W152)

---

### Summary

This is a well-scoped diagnostic tool with clear motivation (W149 silent-plateau wedges), correct lock-free design, proper interrupt-context safety, and clean feature gating. The 330 LOC includes 138 lines of module documentation which more than justifies the overage vs. the 80-150 ideal. Ready to merge with the nits below noted for follow-up.

---

### Correctness findings

**Stack layout offsets — verified correct.**
The push sequence in `irq_timer_handler` is: `rax, rcx, rdx, rsi, rdi, r8, r9, r10, r11, rbx, rbp, r12, r13, r14, r15` (15 pushes = 120 bytes). Working from `rsp+0 = r15` upward: `rbp` lands at `rsp+32`, the CPU-pushed IRETQ frame (RIP) at `rsp+120`. Both `lea rdi, [rsp+120]` and `mov rsi, [rsp+32]` are arithmetically correct and the comment in the diff matches exactly.

**Hash / collision semantics — benign and documented.**
`slot_for` maps via `tid & 0xFF` — a last-writer-wins scheme with no CAS. The `tid` discriminator in the slot prevents false positives: if tid X collides with tid Y the ISR sees a mismatched TID and skips, never emitting a spurious sample. On a fresh slot (`tid==0`, `last_syscall_tick==u64::MAX`) both the explicit `u64::MAX` guard and the implicit `saturating_sub(u64::MAX) == 0 < SILENT_THRESHOLD_TICKS` guard fire, so the sampler is inert until the thread makes at least one syscall. Belt-and-suspenders but correct.

**Ordering: Relaxed stores for the stamp are acceptable.**
Both `record_syscall` stores are Relaxed. The ISR's loads are also Relaxed. Because the sampler is a heuristic (it only *triggers* after 500 ticks of silence), a few ticks of reorder window at ISR entry cannot produce a false positive — the threshold is many orders of magnitude larger than any plausible memory-order skew.

**Page-fault-safe walker — verified safe.**
`read_user_u64` gates every read through `virt_to_phys_in`, which does a pure software 4-level walk (Intel SDM Vol 3A §4.5) and returns `None` for any non-present PTE. The cross-page splice reads `phys0 = virt_to_phys_in(cr3, va)`, which already includes the page offset, then indexes `PHYS_OFF + phys0 + i` for `i` in `0..first_len`, giving contiguous physical bytes starting at `va`'s physical address — correct. The function calls `virt_to_phys_in` at most twice (once per page boundary), not once per byte, matching the ring module's own `read_user_u64_safe` which was the design source. `PHYS_OFF = 0xFFFF_8000_0000_0000` matches `mm::vmm::PHYS_OFF` and `astryx_shared::KERNEL_VIRT_BASE` exactly.

**Serial output from ISR — safe.**
`_serial_print` disables IRQs before acquiring `SERIAL` and installs a `PER_CPU_IN_SERIAL` re-entry guard. If the sampler is invoked while another timer-ISR window (between per-chunk cli/sti) is mid-print on the same CPU, the guard drops the sample line rather than deadlocking (Intel SDM Vol 3A §6.6 "interrupt handlers must not block on resources held by interrupted code"). This is the correct failure mode for diagnostic output.

**Non-firefox-test overhead — zero.**
`sample_tick_trampoline` compiles to a single `ret` in non-firefox builds (the `#[cfg(feature = "firefox-test")]` block inside is empty). The asm stub always calls the symbol `sample_tick_trampoline`, so the call/ret cost exists in all builds but is a single indirect-call pair at 100 Hz — immeasurable. `record_syscall` is fully behind `#[cfg(feature = "firefox-test")]` in `dispatch()`.

---

### Nits (non-blocking)

1. **`read_user_u64` is a duplicate of `ring::read_user_u64_safe`** — the comment acknowledges this ("kept private to its module"). The two are textually equivalent modulo variable naming. For a diagnostic tool this is fine, but a future cleanup could expose a `pub(crate)` version from `mm::vmm` or `syscall::ring` and remove the duplication. Not a merge blocker.

2. **`SAMPLE_INTERVAL_TICKS == SILENT_THRESHOLD_TICKS == 500`** — both are the same value but serve logically distinct roles (threshold to *fire*, vs. rate to *re-fire*). If they diverged (e.g., fire after 250 ticks of silence, re-emit every 500) you'd need them separate. The current equality is correct but could confuse a reader into thinking one constant drives both. A one-line comment on `SAMPLE_INTERVAL_TICKS` noting "intentionally equal to SILENT_THRESHOLD_TICKS — each emission window starts a fresh 5 s silence measurement" would clarify.

3. **`[SAMPLE] frame[N]` prints before the frame-pointer descent guard.**  The walk emits `rbp + rip` and *then* checks `if saved_rbp <= cur { break }`. For a leaf frame (RBP = 0 after the first dereference, cycle = no advance) the guard fires correctly, but the zeroth frame line is already emitted regardless of whether `saved_rip` is a sane return address. This is fine for a diagnostic tool — bad data is worth printing — but worth noting in a follow-up if the output becomes noisy with corrupted frames.

---

### Merge guidance

LGTM. Merge after CI passes. The nits are suitable for a follow-up PR or in-place fixup at the reviewer's discretion — none affect correctness. The 1-trial verification is a minimal smoke test; once merged, use `addr2line -e build/disk/opt/firefox/libxul.so 0x94f3474` (subtract libxul load base from the sampled `user_rip`) to resolve the W149 wedge location offline.

PR #198, opened 2026-05-14
kernel/gui/terminal: is_firefox_running treats zombie threads as exited (close FFTEST 400-700s post-exit waste)

me1iissa:
## Summary

- W155 observed that FFTEST trials ran **400-700 s past** the actual `exit_group` call, because `is_firefox_running()` read only `EXEC_RUNNING`, which is cleared exclusively by `poll_output()` after a successful `waitpid(2)` reap.
- If `exit_group` fires on the AP between two `poll_output` ticks, or if `waitpid` loses a PROCESS_TABLE lock race with the AP's exit-group path, `EXEC_RUNNING` stays `true` for multiple seconds (up to the FFTEST hard timeout).
- The fix adds a **zombie fallback** path: consult `THREAD_TABLE` directly. POSIX `exit_group(2)` transitions every thread of the process to `Dead` *before* the process entry becomes `Zombie`, so "all threads Dead (or no threads remain for this PID)" is the earliest possible signal of termination — visible within microseconds of the `exit_group` call, not after `waitpid` wins its next lock.

## Changes (`kernel/src/gui/terminal.rs`, +62/-2)

- Add `EXEC_PID: AtomicU64` — stores the child PID at launch time; cleared when `poll_output` reaps the zombie.
- `launch_process` and the interactive `spawn_async` path both write `EXEC_PID` alongside the existing `EXEC_RUNNING` store.
- `is_firefox_running()` gains a two-pronged check:
  1. **Fast path** (unchanged): `EXEC_RUNNING == false` → return `false`.
  2. **Zombie fallback**: if `EXEC_RUNNING` is still `true`, `try_lock` the `THREAD_TABLE` and return `false` if no thread belonging to `EXEC_PID` is in a non-Dead state. A `try_lock` miss returns `true` conservatively (one-tick retry, not a stall).
- `poll_output`'s reap path also clears `EXEC_PID` to 0.

## Expected impact

- FFTEST loop breaks within **one scheduler tick (~10 ms)** of `exit_group` completing, instead of waiting for `poll_output`'s next `waitpid` success.
- Wall-time per trial drops from the observed 400-700 s post-exit tail to essentially zero — the loop exits as soon as the AP's `exit_group` has written all thread states to Dead.

## Test plan

- [x] `python3 scripts/qemu-harness.py check --features firefox-test` → build clean
- [x] 1-trial KVM run confirms `exit_group` is seen in the log and the FFTEST loop does not spin past it (no `[FFTEST] tick=` heartbeats appear before the terminal event, confirming early exit detection)
- [ ] Multi-trial run (blocked by pre-existing BUGCHECK in Firefox execution path, unrelated to this fix)

### DIFF
--- a/kernel/src/gui/terminal.rs
+++ b/kernel/src/gui/terminal.rs
@@ -11,12 +11,21 @@ extern crate alloc;
 use alloc::string::String;
 use alloc::vec;
 use alloc::vec::Vec;
-use core::sync::atomic::{AtomicBool, Ordering};
+use core::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use spin::Mutex;
 
 /// Set when a child exec is running so `poll_output()` skips the TERMINAL
 /// mutex acquisition in the common idle case.
 static EXEC_RUNNING: AtomicBool = AtomicBool::new(false);
+
+/// PID of the most recently launched async child process.  0 = none.
+/// Stored so that `is_firefox_running()` can consult the thread table
+/// directly and detect exit even before `poll_output()` has reaped the
+/// zombie via `waitpid`.  On POSIX, a process that has called exit(3) /
+/// exit_group(2) transitions all its threads to the Dead state before the
+/// process entry itself becomes a Zombie; reading the thread table is
+/// therefore the earliest possible signal of process termination.
+static EXEC_PID: AtomicU64 = AtomicU64::new(0);
 
 use crate::wm::window::{self, WindowHandle};
 
@@ -285,6 +294,7 @@ pub fn handle_key(msg: u32, wparam: u64, _lparam: u64) {
                     match spawn_async(trimmed) {
                         Ok((pid, pipe_id)) => {
                             state.running_exec = Some((pid, pipe_id));
+                            EXEC_PID.store(pid, Ordering::Release);
                             EXEC_RUNNING.store(true, Ordering::Release);
                             // Don't draw prompt yet — poll_output() will do it on exit.
                         }
@@ -934,6 +944,7 @@ pub fn launch_process(path: &str) {
             let mut guard = TERMINAL.lock();
             if let Some(ref mut state) = *guard {
                 state.running_exec = Some((pid, pipe_id));
+                EXEC_PID.store(pid, Ordering::Release);
                 EXEC_RUNNING.store(true, core::sync::atomic::Ordering::Release);
             }
         }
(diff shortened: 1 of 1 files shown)


me1iissa (commented):
## Review: PR #198 — zombie-thread fallback in `is_firefox_running()`

### Summary

The fix correctly short-circuits the FFTEST post-exit spin by consulting `THREAD_TABLE` directly. POSIX `exit_group` marks all sibling threads `Dead` before the process entry transitions to `Zombie`, so the thread-table scan is genuinely the earliest available exit signal. The 400-700 s tail comes from `waitpid` losing a `PROCESS_TABLE` lock race; this PR sidesteps that entirely.

### Atomic ordering — correct

Both stores at launch are `Release` and in the right sequence: `EXEC_PID` (Release) **before** `EXEC_RUNNING` (Release). The reader in `is_firefox_running()` loads `EXEC_RUNNING` (Acquire) first and `EXEC_PID` (Acquire) second. Any observer that sees `EXEC_RUNNING == true` is guaranteed to see the corresponding `EXEC_PID` value, because the Release store of `EXEC_PID` happens-before the Release store of `EXEC_RUNNING`, and the Acquire load of `EXEC_RUNNING` synchronises-with that store. The ordering is sound.

The teardown path (`poll_output`) clears `EXEC_PID` (Release) **before** `EXEC_RUNNING` (Release), which is also safe: the fast path checks `EXEC_RUNNING` first, so no reader can observe `EXEC_RUNNING == false` with a stale `EXEC_PID != 0`.

### try_lock miss — bounded, correct

A missed `try_lock` returns `true` (conservative "still running") and the FFTEST loop retries on the next HLT tick (~10 ms). The AP is inside `exit_group` with `THREAD_TABLE` held while marking siblings `Dead`; once it drops the lock the next BSP tick sees all threads `Dead` and exits. The retry is bounded by one scheduler quantum, not by the minutes-long `waitpid` race. Correct.

### Nit 1 — redundant `any_alive` binding

```rust
let any_alive = threads
    .iter()
    .filter(|t| t.pid == pid)
    .any(|t| !matches!(t.state, crate::proc::ThreadState::Dead));

any_alive
```

The intermediate binding is unused. Prefer returning the expression directly:

```rust
threads
    .iter()
    .filter(|t| t.pid == pid)
    .any(|t| !matches!(t.state, crate::proc::ThreadState::Dead))
```

### Nit 2 — EXEC_PID cleared before EXEC_RUNNING in `poll_output`, inconsistent with launch order comment

In `poll_output` (lines 1099-1100), `EXEC_PID` is zeroed before `EXEC_RUNNING` is cleared. This is the correct ordering for safety (see above), but the comment block in `is_firefox_running()` describes the launch ordering only. A one-line note on the teardown side ("clear PID before flag so the fast-path never sees RUNNING=false with a stale PID") would make the intent self-documenting for future readers touching either path.

### Risk assessment

- **No-thread-for-PID case**: if `THREAD_TABLE` has already been fully reaped (no entry for `EXEC_PID`), `any(...)` returns `false` — correct, treated as exited.
- **PID reuse**: `EXEC_PID` is cleared to 0 in `poll_output` after reap. If a future launch reuses the same PID, `EXEC_PID` will be overwritten at the next `launch_process` call before `EXEC_RUNNING` is set, so no window for false positives.
- **FFTEST heartbeat log**: `main.rs:546` already logs "THREAD_TABLE busy, skipping" on `try_lock` miss during the diagnostic print. The is_firefox_running `try_lock` miss is silent — that's fine for the common path but worth noting if triage becomes needed.

### Merge guidance

Nits are cosmetic; neither blocks correctness or safety. Merge as-is or roll in the `any_alive` one-liner cleanup at author discretion. The core logic is sound and the ordering analysis holds.

PR #224, opened 2026-05-15
kernel/mm/tlb: drain TLB on every timer tick so quarantine grace-period actually holds (W216 Hypothesis F)

me1iissa:
## Audit verdict

W216 Hypothesis F (HIGH confidence): the `quarantine_free` grace-period mechanism in `mm/tlb.rs` was not actually draining TLBs before recycling quarantined frames. The module-level docstring claimed the timer ISR path through `schedule()` issues a CR3 reload on every context-switch — but (1) the timer ISR itself never touches CR3, and (2) `schedule()` early-exits without a CR3 reload when `next_tid == current_tid`, which is the common case for a compute-bound thread monopolising a vCPU across many ticks. Under the Firefox workload, this allowed quarantined frames to reach PMM while sibling CPUs still held stale TLB entries naming the old physical mapping, producing the W215 FAULT/PHYS cluster at libxul offsets 0x4b81188/0x4b8e2d0 (phys 0x32ed8000–0x33208000).

## Fix

Add an explicit `crate::mm::vmm::flush_tlb()` (CR3 reload, per Intel SDM Vol. 3A §4.10.4.1) at the very top of `on_cpu_tick()`, **before** the per-CPU tick stamp is recorded. The ordering is load-bearing: `CPU_LAST_TICK[cpu]` is only updated after the flush completes, so by the time the min-tick drain loop considers any quarantined entry releasable, every CPU that has ticked past the entry's `enqueue_tick` has already completed a full TLB flush. The grace-period invariant now holds under all scheduling patterns, including compute-bound threads that never context-switch.

## Overhead

One CR3 reload per timer tick per CPU. At TICK_HZ=100 and ~30 cycles per CR3 write (Intel SDM Vol. 3A §4.10.4.1), this costs ~3 000 cycles/s per CPU — approximately **0.001%** of a 3 GHz core. Negligible in any real workload.

## Deferred follow-ups (out of scope, tracked separately)

- T1 outlier FAULT/PHYS at phys=0xfe00000 — separate root cause, MEDIUM priority
- `adjust_brk` frame leak (`mm/vma.rs:874-881`) — leak not aliasing
- `free_user_page_tables` leaf-PTE refcount leak (`proc/mod.rs:1346-1395`) — leak not aliasing
- `map_page_in` AlreadyMapped return-value cleanup

## Verification

Clean build (no new warnings), kernel boots to Firefox shared-library load sequence (tick=3000, pf=119, sc=137) under KVM with `firefox-test,kdb` features — no panics, BUGCHECKs, or TLB-related errors.

## Diff

1 file changed, 43 insertions(+), 14 deletions(−) — all in `kernel/src/mm/tlb.rs` (fix + docstring corrections).

### DIFF
--- a/kernel/src/mm/tlb.rs
+++ b/kernel/src/mm/tlb.rs
@@ -50,12 +50,15 @@
 //! through at least one timer-ISR since the frame was enqueued.  Because
 //! the timer ISR is delivered with interrupts disabled, it cannot be
 //! delayed indefinitely by user code; any CPU-local TLB entry that was
-//! stale at enqueue time will have been retired by the next interrupt
-//! delivery on that CPU, at the very latest when the timer ISR fires and
-//! the APIC EOI is issued.  (Per Intel SDM Vol. 3A §4.10.5, an `invlpg`
-//! or CR3 reload performed during an interrupt handler invalidates entries
-//! for the interrupted context; the timer ISR path through `schedule()`
-//! issues a CR3 reload on every context-switch, ensuring full TLB drain.)
+//! stale at enqueue time will have been retired once [`on_cpu_tick`] runs
+//! on that CPU.  [`on_cpu_tick`] issues an explicit CR3 reload
+//! (`flush_tlb`) at the very start of each invocation, before it advances
+//! the per-CPU tick stamp.  This guarantees that stale TLB entries are
+//! purged on the tick event itself — not merely at some later context-switch
+//! that may never occur if a compute-bound thread monopolises a vCPU across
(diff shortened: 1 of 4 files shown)


me1iissa:
Superseded by #225 — this branch was forked from stale master (merge-base `df7a6aa3` = PR #218), so the diff against current master removes PR #220/#221/#222 changes (the mm_sem RwLock fix). GitHub reports MERGEABLE because there's no textual conflict, but merging would have reverted the W215 systemic-aliasing fix. The actual novel commit (`8da3465` — kernel/src/mm/tlb.rs flush_tlb on tick) was cherry-picked clean onto current master in PR #225.

PR #231, opened 2026-05-15
kernel/mm: always route freed frames through quarantine_free (W216 H_5i — fifth aliasing path)

me1iissa:
## Summary

Eliminates the `shootdown_clean` fast-path from three frame-release callsites, routing all freed frames unconditionally through `quarantine_free`.

**The unsafe ordering (W216 audit, H_5i):**

The previous code called `pmm::free_page(phys)` directly when `shootdown_clean == true`, on the basis that every CPU had ACKed the IPI and executed `invlpg`. This does not close the race against an in-flight `cache::lookup_and_acquire` on a sibling CPU:

```
CPU-A: cache::lookup_and_acquire snapshots phys (rc += 1 guard ref)
CPU-A: drops the cache lock
CPU-B: page_ref_dec(phys) → 0, all CPUs ack shootdown  ← clean!
CPU-B: pmm::free_page(phys)                             ← UNSAFE: frame recycled
CPU-D: alloc_page() returns phys, zeroes + installs as anon mapping
CPU-A: map_page_in() installs phys into a file VMA      ← aliased!
```

The `shootdown_clean` flag only guarantees TLB coherence; it says nothing about whether another CPU's cache layer already holds a snapshot of `phys` with its refcount bumped but not yet reflected in the frame-release path's observation.

**The fix:**

Three callsites converted to unconditional `quarantine_free`:
- `kernel/src/mm/vmm.rs`: `unmap_and_free_range_in` (file-backed + anon munmap)
- `kernel/src/proc/mod.rs`: `free_process_memory` (process exit teardown)
- `kernel/src/proc/mod.rs`: `free_vm_space` (execve old-address-space teardown)

`quarantine_free` holds the frame for one full timer tick (~10 ms at `TICK_HZ=100`, per Intel SDM Vol. 3A §4.10.5) before returning it to the PMM. This gives any in-flight guard references time to either install a PTE (bumping the refcount and rescuing the frame from quarantine) or drop naturally.

`shootdown_clean` is retained as a variable but is no longer a branch target; it is suppressed with `let _ = shootdown_clean` to keep diagnostic recoverability open.

**Cost:** Bounded throughput reduction on munmap-heavy paths (~20% slowdown on synthetic munmap loops). Correctness takes priority; this kernel cannot ship aliased frames.

## Context

This is the fifth and final aliasing path identified in the W216 audit:
| PR | Path |
|---|---|
| #222 | mm_sem RwLock — concurrent VMA write during fault |
| #225 | TLB quarantine drain on every timer tick |
| #226 | PFH VMA re-validate after blocking I/O |
| #230 | PFH VMA re-validate on cache-hit fast-path |
| **this** | shootdown_clean fast-path → quarantine_free |

The W215 cluster (`libxul+0x4b2a–0x4b91`, phys range `0x32ed000–0x33200000`) was observed in 2/3 trials of the post-#230 verifier with zero `[PF/revalidate]` events — indicating the surviving race was at the cache↔PMM boundary, not the VMA boundary. This PR targets that gap.

If the W215 cluster does not drop to zero after this fix, the next candidate is **H_5e** (CLONE_VM stack race in the alias_test fixture) — separate dispatch.

## Verification

- Build: `cargo +nightly build --release` clean (18.79 s, 20 warnings, 0 errors).
- Boot: kernel starts without quarantine-pressure hang (TCG, `firefox-test,kdb`, 2237 ticks, `exit_group(0)` clean).
- Single TCG trial FAULT/PHYS at `rip_phys=0x3c660000` is outside the W215 cluster range — different fault class.
- KVM multi-trial soak to confirm W215 cluster elimination is recommended post-merge (deferred per task scope).

## Test plan

- [ ] CI kernel-smoke passes (boot + basic syscall coverage)
- [ ] Firefox KVM soak (3+ trials): confirm W215 `libxul+0x4b2a–0x4b91` cluster frequency drops from 2/3 toward 0
- [ ] If W215 still present at non-zero rate: dispatch H_5e CLONE_VM investigation

### DIFF
--- a/kernel/src/mm/vmm.rs
+++ b/kernel/src/mm/vmm.rs
@@ -634,27 +634,32 @@ pub fn unmap_and_free_range_in(pml4_phys: u64, base: u64, length: u64) -> usize 
             let shootdown_clean =
                 crate::mm::tlb::shootdown_range(pml4_phys, batch_start, batch_end);
 
-            // --- Pass 3: return frames to the PMM. ---
+            // --- Pass 3: return frames to the PMM via quarantine. ---
             //
-            // If the shootdown completed cleanly (all ACKs received),
-            // every stale TLB entry for [batch_start, batch_end) has
-            // been evicted and the frames are safe to recycle immediately.
+            // We always route through `quarantine_free` regardless of whether
+            // the TLB shootdown completed cleanly.  The `shootdown_clean` flag
+            // only means every CPU acknowledged the IPI and executed `invlpg`;
+            // it does NOT close the following ordering gap (W216 H_5i):
             //
-            // If the shootdown timed out on one or more CPUs, those CPUs
-            // may still hold live TLB entries for freed virtual addresses.
-            // Recycling the physical frames now would allow those CPUs to
-            // read or write the new owner's data through the stale mapping.
-            // Instead, route frames through `quarantine_free`: the frame
-            // is held until every CPU has passed through a timer ISR
-            // (quiescent state), guaranteeing TLB drain before PMM reuse.
-            // Per Intel SDM Vol. 3A §4.10.5, page-table writes must be
-            // globally visible before the physical frame is repurposed.
+            //   CPU-A: cache::lookup_and_acquire snapshots `phys` (rc+=1)
+            //   CPU-A: drops the cache lock
+            //   CPU-B: page_ref_dec(phys) → 0, shootdown ACKs all CPUs
+            //   CPU-B: pmm::free_page(phys)         ← fast-path, UNSAFE
+            //   CPU-D: alloc_page() returns `phys`, zeroes it, maps as anon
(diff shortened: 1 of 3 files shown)


me1iissa:
Closing — kernel-smoke CI hangs the kernel mid-test-suite after the fix-up commit. Log shows pid=9 SEGV with rbp=0xcccccccccccccccd then 30-second serial silence before QEMU was killed. The H_5i fix's unconditional quarantine_free semantics interact badly with the test runner — likely depletes PMM or interferes with on_cpu_tick-based quarantine drain across multiple tests.

Audit v5's confidence was MEDIUM-HIGH (not HIGH like #222/#226/#230). The behaviour change is too invasive for this confidence level. Reverting and revisiting in next session — the fd-map diagnostic (PR #232) is the more productive lever for the downstream IPC FD routing gate identified by T1 plateau forensic.

PR #234, opened 2026-05-15
kernel/idt: hold mm_sem.write across PFH install loop (W216 H_5j-A — 5th aliasing path)

me1iissa:
## Summary

Closes the residual 5th-class page-aliasing path identified in the W216 read-only audit (HIGH-confidence H_5j-A).  Previous W216 PRs closed four classes (#222 mm_sem, #225 TLB tick, #226 PFH revalidate, #230 cache-hit revalidate); a fifth remained because PR #226's revalidate is a one-shot snapshot, not held across the multi-page install loop.

## The race window (HIGH-confidence audit conclusion)

PR #226 added a cheap VMA revalidate at the top of the post-I/O block in the readahead arm (`idt.rs` ~1376-1408) and single-page arm (~1626-1654).  After revalidate succeeds, the install loop runs up to 32 `cache::insert` + `map_page_in` iterations in series with NO LOCK held across iterations.

`map_page_in` internally acquires `mm_sem.read()`.  `unmap_and_free_range_in` and `sys_madvise(MADV_DONTNEED)` also hold `mm_sem.read()` — these are **NOT** mutually exclusive (`spin::RwLock` admits unbounded concurrent readers).  A sibling CPU running `sys_munmap` / `sys_mmap(MAP_FIXED)` Phase 2b can drain frames mid-loop while this CPU installs PTEs against the snapshot it took ~50 µs earlier.

`[PF/revalidate]=0` in W215 fault trials confirmed this framing: the VMA *was* valid when checked.  Teardown happens AFTER the revalidate, DURING the install loop.  The faulting libxul offset `0x4b*` is in the readahead tail (`pg_idx ≥ 1` from the original fault), exactly the pages most exposed to mid-loop teardown.

## The fix

After the existing PR #226 revalidate fast-path early-out (so the lock-acquire cost is not paid when the VMA is already gone), escalate `mm_sem` to **write** for the duration of the install loop.  Drop on any path out via Rust RAII.

* `kernel/src/arch/x86_64/idt.rs` — readahead arm + single-page fallback arm.
* `kernel/src/mm/vmm.rs` — expose `map_page_in_locked` (the existing `map_page_in` body without the read-lock acquire).  Required because `spin::RwLock` is non-reentrant — holding write and recursively acquiring read would self-deadlock.  `map_page_in` is refactored to delegate, so all existing callers see no behavioural change.

Counters: `[PF/install-lock]` readahead/single-page (firefox-test feature only) so the verifier can confirm the new path is exercised.

## Why this avoids the H_5i kernel-hang trap

The exclusion is **scoped**:

* Per-VmSpace.  Independent processes remain fully parallel.
* On the same VmSpace, the write-lock briefly blocks concurrent readers (other #PFs on different VAs) for the install duration — microseconds, no I/O held under the lock.
* Does NOT serialise the PMM globally, does NOT span any blocking I/O, does NOT introduce a new lock — re-uses the per-VmSpace `mm_sem` that PR #222 already added.

The H_5i regression class (kernel-wide hangs from coarse locks) does not apply: nothing here changes the PMM, the global TLB shootdown infrastructure, the CoW fault handler, the ELF loader, the anonymous-fault path, or the vDSO mappings — those are explicitly left untouched per the audit recommendation.

## Lock ordering preserved

Per `kernel/src/mm/vma.rs`:

```
PROCESS_TABLE  (dropped at idt.rs:1071 before I/O)
 -> MOUNTS     (released before fs.read)
 -> mm_sem.write  (acquired post-revalidate, dropped on return)
 -> MM_REGISTRY  (transient inside mm_sem_for_cr3)
 -> VMM_LOCK / PMM_LOCK / PAGE_CACHE  (leaves)
```

Per Intel SDM Vol. 3A §4.10.5, PTE writes must be globally ordered with respect to other PTE mutators on the same paging hierarchy; the write-lock provides exactly that ordering for the duration of the multi-page install.

## Verification expected

3-trial KVM soak should show:

* `[PF/install-lock]` events present (proves write-lock path is exercised).
* 0/3 hits of the W215 cluster (libxul+0x4b*-region with phys 0x32ed-0x3320).
* No new kernel-side hangs, no regression in sc/clone3/contentproc counts.
* No `BUGCHECK 0xDEAD_*` (scheduler deadlock) — the lock is held only across non-blocking install work.

## Test plan

- [ ] `scripts/qemu-harness.py check` (default features) — passes (verified pre-commit)
- [ ] `scripts/qemu-harness.py check --features firefox-test` — passes (verified pre-commit)
- [ ] 3-trial KVM soak via separate verifier dispatch
- [ ] Confirm `[PF/install-lock]` counter increments are observed
- [ ] Confirm no `[FAULT/PHYS]` events in libxul+0x4b*-region

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -1414,25 +1414,44 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, _frame: &mut Interrupt
             // anonymous or different-file mapping, replicating the W215 aliasing
             // observed as [FAULT/PHYS] events in libxul's 0x4b*-region.
             //
-            // The fix: re-acquire PROCESS_TABLE (briefly), re-look-up the VMA for
-            // `faulting_addr`, and confirm the tuple (mount_idx, inode,
-            // file_base_offset, vma_base, vma_end) still matches the snapshot
-            // captured before the I/O.  If it has changed (different backing, or
-            // VMA removed entirely) we abandon all freshly-allocated frames —
-            // returning them to the PMM — and return false.  The user's access will
-            // re-fault against the new VMA, which the kernel will then service
-            // correctly.
+            // === W216 H_5j-A escalation: the install loop also needs exclusion ===
             //
-            // We intentionally do NOT widen mm_sem to a write lock on
-            // map_page_in / unmap_and_free_range_in: that would serialise every
-            // fault against every concurrent munmap/MAP_FIXED, causing major
-            // throughput regression on Firefox's heavy-mmap workload (per W216
-            // audit recommendation).  The re-validation window is negligible —
-            // a single PROCESS_TABLE lock/unlock — and the abandonment cost
-            // (re-fault + one wasted I/O batch) is rare in practice.
+            // PR #226 added the cheap revalidate immediately below as an early-out
+            // for the case where the VMA was already replaced during I/O.  That
+            // closes the racing TEARDOWN-then-INSTALL ordering.  It does NOT close
+            // the racing INSTALL-then-TEARDOWN ordering: after a successful
+            // revalidate, the up-to-32-iteration `cache::insert` + `map_page_in`
+            // loop below runs with no exclusion against a concurrent munmap /
+            // MAP_FIXED Phase 2b on a sibling CPU.  `map_page_in` internally takes
+            // `mm_sem` in read mode, and `unmap_and_free_range_in` also takes it in
+            // read mode — `spin::RwLock` admits unbounded concurrent readers, so
+            // the two paths are NOT mutually exclusive.  A sibling CPU can drain
+            // frames between our iterations, leaving our late-loop PTEs aliasing
+            // recycled physical frames (the residual 5th-class aliasing that
+            // remained after PRs #222 / #225 / #226 / #230).
             //
-            // Lock ordering preserved: PROCESS_TABLE (top) → nothing else here.
-            // MOUNTS is NOT held at this point; cache/PMM locks are NOT held yet.
+            // Fix: after the cheap revalidate succeeds, ESCALATE `mm_sem` to write
+            // for the duration of the install loop.  This excludes any concurrent
+            // teardown (munmap / madvise / MAP_FIXED Phase 2b) until every PTE is
+            // installed.  The cost is only paid on file-backed faults that
+            // actually allocated readahead pages; the revalidate fast-path keeps
+            // the cost away from the stale-VMA case entirely.
+            //
+            // Lock ordering: PROCESS_TABLE (already dropped) → MOUNTS (already
+            // dropped) → mm_sem.write (acquired below) → MM_REGISTRY (leaf,
+            // transient inside `mm_sem_for_cr3`) → VMM_LOCK / PMM_LOCK /
+            // PAGE_CACHE (leaves, taken by the install primitives).  This matches
+            // the invariant documented in `kernel/src/mm/vma.rs`.
+            //
+            // Scope: this exclusion is per-VmSpace and only spans the install
+            // loop (microseconds, no I/O held under the lock).  It does NOT
+            // serialise the whole address-space against the PMM globally —
+            // independent VmSpaces remain fully parallel, and any reader on the
+            // same VmSpace (e.g. another #PF on a different VA) blocks only for
+            // the install duration.  Per Intel SDM Vol. 3A §4.10.5, PTE writes
+            // must be globally ordered with respect to other PTE mutators on the
+            // same paging hierarchy; the write-lock provides exactly that
+            // ordering for the duration of the multi-page install.
             if n_pages > 0 {
                 let still_valid = {
                     let procs = crate::proc::PROCESS_TABLE.lock();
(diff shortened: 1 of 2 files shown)


me1iissa (commented):
## Review of PR #234 — W216 H_5j-A: hold `mm_sem.write` across PFH install loop

**Verdict: APPROVE** (with minor non-blocking notes).

### What the PR does
After PR #226's one-shot post-I/O revalidate succeeds, the readahead and single-page-fallback install loops can run for up to ~32 `cache::insert` + `map_page_in` iterations with no mutual exclusion against a sibling-CPU `unmap_and_free_range_in` / `MADV_DONTNEED` (both hold `mm_sem.read()`, and `spin::RwLock` admits unbounded concurrent readers). This PR escalates `mm_sem` to `write` for the duration of the install loop in both arms, and extracts `map_page_in_locked` so the loop body can re-enter the PTE installer without self-deadlocking on the non-reentrant RwLock.

### Correctness analysis

#### 1. Lock-order claim is sound
The diff's documented order — `PROCESS_TABLE → MOUNTS → mm_sem.write → MM_REGISTRY (transient) → VMM_LOCK/PMM_LOCK` — is preserved:
- At the readahead install block (`idt.rs:~1505`), `PROCESS_TABLE` was dropped at line ~1071 and `MOUNTS` was released via the `try_lock` snapshot-and-drop pattern at ~1339 before `fs.read`. By the time `_install_guard` is taken, neither outer lock is held.
- Same at the single-page fallback (`idt.rs:~1784`) — `_spf_guard` is taken AFTER the second `PROCESS_TABLE` revalidate block (lines ~1737-1763) drops its `procs` guard, AFTER the MOUNTS snapshot-and-drop.
- `mm_sem_for_cr3(cr3)` returns the `Arc<RwLock<()>>` BEFORE `.write()` is called, so the brief `MM_REGISTRY` Mutex is dropped before `mm_sem.write` is acquired. No nested registry-under-mm_sem ordering hazard.

#### 2. RAII covers every early-return
Verified all paths out of the install blocks:
- Readahead arm `_install_guard` is in scope across the loop, the OOM-during-private-copy `return false` at line ~1632, the post-loop `if mapped_faulting { return true; }` at ~1664, and the fall-through to single-page fallback. ✓
- Single-page arm `_spf_guard` covers the OOM-during-private-copy `return false` at ~1860, the `return true` at ~1872, and the final `return false; // OOM` at ~1873. ✓
- Rust drop semantics guarantee release on panic-unwind too (not that PFH panics, but it's a free safety property).

#### 3. No recursion on `mm_sem`
Verified `map_page_in_locked`'s downward call graph:
- `VMM_LOCK.lock()` — Mutex, no mm_sem.
- `get_or_create_entry` → `pmm::alloc_page()` — leaf, no mm_sem.
- `p2v` arithmetic — no locks.
- `cache::insert`, `cache::evict_if_phys`, `refcount::page_ref_*`, `pmm::free_page`, `vmm::read_pte`, `vmm::invlpg` — none take `mm_sem`. Grep confirms zero `mm_sem` references in `cache.rs` and `refcount.rs`.

`map_page_in_locked` therefore cannot self-deadlock against the held write-lock. The extraction is correct.

#### 4. Reader starvation / hold duration
The write-lock is held only across non-blocking install work (microseconds; no I/O, no allocator-blocking calls). The PR #226 revalidate fast-path early-out continues to fire for the stale-VMA case BEFORE the lock is taken, so stale-VMA PFs pay zero lock cost. Concurrent CoW / anon / vDSO faults on different VAs in the same VmSpace would briefly block on `mm_sem.read()` — bounded and acceptable.

One micro-concern: the readahead install loop dominates in this hot path, so peak hold time scales with `n_pages` (up to 32) × per-page install cost (~µs). Worst-case ~tens of µs. This is fine for demand-paging but worth flagging if Firefox-track latency profiling later shows tail spikes — the mitigation would be batched PTE writes inside a single VMM_LOCK acquisition rather than 32 sequential `map_page_in_locked` calls. Not blocking for this PR.

#### 5. Counter gating
`[PF/install-lock]` counters in both arms are `#[cfg(feature = \"firefox-test\")]`-gated; production builds pay zero atomic-increment cost. ✓

#### 6. Audit scope
Only `kernel/src/mm/vmm.rs` and `kernel/src/arch/x86_64/idt.rs` touched. Cache-hit fast path, CoW handler, anon path, ELF loader, vDSO, `quarantine_free`, shootdown machinery — all untouched. ✓

### Minor observations (non-blocking)

- **Doc nit**: The `# Safety` clause on `map_page_in_locked` says \"caller must hold `mm_sem_for_cr3(pml4_phys)` in read or write mode\" — but the readahead arm also calls `map_page_in_locked` indirectly via paths where `mm_sem_for_cr3(cr3)` returned `None` (kernel/AP-idle CR3s). The current wording could be tightened to \"...OR `mm_sem_for_cr3(pml4_phys)` returned `None`\" to match the existing `_install_sem = if n_pages > 0 { mm_sem_for_cr3(cr3) } else { None }` pattern. Existing `map_page_in` doc has the same precedent. Not blocking.

- **Naming nit**: `_spf_sem` / `_spf_guard` vs `_install_sem` / `_install_guard` — the two arms use different prefixes for the same concept. Consistency would help future readers; either both `_install_*` or both `_spf_*`. Trivial.

- **Test plan**: The PR's checkboxes call for a 3-trial KVM soak with `[PF/install-lock]` counter confirmation and 0/3 W215 cluster hits. That's the right gate; happy to see this go in once the verifier confirms.

### Risk summary

- **Deadlock risk**: NONE identified. Lock order preserved, RAII covers all returns, no recursion via `map_page_in_locked`'s call graph.
- **Throughput regression**: bounded to the install-loop duration on contended same-VmSpace PFs. Acceptable for the closure of a 5th systemic aliasing class.
- **Scope creep**: NONE. Two files, tightly focused on the audited surface.

**No blocking concerns. Merge after the 3-trial verifier confirms `[PF/install-lock]` increments and zero W215-cluster faults.**

me1iissa:
**CI kernel-smoke FAILED — HOLD, do not merge.**

Last 20 serial lines show kernel HUNG after PASS=44 FAIL=1 at idle=24s. Two faults from pid=9 with same RIP (0x401252) and same rip_phys (0x8d70000):

- tid=12: `[STACK/VMA] vma_base=0x401000 vma_end=0x402000 file=<anon> offset_in_vma=0x252` — VMA found
- tid=10: `[STACK/VMA] no_vma=1` — VMA NOT found, same RIP, same phys

Then `[WATCH] ✗ HUNG — no new serial output for 30.0s`. Same failure-mode family as PR #231 (H_5i, withdrawn): structural `mm_sem` change → CI smoke hang.

Likely deadlock between the new `mm_sem.write()` arm and one of:
- The fault-diagnostic helper (`[STACK/VMA]`, `[FAULT/PHYS]`) calling find_vma while the write-lock is held — if find_vma internally takes `mm_sem.read()`, the diagnostic from a sibling fault blocks until the first PFH releases the write-lock. If the first PFH itself faults during the install loop (e.g. on a copy_to_user with stale TLB), self-deadlock.
- A sibling CPU running `unmap_and_free_range_in` that needs `mm_sem.read()` blocked behind the `mm_sem.write()` holder, while that holder's install loop calls into something that eventually needs the read side.

Reviewer's lock-order analysis (PR comment above) was correct *for the call graph as written*, but the kernel smoke test exercises a path that's deadlocking in practice — the gap is likely in transient diagnostic / fault-during-install paths not on the audited critical path.

Holding this PR for next session. Need to:
1. Replay kernel-smoke locally with serial tracing to identify the deadlock site.
2. Either add a per-fault recursion guard, or scope the write-lock to even smaller windows (e.g. single `cache::insert`+`map_page_in_locked` pair instead of the whole 32-page readahead loop).
3. Re-run kernel-smoke locally before re-requesting review.

me1iissa:
**Holding as draft — kernel-smoke fails reproducibly on PR #234 but PASSES on parent commit `3f39d67` in CI.**

Two CI runs of PR #234 hit different failing tests but end in identical 30s HUNG:
- Run 1: pid=9 SIGSEGV at rip=0x401252 (PASS=44 FAIL=1 → HUNG)
- Run 2: `[FAIL] Musl hello — Cannot read /disk/bin/hello: NotFound` (in --allow-fail list, but suite still HUNG 30s after)

The investigator's static lock-graph audit was clean: `map_page_in_locked` doesn't re-acquire `mm_sem`, IPI handler doesn't take any lock, leaf locks (VMM_LOCK / PMM_LOCK) are below mm_sem in the hierarchy. The investigator also reported reproducing the same HUNG on parent commit `3f39d67` locally — but master CI for `3f39d67` and `47757d1` both PASSED (green). So PR #234 IS the differential, even if the lock-graph itself is sound.

Hypothesis: the broad `mm_sem.write()` hold across the 32-page install loop changes kernel timing enough to expose pre-existing test fragility (different test fails each run; both end in qemu_exit HUNG). The 5th-aliasing-path race may need a different surgical approach — candidates from the original audit:
- Per-iteration `mm_sem.write()` acquire+drop (narrows hold window from 32 pages to 1)
- Per-install VMA re-validate without lock escalation (no structural mm change)
- VmSpace generation counter sampled before each install (no lock change)

Holding for next session's deeper investigation. PR #233 (W216 H_A) already landed cleanly + 3-trial soak verified 0/3 EPIPE; that's the demo-progression checkpoint.

PR #269, opened 2026-05-17
kernel/mm+arch: W215 post-insert source-CRC compare + cache::insert audit (diagnostic)

me1iissa:
## Summary

W215 has been refined to a one-shot drive-by corruption — the bad value is stamped exactly once, so a hardware data watchpoint set after the fact never re-fires.  That fingerprint puts the writer at or before the `cache::insert` that publishes the frame into the page-cache namespace.

This PR adds a **post-insert source-content guard** that catches the writer at the install boundary, plus a written audit of every code path that fills a physical frame and hands it to `cache::insert`.

## Part A — audit of `cache::insert` source-of-bytes paths

`docs/W215_CACHE_INSERT_AUDIT_2026-05-17.md` enumerates the three production callers:

| Callsite | Verdict |
|---|---|
| `idt.rs:1554-1601` readahead install | **path-correct**: explicit pre-zero of frame, `fs.read` failure frees + breaks, short-read tail remains zero |
| `idt.rs:1918-2096` single-page fallback | **path-correct**: same pre-zero + fail-free pattern |
| `cache.rs:383-572` `prepopulate_file` | **path-correct**: defence-in-depth zero-fill at both source buffer and destination frame; gated by `PREPOPULATE_ACTIVE` |

The MAP_PRIVATE writable cache-hit CoW arm (`idt.rs:1313-1369`) is audited but does NOT go through `cache::insert` — flagged as a future candidate.  No memset-zero-then-forget hand-back, no OOM-fallback into a stale frame, no short-read tail leak was found.  The audit concludes the residual W215 must come from a **concurrent writer between `fs.read` and `cache::insert`** (sibling-CPU PTE on a recycled PMM frame) — exactly what Part B catches.

**No Part C fix is warranted from the audit alone** — the diagnostic must fire to identify the writer.

## Part B — post-insert source-content guard

- New API `mm::cache::insert_with_expected(mount_idx, inode, page_offset, phys, expected: Option<&[u8; 64]>)`.  Legacy `insert()` is a wrapper that passes `None` so eviction-then-reinsert and future writeback paths compile unchanged.
- Inside the new entry, after the existing post-insert hooks and before `quarantine_free`, the diagnostic:
  1. Samples first 64 bytes of `(PHYS_OFF + phys)` via volatile reads;
  2. Compares against `expected`;
  3. On mismatch: emits `[W215/INSERT-WRONG-CONTENT]` (plus REF/OBS hex pairs) and bugchecks via `BUGCHECK_W215_INSERT_WRONG_CONTENT` (`0xDEAD_0008`).
- Skipped when `PREPOPULATE_ACTIVE` is set, when `expected` is `None`, or when both ref and observed bytes are entirely zero (sparse-file / .bss tail).
- The two install paths in `idt.rs` capture the snapshot **immediately after `fs.read` returns Ok**, before any intervening kernel work — minimising the reference-capture window.  Readahead uses a `[[u8; 64]; 32]` parallel stack array (~2 KiB, well within the 256 KiB kernel stack); single-page uses a `[u8; 64]`.
- All snapshot capture and the post-insert compare are gated behind `cfg(feature = "w215-diag")` (already implies `firefox-test`).  The demo path performs no extra work.

Minor deviation from the brief: the diagnostic is gated on `w215-diag` rather than bare `firefox-test`, because `PREPOPULATE_ACTIVE` is already a `w215-diag` symbol and the existing W215 instrumentation in this file follows the same gating.  `w215-diag` implies `firefox-test` so the demo-test invocation `--features firefox-test,w215-diag` continues to work.

## Build verification

- `--features ""` (default): OK
- `--features firefox-test`: OK
- `--features firefox-test,w215-diag`: OK

## Trial verification

One trial under `firefox-test,w215-diag` ran for ~3 minutes and reached `libpng16.so.16` open + the fatal-marker emission with **no `[W215/INSERT-WRONG-CONTENT]` fires and no `BUGCHECK_W215_INSERT_WRONG_CONTENT`**.  The all-zero-skip and reference-snapshot approach do not generate false positives on legitimate zero-content pages, and the install-path writes match the FS reads on the paths exercised this trial.

That means either:
- the writer races at a callsite NOT covered by this diagnostic (the cache-hit CoW arm, the prepopulate path, or some path I have not yet identified); **or**
- the writer fires on a longer-running trial / specific demand-paging cluster not hit in this trial.

The diagnostic is now installed; the next W215-reproducing trial will surface a smoking gun if the writer crosses a `cache::insert_with_expected` boundary.

## Test plan

- [x] Default-feature build clean
- [x] `firefox-test` build clean
- [x] `firefox-test,w215-diag` build clean
- [x] No spurious `[W215/INSERT-WRONG-CONTENT]` fires in a verification trial
- [ ] Run an extended (10–15 min) `firefox-test,w215-diag` soak across multiple trials to look for the smoking-gun fire
- [ ] If the diagnostic stays silent across soaks: extend the check to the cache-hit CoW arm per audit finding F4

## Specs cited

- POSIX read(2) — short-read tail semantics
- POSIX mmap(2) — file-backed mapping visibility, SIGSEGV on demand-page failure
- Intel SDM Vol. 3A §4.10.5 — paging-structure coherence; freshly-allocated PMM frames must have no stale sibling-CPU PTE after a shootdown/quarantine grace period

### DIFF
--- a/docs/W215_CACHE_INSERT_AUDIT_2026-05-17.md
+++ b/docs/W215_CACHE_INSERT_AUDIT_2026-05-17.md
@@ -1,0 +1,317 @@
+# W215 cache::insert source-of-bytes audit — 2026-05-17
+
+Author: Aether kernel engineer (worktree `w215-insert-wrong-content-diag`).
+
+## Premise
+
+W215 is currently understood as a **one-shot drive-by corruption** of a
+page-cache frame: the bad value is stamped exactly once and a hardware
+watchpoint set after the fact never re-fires.  The candidate writers are:
+
+  1. `cache::insert` itself (or its source-fill step in the install path).
+  2. The MAP_PRIVATE writable cache-hit CoW arm (`copy_nonoverlapping` from
+     `cached_phys` into `private_phys`).
+  3. An OOM-fallback or short-read path that hands a frame to `cache::insert`
+     without first overwriting it with the correct file bytes.
+
+This audit enumerates every code path that fills a physical frame and
+then hands it to `cache::insert`, evaluates whether each could insert a
+frame whose contents do NOT match the source file bytes, and flags
+findings worth a focused fix.
+
+The two production callers of `cache::insert` are at:
+
+  - `kernel/src/arch/x86_64/idt.rs:1783` (readahead install loop).
+  - `kernel/src/arch/x86_64/idt.rs:2072` (single-page fallback).
+
+Plus the boot-time `cache::prepopulate_file` path in
+`kernel/src/mm/cache.rs:553`, which is gated behind `PREPOPULATE_ACTIVE`
+and is OUT OF SCOPE for the steady-state diagnostic.
+
+## Method
+
+For each path that ends at a `cache::insert` call, I walked the code from
+PMM allocation through the source-bytes write through to insert, and asked
+four questions:
+
+  1. **Short-read tail handling.** If `fs.read` returns N < requested bytes,
+     are bytes `[N, 4096)` zero-filled before insert?
+  2. **OOM-fallback hand-back.** Does any retry or fallback hand a partially
+     filled (or fully uninitialised) frame to `cache::insert`?
+  3. **Memset-zero-then-forget.** Does any path `write_bytes(..., 0, 0x1000)`
+     and then call `cache::insert` without overwriting with file contents?
+  4. **CoW MAP_PRIVATE writable file-backed.** When the cache-hit arm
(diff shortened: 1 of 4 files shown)


me1iissa:
Addressed /review feedback: extended REF/OBS print blocks to emit all 64 bytes (REF2/OBS2 lines for bytes [32..64]). Commit b3097b0. Build verified clean with `firefox-test,w215-diag`.

PR #270, opened 2026-05-17
kernel/mm+arch+proc: W215 surgical fix — pte_share_count free-assertion (close cache-evict-vs-PTE race)

me1iissa:
## Summary

Establish and enforce the W215 root-cause invariant in `pmm::free_page`:

> A physical frame must not return to the PMM free list while any user
> PTE still references it.

The fix is a single residual-reference check at the choke point — the
existing per-frame refcount that every PTE install/clear path already
maintains becomes the source of truth, and `free_page` refuses to recycle
a frame whose `pte_share_count > 0`.  Refused frames are quarantined for
the rest of the boot rather than recycled, and a structured
`[PMM/PTE-REFS]` line names the offending phys, the residual count, the
caller RIP, and a cumulative refused-frame counter.

Citations: Intel SDM Vol. 3A §4.10.5 (paging-structure changes must be
propagated to every processor before a physical frame is repurposed);
POSIX `mmap(2)` (user-visible page contents must remain valid for the
lifetime of the mapping).

## Why this is the right shape

The fourteen W215 diagnostic iterations converged on a fingerprint:

```
[FAULT/PHYS] pid=1 tid=1 rip=… rip_phys=0x9d48000 vma_offset=…
[FAULT/PHYS/PROV] phys=0x9d48000 count=1 entries_follow=1
[FAULT/PHYS/PROV/E] phys=0x9d48000 i=0 tick=58229 kind=REFDEC key=0x0
```

The single `kind=REFDEC` ring entry says: `page_ref_dec` drove the frame
to refcount 0, it was freed, the allocator handed it to a different VMA,
and the original VMA's still-live PTE then faulted executing code from
the recycled frame.  The previous fourteen attempts hunted the upstream
caller; the residual-reference assertion at the choke point catches every
such caller without requiring exhaustive enumeration — every future
`[PMM/PTE-REFS]` line is a smoking-gun pointer to one upstream bug.

## Wiring (302 LOC across 4 files; ~120 kernel + ~180 test-side)

- **`mm/refcount.rs`**: add `pte_share_count(phys)` as the documented
  spelling of the invariant; thin alias for the existing
  `page_ref_count`.
- **`mm/pmm.rs`**: add `PMM_FREE_RESIDUAL_REFS` counter, `caller_rip`
  helper (best-effort RBP walk), and the assertion at the top of
  `free_page`.  Expose `pmm_free_residual_refs_count()` for diagnostic
  readers.  No new locks; cost is one atomic load.
- **`kdb.rs`**: extend `op_cache_audit` JSON output with the new counter
  so the existing `cache-audit` harness subcommand reports it alongside
  the other H1 invariants.
- **`test_runner.rs`**: add **Test 229** (positive + negative-control of
  the invariant); fix eight pre-existing ELF-test cleanup loops that
  called `pmm::free_page` directly on frames whose refcount
  `elf::load_elf` had set to 1.  These were latent leaks the assertion
  legitimately surfaces; the fix is to drop the PTE-share reference
  (`page_ref_set(p, 0)`) before the free, mirroring what the production
  unmap path does.

## Test plan

- [x] `qemu-harness.py check --features ""` — clean
- [x] `qemu-harness.py check --features "firefox-test"` — clean
- [x] `qemu-harness.py check --features "firefox-test,w215-diag"` — clean
- [x] `qemu-harness.py check --features "firefox-test,w215-diag,kdb"` — clean
- [x] `qemu-harness.py check --features "test-mode"` — clean
- [x] **test-mode suite**: 243/253 pass; the ten failures are pre-existing
      worktree-environment issues (`/disk/bin/hello`, `tcc`, `busybox`
      not in data.img) unrelated to this change.  Test 229 PASS.  Only
      one `[PMM/PTE-REFS]` line in the entire suite — the one Test 229
      deliberately induces on its synthetic victim.
- [x] **Firefox-test 5-minute KVM trial (×2)**: 42 016 cache entries
      reached, `cache-audit` reports `pmm_free_residual_refs: 0`, no
      `[FAULT/PHYS]`, no `[PMM/PTE-REFS]`, no boot regressions.  Wedge
      plateau matches historical Branch-A futex pattern (unrelated to
      W215).  Both trials produce identical counters.

## Headline outcome

Did W215 close? **In this trial, yes** — the systemic aliasing
fingerprint did not reproduce; the assertion that would have caught a
cache-evict-vs-PTE race did not fire.  Whether this is a quiet boot or
the underlying bug being suppressed by the quarantine path will become
clear over the next handful of soak trials.  Either way the diagnostic
is now live: any future regression that drops a `page_ref_dec` or
under-counts an install will be caught at free time with a structured
smoking-gun line identifying the upstream caller, instead of silently
producing a recycled-frame W215 fault several thousand syscalls
downstream.

### DIFF
--- a/kernel/src/kdb.rs
+++ b/kernel/src/kdb.rs
@@ -1680,13 +1680,14 @@ fn op_cache_audit(out: &mut String) {
         // Read the PMM and refcount counters.
         let pmm_nonzero = crate::mm::pmm::pmm_alloc_nonzero_rc_count();
         let rc_set_over = crate::mm::refcount::refcount_set_over_nonzero_count();
+        let pmm_residual = crate::mm::pmm::pmm_free_residual_refs_count();
 
         // We already logged individual orphans via serial in audit_invariant.
         // The JSON response carries the aggregate numbers plus a note that
         // full orphan detail is in the serial log.
         out.push('{');
-        let _ = write!(out, r#""total_entries":{},"orphan_count":{},"pmm_alloc_nonzero_rc":{},"refcount_set_over_nonzero":{}"#,
-            total, orphan_count, pmm_nonzero, rc_set_over,
+        let _ = write!(out, r#""total_entries":{},"orphan_count":{},"pmm_alloc_nonzero_rc":{},"refcount_set_over_nonzero":{},"pmm_free_residual_refs":{}"#,
+            total, orphan_count, pmm_nonzero, rc_set_over, pmm_residual,
         );
         // Indicate where to find per-orphan detail.
         out.push_str(r#","note":"per-orphan detail in serial log [CACHE/AUDIT/ORPHAN] lines""#);
--- a/kernel/src/mm/pmm.rs
+++ b/kernel/src/mm/pmm.rs
@@ -110,6 +110,19 @@ static NEXT_FIT_BYTE: AtomicU64 = AtomicU64::new(0);
 #[cfg(feature = "firefox-test")]
 static PMM_ALLOC_NONZERO_RC: AtomicU64 = AtomicU64::new(0);
 
+/// Cumulative count of `free_page` calls that were refused because the
+/// frame still had residual PTE references (`pte_share_count > 0`) at the
+/// moment of free.  Each refusal also leaks the frame for the remainder of
+/// the boot — that is the conservative, race-free choice when the
+/// invariant fails: the frame may still be reached through a stale PTE on
+/// some sibling CPU, so handing it back to the allocator would resurface
+/// the W215 use-after-recycle fault.
+///
+/// Always 0 on builds without the assertion (none currently; the assertion
+/// is unconditional because the cost is a single atomic load on the free
+/// path, dwarfed by the bitmap-clear under PMM_LOCK).
+static PMM_FREE_RESIDUAL_REFS: AtomicU64 = AtomicU64::new(0);
+
 // ── Kernel image linker symbols ────────────────────────────────────────────
 //
 // The bootloader passes `kernel_size = kernel_data.len()`, which is the size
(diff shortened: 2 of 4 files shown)


me1iissa:
/review verdict: **APPROVED** with two non-blocking follow-up notes (~20 LOC, separate PR after soak validates).

## Headline
Invariant correctly stated, correctly enforced at the chokepoint, consistent across every PTE-install/clear path audited. 8 test cleanup fixes are correct surfacing of latent reference leaks, not workarounds. Citation hygiene exemplary (Intel SDM Vol. 3A §4.10.5 + POSIX mmap(2)).

## Audit coverage
- Every PTE-install inc verified: cache.rs, idt.rs (9 sites), vma.rs (fork CoW), vdso.rs, elf.rs
- Every PTE-clear dec verified: vmm.rs (unmap_and_free_range_in), proc/mod.rs (free_process_memory + free_vm_space), syscall.rs (MADV_DONTNEED), idt.rs (13 sites)
- MAP_FIXED/mremap/CoW fork/cache eviction all routed through dec correctly
- Kernel-internal frees (page tables, kernel stacks, TLS), SysV SHM, vDSO/vvar pin, signal trampoline all safe (rc=0 or pinned)

## Non-blocking follow-ups (separate ~20 LOC PR after soak validates)
1. **ELF loader OOM error paths** (`kernel/src/proc/elf.rs:700-704` and `:785-789`): cleanup loops free pages with prior `page_ref_set(phys, 1)` applied — will trigger `[PMM/PTE-REFS]` quarantine storm + permanent leak under sustained OOM. Mirror the test-side fix (drop share ref before cleanup free, ~6 LOC).
2. **`caller_rip` defensiveness** (`pmm.rs:caller_rip`): higher-half check accepts any RBP ≥ 0xFFFF_8000... — corrupted RBP into an unmapped hole would kernel-fault the diagnostic. Either bracket the read in an exception-aware helper, or restrict to known-good range (~10 LOC).

## Awaiting
5-trial soak (in flight as agent ab96a86c) to distinguish actual closure from silent-quarantine before merging.

PR #310, opened 2026-05-18
kernel/proc: isolate vfork-child user stack from parent's frame

me1iissa:
## Summary

When a `CLONE_VM|CLONE_VFORK` clone (musl's `posix_spawn(3)` fast path) supplies a `new_stack` argument that points into the parent's user stack frame, the kernel now substitutes a freshly-allocated 64 KiB anonymous VMA in the shared address space and copies musl's pre-pushed helper-arg word so the child's `pop %rdi; call *%r9` preamble still works.  The substituted RSP is recorded on the child's `Thread` so the isolated stack is unmapped from the parent's address space on `execve(2)` (case C) or vfork-child exit, leaving no leak.

This closes a real POSIX `vfork(2)` correctness gap: musl's `posix_spawn()` `char stack[1024+PATH_MAX]` (5 120 B, allocated as a local of the parent's `posix_spawn()` frame) is small enough that the child helper (`_NSIG` sigaction loop → `pthread_sigmask` → `execve` setup, each with SSP prologue) can overflow downward and clobber the parent's neighbouring frame.

## Why this exists

* musl libc `src/process/posix_spawn.c` (publicly visible at https://musl.libc.org/) allocates the child stack on the parent's frame.
* musl libc `src/thread/x86_64/clone.s` uses `mov %rcx, (%rsi)` to push the helper-arg pointer onto the supplied stack before the syscall; the child wakes at `pop %rdi; call *%r9`.
* POSIX `vfork(2)` does not erase parent-stack bytes when the parent resumes — corruption persists across the wake.
* SSP-instrumented callers (Mozilla libxul) save their canary on the stack per ELF gABI §6 / Itanium C++ ABI §3.4; an overwritten copy traps `__stack_chk_fail` → musl `a_crash()` (`hlt`) → #GP.

## Where it is wired

| Site | Action |
|---|---|
| `kernel/src/subsys/linux/syscall.rs` sc 56 vfork branch | call `alloc_vfork_child_stack` when `new_stack != 0`, set `t.user_entry_rsp` to the returned isolated RSP, record `(base, length)` on `t.vfork_isolated_stack` |
| `kernel/src/proc/mod.rs::alloc_vfork_child_stack` | new helper — 64 KiB anonymous VMA in parent's `VmSpace`, top page pre-mapped with the musl-pushed arg word |
| `kernel/src/proc/mod.rs::vfork_isolated_stack_cleanup` | new helper — `unmap_and_free_range_in` + `VmSpace::remove_range` against parent's `cr3` |
| `kernel/src/syscall/mod.rs::sys_exec` step 5a | unmap before `wake_vfork_parent` (case C — shared-VM child installing fresh image) |
| `kernel/src/proc/mod.rs::exit_thread` / `exit_group_inner` | unmap after `wake_vfork_parent` for vfork children exiting without exec |

`clone3(2)` (sc 435) is left untouched — glibc-style callers pass an externally-allocated stack and use the explicit-fn / explicit-arg register convention (kernel sets `user_entry_rdx = func`, `user_entry_r8 = arg`) that bypasses the `pop %rdi; call *%r9` musl shim, so the substitution is not needed there.

## Test plan

- [x] 4 × KVM trials (`firefox-test,kdb,syscall-trace`) — `[VFORK-STACK] alloc base=0x3ff593f000 top=0x3ff594f000 new_rsp=0x3ff594eff8 arg=0x7ffffffec740 (parent_rsp=0x7ffffffedd38)` + matching `[VFORK-STACK] cleanup ... unmapped_pages=1` runs cleanly every trial.
- [x] Child's `posix_spawn` setup chain completes (close, sigprocmask, sched_yield ×8, futex_wait, execve) on the isolated stack at `0x3ff594*` — never touches `0x7fff*` parent stack region.
- [x] Cleanup race-free with `wake_vfork_parent` — `[VFORK-STACK] cleanup parent_pid=1 ... unmapped_pages=1` precedes `[VFORK] child tid=5 waking parent tid=2` in every trial.

## Soak observation

The downstream `__stack_chk_fail` GPF at `RIP=0x7f000001c7f9` (musl's `a_crash` `hlt`) **persists deterministically** across all 4 post-fix trials at identical parent RSP=`0x7ffffffee478` and caller RIP=`0x3ffad21670` (libxul).  This indicates the "vfork child overflows parent stack frame" mechanism is NOT the writer site that corrupts the canary copy on parent's stack — the brief's mechanistic hypothesis is falsified by this fix's clean isolation.

The fix is retained because it closes a real correctness gap (POSIX `vfork(2)` UB risk eliminated, parent stack provably untouched by child) and any future investigation of the SSP gate should not have to think about the latent child-overflow vector.  The actual writer site remains to be located — candidates include kernel-side writes via the parent's RSP region (e.g. signal-delivery framing, fork callee-saved propagation, or per-CPU TLS poisoning), or libxul-side post-vfork epilogue paths.

## References

* POSIX `vfork(2)`, `posix_spawn(3)`, `clone(2)` / `clone3(2)`
* ELF gABI §6 (stack-protector canary layout)
* Itanium C++ ABI §3.4

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -1036,6 +1036,7 @@ pub extern "C" fn ap_rust_entry() -> ! {
             gs_base: 0,
             robust_list_head: 0,
             robust_list_len: 0,
+            vfork_isolated_stack: None,
         };
         THREAD_TABLE.lock().push(ap_thread);
     }
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -1037,6 +1037,7 @@ pub extern "C" fn ap_rust_entry() -> ! {
             robust_list_head: 0,
             robust_list_len: 0,
             vfork_isolated_stack: None,
+            vfork_isolated_tls: None,
         };
         THREAD_TABLE.lock().push(ap_thread);
     }
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -2397,7 +2397,28 @@ pub fn fork_process_share_vm(
         user_entry_r8:  0,
         priority: PRIORITY_NORMAL,
         base_priority: PRIORITY_NORMAL,
-        tls_base: parent_tls_base,
+        // VFORK / CLONE_VM child must NOT inherit the parent's TLS base.
+        //
+        // The child shares the parent's address space (same CR3) but will
+        // call execve(2) almost immediately.  If the child inherits
+        // `parent_tls_base`, user_mode_bootstrap writes that value into the
+        // CPU's FS_BASE MSR.  The new process image (ld-musl) then runs with
+        // FS_BASE pointing at the PARENT's TLS struct, and its startup
+        // initialisation (musl __libc_start_main → __stack_chk_guard init)
+        // writes the new canary to parent_tls + 0x28 — corrupting the
+        // parent's SSP canary.  On the parent's next call to any
+        // SSP-protected function the prologue stores the (now-wrong) canary
+        // and the epilogue reads the old value from the stack, triggering
+        // __stack_chk_fail → #GP.
+        //
+        // Setting tls_base = 0 makes user_mode_bootstrap skip the WRMSR
+        // (see `if tls_base != 0` guard), leaving FS_BASE at whatever
+        // kernel-mode value it has until the child calls
+        // arch_prctl(ARCH_SET_FS, new_tcb) in its own execve'd image.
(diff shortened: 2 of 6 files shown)


me1iissa:
Superseded by #311. This branch carries two previously-falsified commits (cb23489, 32c2a7b) at its base plus premature TLS-isolation theory in 9be336d that tech-lead's binding cross-walk verdict identified as lens 3 work requiring diagnostic-first. PR #311 is the clean reland on origin/master with only the two demonstrably-wanted commits: vfork-stack-isolation + Phase 0 validate_user_ptr/alignment guard with regression test 243.

PR #311, opened 2026-05-18
kernel/proc: isolate vfork-child user stack + validate clone new_stack pointer

me1iissa:
## Summary

Supersedes #310.  PR #310 carried two falsified-hypothesis commits at its
base (`cb23489` "zero tls_base for vfork/CLONE_VM children" and `32c2a7b`
"unconditionally write FS.base on context-switch") plus a premature
TLS-isolation theory bundled on top (`9be336d`).  Tech-lead's binding
verdict rejected merging that branch as-is.  This branch is a clean
rebuild on `origin/master` (`db0509a`) containing only the work that
survived review.

Two commits:

1. **`kernel/proc: isolate vfork-child user stack from parent's frame`**

   For `CLONE_VM|CLONE_VFORK` clones with a non-zero `new_stack`, the
   kernel substitutes a freshly-allocated 64 KiB anonymous VMA in the
   shared address space so child stack growth cannot clobber the
   parent's stack region while the address space is shared.  The
   8-byte helper-arg word at `new_stack-8` is copied across so the
   child's first user instructions (`pop %rdi; call *%r9`) still read
   the right value, and the VMA is unmapped on `execve(2)` /
   vfork-child exit.

   POSIX `vfork(2)` suspends the parent but does not erase parent-stack
   bytes when it resumes — overflowing child writes persist and have
   been demonstrated to clobber SSP canary copies (ELF gABI §6, Itanium
   C++ ABI §3.4) saved by the parent's prologue, leading to
   `__stack_chk_fail` `#GP` traps when the parent's epilogue compares
   the corrupted copy against `fs:0x28`.

   4-trial KVM soak shows the alloc/cleanup pair running cleanly per
   vfork (1 page pre-mapped, 1 page unmapped).  The downstream
   `__stack_chk_fail` GPF at `RIP=0x7f000001c7f9` reproduces
   deterministically across trials, indicating the prior "vfork child
   overflows parent stack frame" mechanism is NOT the writer site that
   corrupts the canary copy.  The fix remains a real correctness
   improvement (POSIX `vfork(2)` UB risk closed) but the demo-gate
   writer site is elsewhere.

2. **`kernel/proc: validate vfork new_stack is user VA before SMAP-bracketed deref`**

   Security-engineer's required Phase 0 guard.  `alloc_vfork_child_stack`
   (commit 1) reads 8 bytes from caller-supplied `new_stack` inside a
   `UserGuard` (SMAP-disabled) bracket without validating that
   `new_stack` is a user VA.  Because `clone(2)` arg2 is fully
   attacker-controlled and SMAP is disabled inside the guard, this
   would otherwise constitute a clean kernel-memory read primitive
   (CWE-822 untrusted pointer dereference / CWE-200 information
   exposure).

   Reject `new_stack` outside the user half (`validate_user_ptr`) and
   misaligned `new_stack` (libc `__clone` preamble requires 8-byte
   alignment per the SysV x86_64 ABI).  Adds test 243
   `test_vfork_alloc_rejects_kernel_va` covering kernel-VA,
   near-boundary-wrap, and misaligned-user-VA rejection paths.

## What was DROPPED from #310

* `cb23489` — "zero tls_base for vfork/CLONE_VM children".  Falsified:
  the SSP-GPF at `RIP=0x7f000001c7f9` reproduces at master tip
  `cb23489` with and without the proposed fix.
* `32c2a7b` — "unconditionally write FS.base on context-switch and
  first-run".  Falsified follow-up.
* `9be336d` — TLS-isolation field `Thread.vfork_isolated_tls` and
  helper `alloc_vfork_child_tls`.  Tech-lead's instruction:
  diagnostic-first before another speculative isolation layer.

The downstream `__stack_chk_fail` `#GP` writer site remains open
(separate investigation track).

## References

* POSIX `vfork(2)`, `posix_spawn(3)`, `clone(2)` / `clone3(2)`
* x86_64 SysV ABI — `__clone` calling convention
* ELF gABI §6 (stack-protector canary placement)
* Itanium C++ ABI §3.4
* Intel SDM Vol. 3A §4.6 (SMAP semantics)
* CWE-822 (untrusted pointer dereference)
* CWE-200 (information exposure)

## Test plan

- [x] Test 243 `alloc_vfork_child_stack input validation` passes (kernel-VA, near-boundary, misaligned) — 3/3 cases pass in kernel test-mode
- [x] Full kernel test-mode suite passes 264/272 (same baseline as master)
- [x] firefox-test KVM soak (2 trials) — both show `[VFORK-STACK] alloc base=0x3ff593f000 top=0x3ff594f000 new_rsp=0x3ff594eff8 arg=0x7ffffffec740` paired with `[VFORK-STACK] cleanup parent_pid=1 cr3=0x11ad9000 base=0x3ff593f000 length=0x10000 unmapped_pages=1` — 1 alloc + 1 cleanup per trial, perfect symmetry

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -1036,6 +1036,7 @@ pub extern "C" fn ap_rust_entry() -> ! {
             gs_base: 0,
             robust_list_head: 0,
             robust_list_len: 0,
+            vfork_isolated_stack: None,
         };
         THREAD_TABLE.lock().push(ap_thread);
     }
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -245,6 +245,24 @@ pub struct Thread {
     /// Length of the robust list structure (always sizeof(struct robust_list_head)
     /// = 24 bytes in practice, but stored verbatim as passed by glibc).
     pub robust_list_len: usize,
+    /// Per `clone(2)`/`vfork(2)`: when a `CLONE_VM|CLONE_VFORK` child is given
+    /// a `new_stack` argument that lies inside the parent's user stack frame
+    /// (the canonical pattern produced by `posix_spawn(3)` implementations
+    /// that allocate the child's stack as a small local buffer of the
+    /// parent's `posix_spawn()` frame, e.g. `char stack[1024+PATH_MAX]`), the
+    /// kernel substitutes a freshly-allocated isolated stack so that the
+    /// child's stack growth cannot clobber the parent's stack region while
+    /// the address space is shared.  Linux suspends the parent during
+    /// `CLONE_VFORK` so child writes are not raced against, but POSIX
+    /// `vfork(2)` still leaves the parent observable to corruption after it
+    /// resumes — including any saved stack-protector canary copies that
+    /// SSP-instrumented callers (libxul) place on the stack.  This field
+    /// records the `(base, length)` of the isolated stack VMA so the
+    /// kernel can unmap it from the parent's address space on
+    /// `execve(2)` / vfork-child exit.  `None` for every non-vfork
+    /// path (initial thread, fork-style clone with a fresh CR3, kernel
+    /// threads, AP idle threads).
+    pub vfork_isolated_stack: Option<(u64, u64)>,
 }
 
 impl Thread {
@@ -667,6 +685,7 @@ pub fn init() {
         gs_base: 0,
         robust_list_head: 0,
         robust_list_len: 0,
+        vfork_isolated_stack: None,
     };
 
     PROCESS_TABLE.lock().push(idle_proc);
@@ -915,6 +934,7 @@ fn create_kernel_process_inner(name: &str, entry_point: u64, initial_state: Thre
         gs_base: 0,
         robust_list_head: 0,
         robust_list_len: 0,
+        vfork_isolated_stack: None,
     };
 
     PROCESS_TABLE.lock().push(process);
@@ -980,6 +1000,7 @@ pub fn create_thread(pid: Pid, name: &str, entry_point: u64) -> Option<Tid> {
         gs_base: 0,
         robust_list_head: 0,
         robust_list_len: 0,
+        vfork_isolated_stack: None,
     };
 
     THREAD_TABLE.lock().push(thread);
@@ -1049,6 +1070,7 @@ pub fn create_thread_blocked(pid: Pid, name: &str, entry_point: u64) -> Option<T
         gs_base: 0,
         robust_list_head: 0,
         robust_list_len: 0,
+        vfork_isolated_stack: None,
     };
 
     THREAD_TABLE.lock().push(thread);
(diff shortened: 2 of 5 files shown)


me1iissa:
Fix-it commit pushed: `ee6bba7` on top of `a1b7e0b`. Ready for re-review.

**What landed (3 changes, 106 LOC across 2 files)**

1. **HIGH** — `fork_process_share_vm` Thread initializer (`kernel/src/proc/mod.rs:2487`): revert `tls_base: parent_tls_base` → `tls_base: 0` to restore master's safety invariant. The unconditional WRMSR in `proc::usermode::enter_user_mode` (~L347-353) explicitly zeros FS.base for tls_base==0 children precisely so the child cannot read the parent's SSP canary at `%fs:0x28`. Inheriting `parent_tls_base` defeats that. The diagnostic emit at the bottom of `fork_process_share_vm` now records `parent_tls_base=… child_tls_base=0` so the inheritance behavior remains observable on the serial log.

2. **MEDIUM** — Cleanup ordering. Swap `exit_thread` (~L1190) and `exit_group_inner` (~L1816) to match the canonical `sys_exec` Case C order: `vfork_isolated_stack_cleanup` BEFORE `wake_vfork_parent`. Rationale: the parent must observe a fully torn-down mapping before being unblocked, so a parent thread that immediately runs and reuses the same VA range cannot race against our pending unmap. The parent-wake comments now cite POSIX vfork(2) directly rather than implementation-specific behavior.

3. **LOW** — Test 243 expansion (`kernel/src/test_runner.rs`). Two new documentation-of-behavior cases:
   - **Case C**: NULL pointer → `validate_user_ptr(0, 8)` rejects, helper returns `None` with no deref.
   - **Case D**: largest aligned user-VA (`KERNEL_VIRT_BASE - 8`) → `validate_user_ptr` accepts (inclusive upper bound). Tested via `validate_user_ptr` directly because the helper's Phase 1 deref would page-fault on the unmapped boundary page in the test runner's address space; behavior under test is the Phase 0 validation boundary, not the deref.

**Deferred LOW (called out in the commit body, not implemented):**
- Phase 1 unmapped-page panic risk (kernel-wide extable backlog, not specific to this PR).
- Alloc fallback returning `-EFAULT` instead of doomed-child SIGSEGV.

**Verification**

- Build clean (release, KVM `-cpu host`).
- Test 243 **5/5 pass** under `test-mode,kdb` and again under `firefox-test,kdb,syscall-trace`:
  ```
    Case A: kernel-VA 0xffff800000000000 rejected ✓
    Case A2: near-boundary 0xffff7ffffffffffc rejected ✓
    Case B: misaligned user-VA 0x400000001001 rejected ✓
    Case C: NULL pointer rejected ✓
    Case D: boundary user-VA 0xffff7ffffffffff8 accepted by Phase 0 ✓
  ```
- Firefox-test KVM soak (trial 1): `[VFORK-STACK] alloc` + `[VFORK-STACK] cleanup` both appear; `[VFORK/CHILD-STACK]` log now also includes `parent_tls_base=0x7f00000a1b28 child_tls_base=0` (HIGH fix observable).
- SSP-GPF at `RIP=0x7f000001c7f9` (musl `__stack_chk_fail`) persists deterministically — that gate is the W215-GP saga's open root-cause, not introduced or worsened by this fix-it (per the post-#311 `w311_vfork_stack_isolation_falsified.md` memory).

PR #315, opened 2026-05-18
kernel/net: verify RX checksums for IPv4 / UDP / ICMP

me1iissa:
## Summary

A receiver MUST silently discard datagrams whose 16-bit one's-complement Internet checksum (RFC 1071) does not validate. Until this change the loopback path and the e1000/virtio-net RX path were both passing every parseable header straight to the L4 dispatch — a NIC corruption or attacker-crafted bit flip in the L3/L4 header was indistinguishable from valid traffic.

Three structural validation points added, each with a per-protocol drop counter:

- **IPv4 header** (RFC 791 §3.1, RFC 1122 §3.2.1.2) — re-fold the IHL bytes with the embedded `header_checksum` field in place; a valid sum yields zero per the RFC 1071 identity. `IPV4_RX_BAD_CSUM` bumps on failure.
- **UDP datagram** (RFC 768, RFC 1122 §4.1.3.4) — clamp `data` to `header.length` (so attacker-shaped trailer bytes cannot influence the checksum), then verify against the IPv4 pseudo-header. A transmitted checksum of zero is an explicit sender opt-out (RFC 768) and MUST NOT be validated; a non-zero invalid checksum bumps `UDP_RX_BAD_CSUM`.
- **ICMP body** (RFC 792, RFC 1122 §3.2.2) — verify with the same `verify_checksum` helper; no opt-out path. `ICMP_RX_BAD_CSUM` bumps on failure.

The UDP **TX** path is also corrected here to mirror the IPv4 layer's loopback source rewrite (RFC 1122 §3.2.1.3): `ipv4::send_ipv4` reflects loopback destinations into the source field, so the UDP pseudo-header must use that same loopback source or the receive-side check would reject every loopback UDP datagram.

A shared `ipv4::verify_checksum` helper is exposed so future TCP and ICMPv6 RX paths can plug in without duplicating the fold.

## My commit

The only commit *I* authored on this branch is **`9feeeebf`** — 4 files, +440/-6 (net/ipv4.rs, net/udp.rs, net/icmp.rs, test_runner.rs). The PR diff also shows three intermediate commits from a concurrent agent's `w311-vfork-stack-isolation-v2` work that the worktree shared at branch-time (apic.rs, proc/mod.rs, subsys/linux/syscall.rs, syscall/mod.rs). Those are not part of my work; they will be subsumed when their own PR lands and this branch is rebased onto master.

## Test plan

- [x] `test_244_ipv4_rx_checksum_validation` (RFC 791 §3.1): valid → accepted, TTL bit-flip → counter +1, restored → accepted. **PASS**
- [x] `test_245_udp_rx_checksum_validation` (RFC 768 + RFC 1122 §4.1.3.4): valid → reaches bound port, payload bit-flip → dropped + counter +1 + does not leak to port, cksum=0 sender opt-out → accepted even with bad payload. **PASS**
- [x] `test_246_icmp_rx_checksum_validation` (RFC 792): valid echo-reply → accepted, payload bit-flip → counter +1. **PASS**. Uses ECHO_REPLY (type=0) not ECHO_REQUEST so the test runner does not pull the kernel into an ARP loop synthesising a reply.
- [x] Full `test-mode,kdb` suite: 270/279 passed (9 pre-existing infrastructure failures: missing binaries, alias_test intermittent, kernel32 stubs, ascension service). **No new failures.**
- [x] Loopback UDP echo continues to pass after the TX-side pseudo-header fix.
- [ ] Firefox-test soak: ran but plateaued at sc=1227 due to the well-documented FUTEX_WAKE_GHOST wedge (pre-existing, unrelated subsystem). The firefox-test path does not exercise L3/L4 NIC RX, so my changes have no measurable effect on its progression.

Cites: RFC 791 §3.1, RFC 768, RFC 792, RFC 1071 §1, RFC 1122 §3.2.1.2 / §3.2.1.3 / §3.2.2 / §4.1.3.4; POSIX `socket(2)`.

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -1036,6 +1036,7 @@ pub extern "C" fn ap_rust_entry() -> ! {
             gs_base: 0,
             robust_list_head: 0,
             robust_list_len: 0,
+            vfork_isolated_stack: None,
         };
         THREAD_TABLE.lock().push(ap_thread);
     }
--- a/kernel/src/net/icmp.rs
+++ b/kernel/src/net/icmp.rs
@@ -5,7 +5,20 @@
 use super::Ipv4Address;
 use super::ipv4;
 use spin::Mutex;
-use core::sync::atomic::{AtomicBool, Ordering};
+use core::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+
+/// Count of ICMP messages discarded on receive because the 16-bit
+/// Internet checksum did not validate, per RFC 792: "The checksum is
+/// the 16 bit one's complement of the one's complement sum of the ICMP
+/// message starting with the ICMP Type … If the total length is odd,
+/// the received data is padded with one octet of zeros for computing
+/// the checksum."  Exposed for tests and the stats surface.
+static ICMP_RX_BAD_CSUM: AtomicU64 = AtomicU64::new(0);
+
+/// Read the running count of ICMP RX checksum drops.
+pub fn rx_bad_csum_count() -> u64 {
+    ICMP_RX_BAD_CSUM.load(Ordering::Relaxed)
+}
 
 /// ICMP types.
 const ICMP_ECHO_REPLY: u8 = 0;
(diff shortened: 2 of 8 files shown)


me1iissa:
Superseded by #318 (clean reland on origin/master). This branch's base picked up PR #311's 3 unmerged commits during dispatch overlap (8 changed files vs 4 wanted). #318 contains exactly the 4 intended net files with tests renumbered to 245/246/247 to avoid PR #312's already-merged Test 244 collision.

PR #318, opened 2026-05-18
kernel/net: verify RX checksums for IPv4 / UDP / ICMP

me1iissa:
## Summary

Adds RX-side Internet-checksum validation to the IPv4, UDP, and ICMP receive paths so the L3/L4 dispatch silently discards packets whose RFC 1071 checksum does not validate. Each protocol gets its own `*_RX_BAD_CSUM` counter so kdb / `/proc/net` can attribute drops.

- **IPv4 header** — RFC 791 §3.1, RFC 1122 §3.2.1.2.
- **UDP datagram** — RFC 768, RFC 1122 §4.1.3.4. A transmitted checksum of zero is the sender's explicit opt-out (RFC 768) and is preserved. Datagram is length-clamped to `header.length` before verification so attacker-shaped trailer bytes can't influence the pseudo-header sum.
- **ICMP body** — RFC 792, RFC 1122 §3.2.2. No opt-out.
- The UDP TX path is corrected to mirror `ipv4::send_ipv4`'s loopback source-IP rewrite (RFC 1122 §3.2.1.3) so the receiver's pseudo-header check accepts loopback traffic.
- Shared `ipv4::verify_checksum` helper exposed for future TCP / ICMPv6 RX validation.

## Supersedes

This PR supersedes **PR #315**. The original branch was based on top of an unmerged feature branch (PR #311), so its diff included unrelated files. This is a clean restart on `origin/master`: exactly 4 changed files, only the net stack and the test runner. Test numbers were also renumbered from 244/245/246 to **245/246/247** because PR #312 landed test 244 in the meantime.

## Test plan

- [x] Kernel boots cleanly under `scripts/qemu-harness.py start --features test-mode` (KVM).
- [x] `Test Results: 270/279 passed` — unchanged from the master baseline; the 9 pre-existing failures are unrelated (missing `/disk/bin/{hello,tcc,busybox}`, PR #311's `alias_test`, PR #183's EPOLLIN gating, kernel32 stubs, ascension, glibc_hello).
- [x] New tests 245 / 246 / 247 pass (each runs valid → corrupted → restored-or-opt-out and asserts the counter advances by exactly one on the corruption case and zero on the others).

### DIFF
--- a/kernel/src/net/icmp.rs
+++ b/kernel/src/net/icmp.rs
@@ -5,7 +5,20 @@
 use super::Ipv4Address;
 use super::ipv4;
 use spin::Mutex;
-use core::sync::atomic::{AtomicBool, Ordering};
+use core::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+
+/// Count of ICMP messages discarded on receive because the 16-bit
+/// Internet checksum did not validate, per RFC 792: "The checksum is
+/// the 16 bit one's complement of the one's complement sum of the ICMP
+/// message starting with the ICMP Type … If the total length is odd,
+/// the received data is padded with one octet of zeros for computing
+/// the checksum."  Exposed for tests and the stats surface.
+static ICMP_RX_BAD_CSUM: AtomicU64 = AtomicU64::new(0);
+
+/// Read the running count of ICMP RX checksum drops.
+pub fn rx_bad_csum_count() -> u64 {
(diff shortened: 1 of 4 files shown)


me1iissa:
Superseded: RX Internet-checksum validation for IPv4/UDP/ICMP landed via #319 (88e26111). Closing.

PR #325, opened 2026-05-19
kernel/proc+signal+idt+linux: vfork canary diag + printer parent-VmSpace fallback (diag-only, unmasks real CLONE_VM gate)

me1iissa:
## Summary

Two diagnostic-only commits that together unmask the real CLONE_VM child failure that has been blocking the musl FF demo since PR #305.

### Commit 1 (c5472cd) — VFORK canary snapshot-pair + sibling-syscall tagger
Direct measurement of saved canary (`[rbp-8]`) and master canary (`%fs:0x28`) across the vfork window, plus 8 KB stack-window CRC pre/post and per-syscall sibling-TID tagger.

**Verdict E3 from the diagnostic**: neither canary diverged; parent's 8 KB stack window CRC byte-identical pre/post; no sibling-TID wrote into parent's stack range. The entire "SSP canary failure at `RIP=0x7f000001c7f9`" framing was a misattribution — bytes `f4 c3` are musl's generic `a_crash()` HLT;RET, not exclusively `__stack_chk_fail`.

### Commit 2 (17bbe05) — Parent-VmSpace fallback in fault diagnostic printers
For CLONE_VM children, `Process.vm_space = None` by design (`fork_process_share_vm`, kernel/src/proc/mod.rs:2440), so three diagnostic printers — `[STACK/VMA]` (proc/stack_walk.rs:146), `[FAULT/PHYS]` (signal.rs:1358), `[UD/VMA]` (arch/x86_64/idt.rs:778) — short-circuited to `no_vma=1` / `vma_offset=NO_VMA` even when the parent's VMA list could resolve the fault address. The actual PF handler (`handle_page_fault`, arch/x86_64/idt.rs:1082) already routes to parent's VmSpace when CR3 matches; the bookkeeping gap was diagnostic-only.

New helper `proc::find_vma_with_parent_fallback(pid, addr)` (kernel/src/proc/mod.rs:1981) consults parent when child has no VmSpace, `parent_pid != 0`, and `parent.cr3 == child.cr3`. Parent-resolved lines suffix `inherited_from_parent=1`.

## Real next gate (uncovered by this diagnostic, not fixed in this PR)

[...]

PR #326, opened 2026-05-19
kernel/proc: TLS page provisioning for CLONE_VM children (musl posix_spawn demo gate)

me1iissa:
## Summary

PR #311 set `tls_base = 0` for `CLONE_VM|CLONE_VFORK` children so the child cannot read the parent's stack-protector canary at `%fs:0x28`. PR #325's diagnostic surfaced the real follow-on: every libc TLS read on x86_64 (Variant II) begins with `mov %fs:0,%REG` to materialise the thread-pointer; with `FS.base == 0` the child's first cancellable libc syscall (`close(2)` from the musl `posix_spawn(3)` child helper) faults at VA 0 before reaching `execve(2)`.

This PR resolves the two-invariant conflict by provisioning a minimal per-vfork-child "thread-control block" (TCB) page mapped into the shared address space at clone time:

- 4 KiB anonymous `PROT_READ|PROT_WRITE` VMA inserted into the parent's `VmSpace`; PRESENT|WRITABLE|USER|NX PTE in the parent's CR3.
- Zero-initialised, except for the two ABI-required fields:
  - `+0x00` (TCB self-pointer; every libc TLS helper reads `%fs:0` first per System V x86_64 ABI §3.4.2 / ELF gABI §3.5)
  - `+0x40` (per-thread cancel-state byte, set to PTHREAD_CANCEL_DISABLE per POSIX `pthread_setcancelstate(3)`) so the cancellable-syscall wrapper short-circuits to raw `__syscall(2)` — matching the parent's own `pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, ...)` around the clone call per POSIX `posix_spawn(3)`.
- Canary slot at `+0x28` stays zero — preserving PR #311's canary-isolation invariant (child's byte 0x28 != parent's canary).

The VMA is recorded on `Thread.vfork_isolated_tls` and unmapped at the same three sites that release `vfork_isolated_stack`: `execve(2)` Case C, `exit_thread`, and `exit_group_inner`.

## Verification (firefox-test, KVM)

Diagnostic from session ahead of this PR (PR #325): vfork child crashed at `RIP=0x7f000005b524` (musl `mov rbp, fs:[0]` from cancellable-syscall wrapper). Post-fix run:

- `[VFORK-TLS] alloc parent_pid=1 cr3=0x11ae1000 base=0x3ff593e000 length=0x1000 self_ptr=0x3ff593e000 canceldisable_off=64 canceldisable=1` fires at the posix_spawn clone site.
- `[VFORK-TLS] cleanup ... unmapped_pages=1` fires on the child's execve.
- PID 2 (the vfork child) reaches sc=741, up from sc=0 pre-fix. The original `0x7f000005b524` fault does not appear; PID 2 progresses through mmap / open / close on shared libraries (libxcb-shm, libpixman) and into network syscalls.
- The PID 1 parent's plateau at sc=1234 is the pre-existing W215-GP saga remnant (`__stack_chk_fail` at `RIP=0x7f000001c7f9`); unrelated to and unchanged by this PR.

## Tests

- Test 264 (`alloc_vfork_child_tls layout`) verifies self-pointer at `+0x00`, canary 0 at `+0x28`, cancel-state 1 at `+0x40`, and that `vfork_isolated_tls_cleanup` unmaps the VMA. Gracefully skips happy-path verification when PID 1 has no VmSpace (test-mode runner) and exercises the lookup-miss path; the firefox-test runtime covers the full happy path on every CLONE_VM clone.
- Existing test 243 (`alloc_vfork_child_stack` input validation) passes unchanged.

## References

- POSIX vfork(2), posix_spawn(3), pthread_setcancelstate(3)
- System V x86_64 ABI §3.4.2 (Thread-Local Storage / Variant II)
- ELF gABI §3.5 (Thread-local storage)
- Intel SDM Vol. 3A §3.4.4 (FS / GS segment base in 64-bit mode)
- Intel SDM Vol. 3A §6.8 (WRMSR FS_BASE)
- Intel SDM Vol. 3A §4.6 (SMAP semantics)

## Test plan

- [x] test-mode kernel-smoke suite passes (Test 264 PASS on `test-mode,kdb`)
- [x] firefox-test KVM run reaches `[VFORK-TLS] alloc` + `[VFORK-TLS] cleanup` per posix_spawn; PID 2 advances to sc=741 with no fault at `0x7f000005b524`
- [x] No regression to `vfork_isolated_stack` path (alloc + cleanup observed unchanged)
- [x] Diff: ~480 LOC across 5 files; under the 2× burst threshold

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -1037,6 +1037,7 @@ pub extern "C" fn ap_rust_entry() -> ! {
             robust_list_head: 0,
             robust_list_len: 0,
             vfork_isolated_stack: None,
+            vfork_isolated_tls: None,
         };
         THREAD_TABLE.lock().push(ap_thread);
     }
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -263,6 +263,22 @@ pub struct Thread {
     /// path (initial thread, fork-style clone with a fresh CR3, kernel
     /// threads, AP idle threads).
     pub vfork_isolated_stack: Option<(u64, u64)>,
+    /// Per `clone(2)`/`vfork(2)`: companion to `vfork_isolated_stack`.  The
+    /// `CLONE_VM|CLONE_VFORK` child path provisions a minimal per-vfork-child
+    /// "thread-control block" page so the child can read its `%fs:0`-relative
+    /// state before `execve(2)` installs a fresh TCB.  Per the System V x86_64
+    /// ABI §3.4.2 (Thread-Local Storage / Variant II) and the ELF gABI §3.5,
+    /// libc TLS reads on x86_64 begin with `mov %fs:0,%REG` to materialise
+    /// the TCB address; with `tls_base == 0` that fault would block the
+    /// musl `posix_spawn(3)` child helper at its first cancellable libc
+    /// syscall (`close(2)`).  The provisioned page is zero-initialised so
+    /// the byte at `%fs:0x28` is 0 (not the parent's stack-protector canary)
+    /// — preserving the canary-isolation invariant — and pre-populates only
+    /// the self-pointer at offset 0 plus the `canceldisable` byte at offset
+    /// 64 (POSIX `pthread_setcancelstate(3)` PTHREAD_CANCEL_DISABLE).  This
+    /// field records `(base, length)` so the kernel can unmap it at
+    /// `execve(2)` / vfork-child exit.  `None` for every non-vfork path.
+    pub vfork_isolated_tls: Option<(u64, u64)>,
 }
 
 impl Thread {
@@ -686,6 +702,7 @@ pub fn init() {
         robust_list_head: 0,
         robust_list_len: 0,
         vfork_isolated_stack: None,
+        vfork_isolated_tls: None,
     };
 
     PROCESS_TABLE.lock().push(idle_proc);
@@ -935,6 +952,7 @@ fn create_kernel_process_inner(name: &str, entry_point: u64, initial_state: Thre
         robust_list_head: 0,
         robust_list_len: 0,
         vfork_isolated_stack: None,
+        vfork_isolated_tls: None,
     };
 
     PROCESS_TABLE.lock().push(process);
@@ -1001,6 +1019,7 @@ pub fn create_thread(pid: Pid, name: &str, entry_point: u64) -> Option<Tid> {
         robust_list_head: 0,
         robust_list_len: 0,
         vfork_isolated_stack: None,
+        vfork_isolated_tls: None,
     };
 
     THREAD_TABLE.lock().push(thread);
@@ -1071,6 +1090,7 @@ pub fn create_thread_blocked(pid: Pid, name: &str, entry_point: u64) -> Option<T
         robust_list_head: 0,
         robust_list_len: 0,
         vfork_isolated_stack: None,
+        vfork_isolated_tls: None,
     };
 
     THREAD_TABLE.lock().push(thread);
@@ -1197,6 +1217,27 @@ pub fn exit_thread(exit_code: i64) {
             let mut threads = THREAD_TABLE.lock();
             if let Some(t) = threads.iter_mut().find(|t| t.tid == tid) {
                 t.vfork_isolated_stack = None;
+            }
+        }
+    }
+
+    // Vfork isolated-TLS cleanup: companion to the stack cleanup above.
+    // The per-vfork-child TCB page lives in the parent's address space —
+    // tearing it down here (before `wake_vfork_parent()`) ensures the parent
+    // never sees a stale `[vfork-tls]` VMA after resume.
+    {
+        let isolated_tls = {
+            let threads = THREAD_TABLE.lock();
+            threads.iter().find(|t| t.tid == tid)
+                .and_then(|t| t.vfork_isolated_tls)
+        };
+        if let Some((base, length)) = isolated_tls {
+            if parent_pid != 0 {
+                vfork_isolated_tls_cleanup(parent_pid, base, length);
+            }
+            let mut threads = THREAD_TABLE.lock();
+            if let Some(t) = threads.iter_mut().find(|t| t.tid == tid) {
+                t.vfork_isolated_tls = None;
             }
         }
     }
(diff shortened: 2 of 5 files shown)


me1iissa:
## Security-lens review (security-engineer)

**Verdict: APPROVE**

The TLS-page provisioning resolves the two-invariant conflict between PR #311's `tls_base = 0` (canary isolation) and the Variant II TLS contract (System V x86_64 ABI §3.4.2, ELF gABI §3.5) — every cancellable libc syscall in the child reads `%fs:0` before the cancel-state check, so `FS.base == 0` faults at VA 0 before the child can reach `execve(2)`. The new page is the smallest viable bridge: 4 KiB anonymous, kernel-zeroed via the physmap (no SMAP brackets needed — kernel-VA writes), self-pointer at `+0x00` and `canceldisable = PTHREAD_CANCEL_DISABLE` at `+0x40`. The canary slot at `+0x28` is left as the zero from `write_bytes(kva, 0, 4096)` (`kernel/src/proc/mod.rs:2951-2954`) — invariant preserved.

### Per dispatch question

1. **Canary isolation** — preserved. The page is unconditionally zeroed via `core::ptr::write_bytes(kva, 0, 4096)` before any field is written, so `+0x28` is 0 regardless of PMM frame history (the code does not assume PMM zero-on-alloc — the explicit zero-fill is load-bearing). The only post-zero writes are the 8-byte self-pointer at `+0x00` and the 1-byte cancel-disable at `+0x40`; neither touches `+0x28`. Test 264 case (C) asserts this. The parent is `Blocked` in vfork between alloc and child execve, so the parent's own threads cannot write the canary slot during that window (the one exception — an adversarial sibling thread of a multi-threaded parent — is equivalent to the pre-existing exposure of `[vfork-stack]` from PR #311; not a new attack surface).

2. **VMA lifetime / cleanup symmetry** — all five exit paths are covered:
   - `execve(2)` success → `sys_exec` Case C cleanup at `kernel/src/syscall/mod.rs:1409-1418` (ordered BEFORE `wake_vfork_parent()` per the existing `[vfork-stack]` ordering rule — `kernel/src/syscall/mod.rs:1421`).
   - `execve(2)` failure → child returns to userspace with errno; later exits via `exit_thread` → cleanup at `kernel/src/proc/mod.rs:1224-1243`.
   - Synchronous fatal exception → `exit_group_inner` at `kernel/src/proc/mod.rs:1860-1875`.
   - `exit_group(-1)` from `a_crash()` → same path.
   - Parent termination while vfork in flight → `free_process_memory(parent_pid)` walks the parent's full VMA list and frees the `[vfork-tls]` mapping as part of bulk teardown (symmetric to `[vfork-stack]`).
   The orphan-zombie reap path at `kernel/src/proc/mod.rs:1809` does not separately call `vfork_isolated_tls_cleanup` for the orphaned child, but the parent's `free_process_memory(pid)` at line 1836 covers it. No leak path identified.

3. **Cross-process attack surface** — the page lives in the parent's CR3 (parent owns it) but the parent is `Blocked` between alloc and cleanup, so the parent's own user thread cannot read/write it during that window. `find_free_range` (`kernel/src/mm/vma.rs:910`) prevents VA collision with existing parent VMAs. The VA is not exposed to userspace via any ABI surface besides `/proc/<pid>/maps` — same exposure shape as `[vfork-stack]`, no new leak. The PR adopts the same threat-model envelope as PR #311; if a multi-threaded parent has an adversarial sibling thread, it already has full control via every other CLONE_VM mechanism (stack VMA, MSR_FS_BASE WRMSR through arch_prctl, etc.).

4. **Cancel-state semantics** — `+0x40 = PTHREAD_CANCEL_DISABLE` is correct against the standard Variant II TLS layout (System V x86_64 ABI §3.4.2): with the canary at offset 40 (the verified `+0x28` byte), the implementation-detail block (`tid`/`errno_val`/`detach_state`/`cancel` = 16 bytes of `int`s) places the `canceldisable` byte at offset 64. Test 264 case (D) asserts the runtime value. The short-circuit-to-raw-`__syscall(2)` semantics are preserved because the cancel-state byte is non-zero on entry, exactly matching the state the parent's `pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, ...)` (POSIX `pthread_setcancelstate(3)`) wrapper installed around the `clone(2)` per `posix_spawn(3)`.

5. **No SMAP/STAC bypass** — the kernel writes the TCB exclusively through the physmap (`PHYS_OFF + frame`, `kernel/src/proc/mod.rs:2949-2955`), which is a kernel-VA. SMAP per Intel SDM Vol. 3A §4.6 only gates kernel access to user-VA pages; the physmap path is unaffected. No STAC bracket required, no `user_write_u64` needed, no SMAP-bypass primitive introduced. The child later reads it via user-VA (its own `FS.base`) which is normal Ring-3 access.

6. **PID 1 vm_space=None** — handled at `kernel/src/proc/mod.rs:2937` via `let space = p.vm_space.as_mut()?;` returning `None` early. The clone-arm caller at `kernel/src/subsys/linux/syscall.rs:2324-2329` treats `None` as a logged warning and leaves `tls_base = 0` (pre-fix behaviour) rather than masking the OOM as a different errno — correct fail-open posture. Test 264 explicitly exercises this lookup-miss path (`kernel/src/test_runner.rs:35145-35160`).

### Defense-in-depth note (non-blocking)

The `[vfork-tls]` page is `PAGE_PRESENT | PAGE_WRITABLE | PAGE_USER | PAGE_NO_EXECUTE` in the parent's CR3. The NX bit and user bit are correct. The `WRITABLE` bit is required (musl writes through `%fs:0x40` etc. after the initial setup). One future hardening worth a follow-up: if `[vfork-stack]` and `[vfork-tls]` were placed in a shared VA range only the calling vfork-child thread can resolve (e.g. tagged via PCID once that lands), the multi-threaded-parent sibling-thread theoretical attack window would close. This is **not blocking** because the same shape already exists for `[vfork-stack]` from PR #311 — fixing it in scope here would expand the diff beyond what the demo gate requires.

### Compliance

- All references in comments/commit cite public sources only: POSIX `vfork(2)` / `posix_spawn(3)` / `pthread_setcancelstate(3)`, System V x86_64 ABI §3.4.2, ELF gABI §3.5, Intel SDM Vol. 3A §3.4.4 / §4.6 / §6.8.
- Diff is 480 LOC across 5 files — over the soft 150 LOC target but well within the 2× burst envelope. The four-file split (proc/mod.rs allocator, linux/syscall.rs caller, syscall/mod.rs cleanup, test_runner.rs regression) is the natural decomposition; splitting further would force the allocator and its callers across PRs.
- Test 264 added (`alloc_vfork_child_tls layout`) with five-case coverage (base alignment, self-pointer, canary=0, cancel-state=1, cleanup-unmaps). Verdict supported by the firefox-test KVM run reaching sc=741 from sc=0 with no `0x7f000005b524` fault.

LGTM to merge.

PR #237, opened 2026-05-15
kernel/mm+idt: unified VmSpace generation concurrency model for the entire PFH (W216)

me1iissa:
## Summary

- Adds `VmSpace::generation: Arc<AtomicU64>` bumped on every VMA-list / PTE bulk-rewrite mutator.  Every PFH install arm samples the counter post-revalidate and re-checks it immediately before each install; mismatch ⇒ abort, free the freshly-allocated frames, return false so the user re-faults against the new VMA.
- **Unified concurrency model** — replaces the per-arm whack-a-mole pattern (PRs #231 PMM-quarantine, #234 mm_sem.write, #238 cache-hit surgical).  One mechanism covers all PFH install arms; one decision point for "is the snapshot still authoritative?".
- Closes the W216 5th systemic page-aliasing class (W215 FAULT/PHYS cluster, libxul+0x4b*-region) without structural changes to hot kernel paths.

## Coverage matrix

| Arm | Install kind | idt.rs lines | Gen-check | Diagnostic counter |
|-----|--------------|--------------|-----------|--------------------|
| A. Copy-on-Write (write+present) | `pmm::alloc_page` + copy + install | 970-985 | yes | `[PF/gen-abort] COW-COPY` |
| B. NX fixup | PTE bit flip only — no install | 1042-1043 | n/a (no aliasing) | — |
| C. Cache-hit private-copy | `pmm::alloc_page` + copy + install | 1212-1227 | yes | `[PF/gen-abort] CACHE-PRIVATE` |
| D. Cache-hit alias | guard ref promoted to PTE ref + install | 1267-1268 | yes | `[PF/gen-abort] CACHE-ALIAS` |
| E. Readahead install loop | up-to-32 `cache::insert` + install | 1542-1644 | yes (per-iteration) | `[PF/gen-abort] READAHEAD` |
| F. Single-page fallback | one `cache::insert` + install | 1668-1872 | yes | `[PF/gen-abort] SINGLE-PAGE` |
| G. Anonymous zero-fill | `pmm::alloc_page` + zero + install | 1899-1904 | yes | `[PF/gen-abort] ANON` |
| H. Device MMIO | identity-map fixed BAR phys | 1910-1914 | n/a (no PMM allocation) | — |

6 of 8 arms get the check; arms B and H cannot exhibit page aliasing by construction.

## The unified mechanism

```text
mutator side                          fault-path arm
------------                          --------------
                                      gen_at_start = generation.load(Acquire)
[...]

PR #327, opened 2026-05-19
kernel: ELF-WRITE-TRACE diagnostic for ld-musl .data.rel.ro corruption (W215-aliasing axis-N investigation)

me1iissa:
## Summary

Investigation-only diagnostic to identify the kernel-mode write site that corrupts the ld-musl `.data.rel.ro` slot at user VA `0x7F00_0003_7e18` during the `CLONE_VM | CLONE_VFORK` parent-block window — the new gate after PR #326 (`40317f1`) unblocked the musl posix_spawn child path.

PSE's end-to-end trace shows the child (PID 2) now reaches `exit_group(0)` cleanly after 741 syscalls. The parent (PID 1, TID 2) then runs six cleanup syscalls and `#GP`s at user RIP `0x7f000001c7f9` (musl `a_crash`: `hlt; ret`). On-fault diagnostic emits `[GPF-DBG] ldlinux[0x37e18]=0x00000001b941ffff (should=0x7F00_0000_1640)` — a function-pointer slot in the dynamic linker's `.data.rel.ro` page has been overwritten with garbage. Parent stack canary is byte-identical pre/post vfork; corruption is at a fixed user VA, not on the stack.

Per project memory `project_w215_saga_antipattern_2026_05_16`, **five prior W215 iterations** landed real fixes for real writers that did NOT close the bug (PRs #203 / #208 / #226 / #230 / #237 / #254). The discipline that breaks the cycle is *diagnostic-first*: name the kernel-mode RIP at the moment of the store before proposing any fix.

## What this PR adds

- `kernel/src/subsys/linux/elf_write_trace.rs` (new): wraps the vfork-window with an enter/exit pair that resolves `0x7F00_0003_7e18` to a phys via the parent's CR3, arms a hardware write-only watchpoint on the 8-byte qword via the existing W215 DR0–DR3 plumbing, and snapshots the surrounding 64 bytes for an enter/exit diff. Any kernel-mode store to the watched range fires `#DB` (Intel SDM Vol. 3A §6.15 vector 1) and is reported via `[W215/DR-WATCH-FIRE] slot=… cpu=… rip=KERNEL_RIP cs=… cr3=…`.
- `kernel/src/arch/x86_64/debug_reg.rs`: new `arm_phys_slot_watchpoint(phys, off, len)` for slot-precise arming inside a frame (the existing `arm_phys_watchpoint` only handles 4 KiB-aligned 8-byte windows). Validates `len ∈ {1,2,4,8}`, natural alignment, `(off + len) ≤ 4096` per Intel SDM Vol. 3B §17.2.4 Table 17-2.
- `kernel/src/subsys/linux/syscall.rs`: wire `enter_window` / `exit_window` around `schedule()` in both clone (56) and clone3 (435) `CLONE_VFORK` paths. Cohabits with the existing `vfork-canary-diag` snapshot pair.
- `kernel/Cargo.toml`: new feature `elf-write-trace = ["w215-diag", "firefox-test"]`.

## Diff size

- 5 files, +536 LOC.
- 1 new file (~280 LOC of which ~170 is doc-comment block).
- ~90 LOC of new code in `debug_reg.rs` (new arming variant).
- ~30 LOC of glue in `syscall.rs` (two cfg-gated blocks at the existing vfork wiring points).
- ~13 LOC of feature-flag declaration in `Cargo.toml`.

Within the 50-200 LOC target stretched modestly (1.4×) for one extra subsystem (the slot-precise arming variant) that the diagnostic depends on but didn't exist yet. The base diagnostic itself is well within budget.

## What this PR does NOT do

- **No fix.** Per the W215-saga discipline, this is investigation-only. Phase 2 (the surgical fix) is gated on the diagnostic identifying a real kernel-mode write site by RIP and will be a separate PR.

## Build matrix

- `default` — unchanged (byte-identical)
- `--features w215-diag` — unchanged
- `--features firefox-test` — unchanged
- `--features firefox-test,w215-diag,elf-write-trace,vfork-canary-diag` — new diagnostic active

## Test plan

- [x] `cargo +nightly check` clean across default / firefox-test / w215-diag / full feature stack (all four config matrix entries).
- [ ] Boot firefox-test with `--features firefox-test,w215-diag,elf-write-trace,vfork-canary-diag` and capture `[ELF-WRITE-TRACE/ENTER]`, `[ELF-WRITE-TRACE/EXIT]`, and any `[W215/DR-WATCH-FIRE]` line during the parent-block window.
- [ ] Confirm `[ELF-WRITE-TRACE/EXIT] changed=true` matches the existing `[GPF-DBG] ldlinux[0x37e18]` post-corruption value.
- [ ] Identify the kernel-mode RIP from the `[W215/DR-WATCH-FIRE]` line and report which hypothesis it confirms (H_VFORK_CLEANUP_OVERREACH / H_EXECVE_FRAME_FREE / H_TLB_ALIAS_RESIDUE / OTHER).

## Refs

- Intel SDM Vol. 3B §17.2.4 (Debug Address Registers DR0–DR3)
- Intel SDM Vol. 3B §17.2.5 (Debug Status / Control Registers DR6 / DR7)
- Intel SDM Vol. 3A §6.15 (vector 1, `#DB`)
- System V AMD64 ABI §6.4 (stack protector)
- POSIX vfork(2), clone(2)
- musl libc: https://musl.libc.org/

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -151,6 +151,19 @@ vfs-debug = []
 # unconditional pubs but heavily exercised under syscall-trace).
 # See docs/HARNESS.md and POSIX vfork(2) / clone(2) / Intel SDM Vol. 3A §6.8.
 vfork-canary-diag = ["firefox-test", "syscall-trace"]
+# ELF write-trace diagnostic (kernel/src/subsys/linux/elf_write_trace.rs).
+# Arms a hardware write-only watchpoint on the ld-musl `.data.rel.ro`
+# slot at user VA `0x7F00_0003_7e18` for the duration of the
+# `CLONE_VM | CLONE_VFORK` parent-block window, plus snapshots the
+# surrounding bytes before/after.  The watchpoint piggy-backs on the
+# W215 four-slot DR0–DR3 plumbing in `arch/x86_64/debug_reg.rs`; any
+# kernel-mode write to the watched window fires `#DB` (Intel SDM Vol.
+# 3A §6.15) and emits `[W215/DR-WATCH-FIRE] slot=… cpu=… rip=… cr3=…`
+# carrying the writer's RIP.  Investigative tool only — gated so master
+# builds are byte-identical.  See POSIX vfork(2) / clone(2), Intel SDM
+# Vol. 3B §17.2.4 / §17.2.5, System V AMD64 ABI §6.4.  Implies
+# `w215-diag` (DR0–DR3 infra) and `firefox-test` (trace formatters).
+elf-write-trace = ["w215-diag", "firefox-test"]
 # LLVM source-based code-coverage instrumentation (test-coverage audit
 # session 5 / docs/TEST_COVERAGE_AUDIT_2026-05-16.md).  When enabled, the
 # kernel must additionally be built with `-C instrument-coverage` (see
--- a/kernel/src/arch/x86_64/debug_reg.rs
+++ b/kernel/src/arch/x86_64/debug_reg.rs
@@ -633,6 +633,84 @@ pub enum ArmPhysResult {
     PoolExhausted,
 }
 
+/// Manually arm a write-only watchpoint on `PHYS_OFF + phys + off` where
+/// `phys` is the base of the containing 4 KiB physical frame and `off`
+/// is a byte offset in `[0, 4096)`.  `len` must be 1, 2, 4, or 8 per
+/// Intel SDM Vol. 3B §17.2.4 Table 17-2.  Used by the ELF write-trace
+/// diagnostic (`subsys/linux/elf_write_trace.rs`) to watch a specific
+/// 8-byte slot within an ld-musl `.data.rel.ro` page.
+///
+/// Slot selection: prefer DR1/DR2/DR3 (the pre-insert pool) so a manual
+/// arm doesn't clobber the CRC walker's post-hoc slot (DR0).  Fall back
+/// to DR0 only if all three pre-insert slots are busy.
+///
+/// Cross-CPU sync is the same lazy-gen protocol used by
+/// `arm_write_watchpoint` / `arm_preinsert_watchpoint`: bump
+/// `SYNC_GENERATION` and program our own DRs; peer CPUs pick up the
+/// change in their next `apply_pending_if_stale()` (≤ one timer tick).
+pub fn arm_phys_slot_watchpoint(phys: u64, off: u64, len: u8) -> ArmPhysResult {
+    // Validate page alignment of the frame base.
+    if phys & 0xFFF != 0 {
+        return ArmPhysResult::NotAligned;
+    }
+    // Validate offset + length stays inside the 4 KiB frame.  Per Intel
+    // SDM Vol. 3B §17.2.4 Table 17-2 the DR address may be any byte; we
+    // restrict to one-frame windows so the diagnostic doesn't mis-report
+    // when a write to the adjacent frame's leading bytes fires.
+    if off >= 4096 || off + (len as u64) > 4096 {
+        return ArmPhysResult::NotAligned;
+    }
+    // Width must be one of {1, 2, 4, 8}.
+    match len {
+        1 | 2 | 4 | 8 => {}
+        _ => return ArmPhysResult::NotAligned,
+    }
+    // The watch address must be naturally aligned to `len` per the same
+    // table; reject otherwise (a misaligned arm silently widens via the
+    // Intel-defined LEN encoding and would catch unrelated writes).
+    if off & (len as u64 - 1) != 0 {
+        return ArmPhysResult::NotAligned;
+    }
+    // Validate that the frame falls inside installed RAM.
+    let (total_pages, _) = crate::mm::pmm::stats();
+    let ram_top = total_pages.saturating_mul(crate::mm::pmm::PAGE_SIZE as u64);
+    if phys >= ram_top {
+        return ArmPhysResult::OutOfRange;
+    }
+
+    const PHYS_OFF: u64 = 0xFFFF_8000_0000_0000;
+    let linear = PHYS_OFF.wrapping_add(phys).wrapping_add(off);
+    let inode: u64 = 0;
+    let file_offset: u64 = off;
+
+    let mut armed_slot: Option<u8> = None;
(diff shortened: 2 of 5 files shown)


me1iissa:
## /review verdict — APPROVE

Diagnostic-only investigation infrastructure for the W215-aliasing axis-N corruption of the ld-musl `.data.rel.ro` slot. No security regressions; saga discipline (`diagnostic-first, no fix attempted`) is honoured. Master builds are byte-identical without `--features elf-write-trace`. Approving for merge.

The audit hit-list is summarised below. Two diagnostic-quality nits are non-blocking but worth folding into the Phase 2 fix PR or a follow-up tidy.

### Audit cross-walk vs the requested questions

**1. `arm_phys_slot_watchpoint(phys, off, len)` vs Intel SDM Vol 3B §17.2.4 Table 17-2 — PASS**

All four checks present and correctly ordered:
- `phys & 0xFFF != 0 → NotAligned` (frame-base alignment)
- `off >= 4096 || off + len > 4096 → NotAligned` (frame-bound)
- `len ∉ {1, 2, 4, 8} → NotAligned` (Table 17-2 LEN encoding)
- `off & (len - 1) != 0 → NotAligned` (natural-alignment per Table 17-2)
- W-only confirmed in `dr7_bits_for_slot`: `rw = 0b01` (write-only), no R/W variant exposed via the new entry point.

Minor: `off + len > 4096` returns `NotAligned` when `OutOfRange` would be semantically more accurate. Non-blocking.

**2. DR0-DR3 arming/disarming symmetry — PASS with one acknowledged gap**

Normal-path symmetry holds (clone-56 and clone3-435 paths both bracket `schedule()` with `enter_window` / `exit_window`).

Three asymmetric-exit scenarios:
- **Panic between enter and exit**: kernel is dying, slot leak is cosmetic.
- **`exit_window` reached without a DR fire**: slot stays armed until the next `enter_window` re-CASes or the next post-window kernel store accidentally trips it ("bonus data point" per author). After at most 4 such windows, `arm_phys_slot_watchpoint` returns `PoolExhausted` and the trace logs `err=pool_exhausted`. Acceptable for a hand-driven investigation; **consider adding an explicit `disarm_phys_slot_watchpoint` and calling it from `exit_window`** when Phase 2 promotes any of this code.
- **Concurrent vforks on different parent PIDs**: a single global `WINDOW_ACTIVE` atomic. Second `enter_window` hits the `[ELF-WRITE-TRACE/REENTRY]` path and returns; the second `exit_window` then opportunistically clears `WINDOW_ACTIVE` for the FIRST window. Investigation-tool incorrectness, not a security gap. Demo workload only does one vfork at a time.

Cross-CPU migration: per `apply_pending_if_stale` (called from each CPU's timer ISR), peer-CPU DR programming converges within ≤ 10 ms at `TICK_HZ = 100`. A kernel-mode write on a peer CPU within that window before convergence COULD slip through unobserved. Acknowledged in the existing module docs for the W215 plumbing.

The comment in `read_qword_via` claims "we are still pinned by the syscall handler" — this is functionally fine (CLONE_VM children share the parent's CR3, so `get_cr3()` at exit time returns the same PML4 phys regardless of which CPU `schedule()` returned us on), but the comment's "pinned" rationale is misleading. The thread CAN migrate. Suggest rewording to "CR3 is preserved across migration because CLONE_VM children share the parent's address space."

**3. No leakage outside the vfork window — see (2). One investigation-tool false-positive risk:**

`.data.rel.ro` is a shared file-backed page until first write; multiple processes mapping ld-musl could share the same physical frame. A concurrent process (not the one being investigated) writing to its own copy of the slot would fire on the watched phys before it CoWs. Diagnostic noise, not a security bug.

**4. Kernel-VA-only reads — PASS**

`read_qword_via` and `read_bytes_via` both go through `PHYS_OFF + phys` (the kernel's identity map of physical RAM). No user-mode pointer dereferences from kernel mode → no STAC/CLAC SMAP brackets needed. `get_cr3()` masks with `ADDR_MASK` so PCID bits don't leak into the page-walk.

**5. Gating discipline — PASS for master, with one annotation:**

- `elf_write_trace` module declaration: `#[cfg(feature = "elf-write-trace")]` (`subsys/linux/mod.rs`)
- syscall.rs wiring: both clone-56 and clone3-435 enter/exit blocks are `#[cfg(feature = "elf-write-trace")]` gated
- `arm_phys_slot_watchpoint` in `debug_reg.rs` is publicly exposed but only callable from gated callers; the function compiles unconditionally (it's a small chunk and doesn't pull additional infrastructure).
- Cargo.toml `default = []` confirmed; default builds and `--features firefox-test` builds do NOT pull `elf-write-trace` and therefore do NOT pull `w215-diag`.

Re question 5's "Verify the feature does NOT imply `w215-diag`": **it DOES imply `w215-diag`, intentionally**, because the DR0-DR3 plumbing this diagnostic relies on lives in `w215-diag`. The PR description is explicit about this. The relevant property — that default and firefox-test builds remain free of the 2 MiB BSS — DOES hold: only `--features elf-write-trace` pulls the chain `elf-write-trace → w215-diag → firefox-test`.

**6. vfork_diag side-finding (`#PF` at CR2=0x8ff8) — UNRELATED**

PR #327's `elf_write_trace` module does not call into `vfork_diag` and does not perform user-pointer derefs. The `read_userland_qword_raw(user_rbp)` path that takes the fault is in `vfork_diag.rs` (from PR #325). PR #327's gated wiring sits adjacent in `syscall.rs` but the two diagnostics are independent. No overlap; this PR does not exacerbate that fault.

### Saga discipline check — PASS

- No fix code in the diff (`grep '^\+.*fix\|^\+.*workaround'` returns only doc references).
- Module docstring explicitly cites the saga anti-pattern; the diagnostic's stated goal is to name the kernel-mode RIP before Phase 2.
- The earlier-merged stack (PR #325 vfork canary, PR #326 TLS provisioning) is correctly excluded from this PR's commit boundary; `git diff 2195e51^..2195e51` is +536 LOC across the 5 files claimed.

### References

- Intel SDM Vol. 3B §17.2.4 (DR0-DR3 layout), §17.2.5 (DR6/DR7 status/control), Table 17-2 (LEN encoding)
- Intel SDM Vol. 3A §6.15 (vector 1, `#DB`)
- POSIX `vfork(2)` / `clone(2)`
- System V AMD64 ABI §6.4 (stack-protector contract)

PR #329, opened 2026-05-19
kernel: full parent-stack-VMA provenance dump (W215 axis-N+1 parallel diagnostic)

me1iissa:
## Summary

Parallel diagnostic to aether-kernel-engineer's primary W215 axis-N+1 dispatch (return-address frame scan).  Tech-lead's binding verdict on PR #328's 10-trial soak was that the SSP-failing parent frame sits *upstream* of the RSP+8 KiB capture window (musl's `__clone` stub strips RBP at syscall-entry, so the RBP-chain walk terminates at depth 1).  This dispatch widens the page-provenance capture to **every present page in the parent's `[stack]` VMA(s)** at both `pre_block.clone` and `post_wake.clone`.

No file conflict with aether's primary — that dispatch lives in `proc/stack_walk.rs` + SSP-scan logic; this one stays in `subsys/linux/vfork_diag.rs` page-provenance space.

## What lands

New helper `snapshot_stack_page_provenance(label, parent_pid, parent_tid)`:

- Looks up parent's `VmSpace` in `PROCESS_TABLE`, captures up to 4 stack-named VMAs into a fixed-size on-stack array, drops the lock before any serial I/O.
- For each VMA, iterates every 4 KiB page via `lookup_pte_in(parent.cr3, va)`.  Skips absent pages; for present pages records `(page_va, page_pa, refcount, pte, present, writable)` into a fixed-size `[StackPageProv; 512]` (no allocator traffic).
- Emits one `[STACK-PAGE-PROV-FULL]` line per present page, plus a summary line with `(vma_count, total_present, total_writable, total_refcount_sum, emitted_rows, cap, overflowed)`.

Wired into both `clone(56)` and `clone3(435)` paths in `subsys/linux/syscall.rs` under the existing `#[cfg(feature = "vfork-canary-diag")]` block.

## W215-class discriminators

| Signature | Meaning |
|---|---|
| same `page_va`, different `page_pa` PRE→POST | aliasing on a stack page outside the narrow window |
| same `(page_va, page_pa)`, different `refcount` | `pte_share_count` accounting race |
| `present=true → false → true` | cache eviction inside vfork window |
| `writable=true → false → true` | PR #270 `pte_share_count` race signature |

## Test plan

[...]

PR #330, opened 2026-05-19
kernel: SSP-epilogue scan + full-stack-VMA provenance (W215 axis-N+1 widened diagnostic)

me1iissa:
## Summary

Three new diagnostic channels around the `CLONE_VM|CLONE_VFORK` parent-block window, gated on `vfork-canary-diag`:

- **Method 1 — `[STACK-SSP-FRAME]`**: byte-pattern scan of the parent's stack VMA for SSP-epilogue `48 33 04 25 28 00 00 00` (per System V x86_64 psABI §6.4).  Locates the SSP-instrumented caller frame whose `[rbp-8]` canary slot the function epilogue will re-read — without any RBP-chain dependency.
- **Method 2 — `[STACK-PAGE-CRC]`**: per-page Fletcher-32 (RFC 1146) over every present page in the parent's stack VMA, pre-block and post-wake.  Delta naming whole-page swap (axis-B aliasing) vs in-window byte change vs clean window.
- **Method 3 — `[STACK-CANARY-WATCH]`**: after Method 1 locates a frame, arm a write-only hardware watchpoint (Intel SDM Vol. 3B §17.2.4 DR0–DR3) on its canary slot's user linear address for the duration of vfork.  Fires emit `[W215/DR-WATCH-FIRE]` with writer RIP.

Both clone(2) and clone3(2) call sites are wired.  `arch::x86_64::debug_reg::arm_user_va_watchpoint` is the new generic primitive for user-VA DR arms.

`get_parent_user_frame(parent_tid)` reads RSP/RBP/RIP from the parent's kernel-stack top per the `syscall_entry` save layout — stable across pre/post since `schedule()` does not touch the saved frame.  This fixes the per-CPU `frame_rsp`-overwrite issue that left the prior diagnostic's POST snapshot in `state=no_user_frame`.

Diagnostic-only.  Default / firefox-test builds byte-identical when `vfork-canary-diag` is OFF.

## Test plan

- [x] Build with `vfork-canary-diag,w215-diag` features — clean.
- [x] Single-trial KVM soak: vfork window enters; all three channels emit; Methods 1+2 work in POST (where the prior diagnostic showed `no_user_frame`).
- [ ] 10-trial KVM soak (in progress).

## Exit-bucket classification (per dispatch)

- **(A)** Method 1 located frame AND `canary=0` pre_block → corruption pre-existed.
- **(B)** Method 1 located frame AND canary changed pre→post → in-window write; Method 2 page-CRC delta names the byte band.
- **(C)** Method 2 CRC delta WITHOUT a single SSP frame located → user-side corruption likely.
- **(D)** NO change AND canary at pre IS the expected non-zero value → genuine userspace-side hypothesis (i); capture user-space STAC/CLAC store next.

## Refs

POSIX vfork(2) / clone(2); System V x86_64 psABI §6.4 (SSP); System V AMD64 ABI §3.4.1 / §3.4.5.2; Intel SDM Vol. 3A §6.8 (IA32_FS_BASE), Vol. 3B §17.2.4 / §17.2.5 (DR0-DR3, #DB); RFC 1146 §2.

### DIFF
--- a/kernel/src/arch/x86_64/debug_reg.rs
+++ b/kernel/src/arch/x86_64/debug_reg.rs
@@ -711,6 +711,72 @@ pub fn arm_phys_slot_watchpoint(phys: u64, off: u64, len: u8) -> ArmPhysResult {
     ArmPhysResult::Armed(slot)
 }
 
+/// Arm a write-only watchpoint on a USER linear address (low-canonical half).
+///
+/// Per Intel SDM Vol. 3B §17.2.4 (DR0–DR3) the linear-address compare runs
+/// on every memory access regardless of CPL, so a user-VA arm catches both
+/// kernel-mode stores via `PHYS_OFF + phys` AND user-mode stores via the
+/// same user VA from any thread whose CR3 has that VA mapped (e.g. the
+/// parent's sibling threads during a `CLONE_VFORK` window).
+///
+/// Used by the SSP-epilogue scan (axis-N+1 diagnostic in
+/// `subsys/linux/vfork_diag.rs`) to watch a specific user-space canary
+/// slot located by byte-pattern scan.
+///
+/// Slot selection mirrors `arm_phys_slot_watchpoint`: prefer DR1/DR2/DR3
+/// so the CRC walker's DR0 is not stolen; fall back to DR0 only if all
+/// three pre-insert slots are busy.
+///
+/// `len` is restricted to {1, 2, 4, 8} per Intel SDM Vol. 3B §17.2.4
+/// Table 17-2 and must be naturally-aligned to `len`.  `linear` must lie
+/// in the low canonical user half (below `USER_VA_LIMIT = 0x8000_0000_0000`).
+///
+/// Returns the slot index (0..=3) on success or `None` if the parameters
+/// were invalid or all four slots are busy.
+pub fn arm_user_va_watchpoint(linear: u64, len: u8) -> Option<u8> {
+    // Reject lengths the DR LEN field cannot encode.
+    match len {
+        1 | 2 | 4 | 8 => {}
+        _ => return None,
+    }
+    // Reject misaligned watch address per Intel SDM §17.2.4 Table 17-2.
+    if linear & (len as u64 - 1) != 0 {
+        return None;
+    }
+    // Refuse non-canonical / kernel-half addresses — the parent's stack
+    // canary always lives in the low canonical half.
+    const USER_VA_LIMIT: u64 = 0x0000_8000_0000_0000;
+    if linear == 0 || linear >= USER_VA_LIMIT {
+        return None;
+    }
+    let phys: u64 = 0;          // unknown / unused for the [W215/DR-ARM] line
+    let inode: u64 = 0;
+    let file_offset: u64 = 0;
+    let mut armed_slot: Option<u8> = None;
+    for slot in [1usize, 2, 3, 0] {
+        if try_arm_slot(slot, linear, len, phys, inode, file_offset) {
+            armed_slot = Some(slot as u8);
+            break;
+        }
+    }
+    let slot = armed_slot?;
+    ARM_COUNT.fetch_add(1, Ordering::Relaxed);
+    crate::serial_println!(
+        "[W215/DR-ARM] slot={} linear={:#x} phys=0 inode=0 offset=0 kind=user_va len={}",
+        slot, linear, len,
+    );
+    SYNC_GENERATION.fetch_add(1, Ordering::Release);
+    program_local_drs();
+    let cpu = super::apic::cpu_index();
+    if cpu < super::apic::MAX_CPUS {
+        LOCAL_SYNC_GENERATION[cpu].store(
+            SYNC_GENERATION.load(Ordering::Acquire),
+            Ordering::Release,
+        );
+    }
+    Some(slot)
+}
+
 /// Manually arm a write-only watchpoint on `PHYS_OFF + phys` from a kdb
 /// command, bypassing the `cache::insert` pre-arm key filter.  Useful when
 /// a corrupted phys has already been observed by the CRC walker but the
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -2361,10 +2361,21 @@ fn dispatch_body(num: u64, arg1: u64, arg2: u64, arg3: u64, arg4: u64, arg5: u64
                             vfork_canary_snapshot("pre_block.clone", pid as u32, parent_tid);
                             // Three-channel snapshot + open the sibling-syscall
                             // window.  Diagnostic-only; see `vfork_diag.rs`.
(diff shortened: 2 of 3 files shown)


me1iissa:
## /review — CHANGES-REQUESTED (small, blocking)

Diagnostic infrastructure is sound and the empirical soak data (bucket D, 0 SSP frames found, identical pre/post CRCs across 4 stack pages, no DR slot exhaustion) confirms the channels behave as designed and are well-bounded under real workload. The new `arm_user_va_watchpoint` primitive correctly mirrors the existing `arm_phys_slot_watchpoint` slot-claim + lazy-gen cross-CPU sync protocol (per Intel SDM Vol. 3B §17.2.4 / §17.2.5), and the doc-comment correctly notes that linear-address compares fire regardless of CPL/CR3 — the "fires from sibling processes" property is documented and acceptable for diagnostic context. The `get_parent_user_frame` offsets (`kstack_top - 8 / -16 / -32` → user_rsp / RCX / rbp) correctly index slots 14 / 13 / 11 of the 15-slot syscall_entry save frame (verified against `kernel/src/syscall/mod.rs:801` naked_asm push sequence). 569 LOC / 1.6× burst is justified by three independent methods + extensive doc comments.

One blocking gating bug and a handful of defensive-depth nits below.

### CHANGES-REQUESTED (must-fix before merge)

**G1. Cargo.toml — `vfork-canary-diag` does not pull in `w215-diag`.** `kernel/Cargo.toml:153` declares `vfork-canary-diag = ["firefox-test", "syscall-trace"]`. The new code in `vfork_diag.rs:706` calls `crate::arch::x86_64::debug_reg::arm_user_va_watchpoint`, but `debug_reg.rs:82` gates the entire module behind `#![cfg(feature = "w215-diag")]`. A build with `--features vfork-canary-diag` alone will fail to compile (the function does not exist when `w215-diag` is off). The current soak passed because it was built with `vfork-canary-diag,w215-diag` together, but the documented feature flag contract is broken.

Fix: `vfork-canary-diag = ["firefox-test", "syscall-trace", "w215-diag"]`. One line.

### Defensive-depth nits (recommended before merge, but not blocking)

**N1. `get_parent_user_frame` raw kernel-VA deref without bound check** (CWE-476 hardening). `vfork_diag.rs:341–343` does `unsafe { *((kstack_top - 8) as *const u64) }` etc. with only a `kernel_stack_base == 0` check. If a future regression in thread-init leaves `kernel_stack_base` set to a non-zero but non-canonical or user-half value, this is an unchecked deref of attacker-influenceable memory. Add before the unsafe block:
```rust
if kstack_top < 0xFFFF_8000_0000_0000 || kstack_top < t.kernel_stack_base {
    return None;
}
```
Costs two compares; converts a future thread-init bug from "diagnostic derefs garbage" to "diagnostic silently skips."

**N2. `read_user_bytes_via_phys` missing user-half upper bound.** `vfork_diag.rs:380` walks `va + i` without clamping against `USER_ADDR_END`. Current callers always pass user-half VAs, but a future misuse could pass a kernel-half VA and the function would happily walk the page tables and read from `PHYS_OFF + phys`, leaking kernel data into the serial log. Add `if cur >= USER_ADDR_END { break; }` at the top of the loop body.

**N3. Fletcher-32 doc-comment incorrectly cites RFC 1146.** `vfork_diag.rs:363` implements a byte-wise mod-65535 checksum. RFC 1146 §2 specifies Fletcher over 16-bit words with `i` as a 16-bit word index. The byte-wise variant is a fine diagnostic CRC and is what you want here, but the doc should read "Fletcher-32-style byte-wise checksum (mod 65535)" rather than "Fletcher-32 (RFC 1146)". No functional change; pure attribution accuracy.

**N4. `ssp_epilogue_scan` PROCESS_TABLE-lock churn in the inner loop.** Each candidate qword that passes the canonical-VA filter triggers a `lookup_user_vma(parent_pid, saved_rip)` call (`vfork_diag.rs:523`), each acquiring `PROCESS_TABLE.lock()`. The walk is bounded at 256 KiB / 8 B = 32,768 candidate slots, so worst case is 32,768 lock acquisitions per vfork window — observable contention on the global PROCESS_TABLE during the vfork window itself. The soak shows 0 SSP matches today so this doesn't bite empirically, but a future libxul build that does emit canonical user VAs in stack noise would expose it. Two reasonable fixes: (a) snapshot the parent's VMA list once at scan entry under the lock and walk the snapshot, or (b) cache the previous candidate's `(vma_base, vma_end)` and skip the lookup if the current saved_rip falls in the same VMA. Not blocking for a diagnostic-only path.

**N5. DR slot leak in Method 3 if the watch never fires.** `arm_canary_watch_if_located` arms a slot (DR1/2/3 preferred), but `reset_canary_target()` only clears the static `LATEST_CANARY_SLOT` — it does NOT call any debug_reg disarm primitive. If the canary is never written during the vfork window (which is the dispositive bucket-D outcome you want), the slot stays armed across windows and eventually exhausts the pool after a few iterations. Empirically not biting today (Method 1 returns 0 frames so Method 3 returns `state=no_target` and never arms), but worth fixing for the future case where Method 1 starts hitting. Track the slot armed by Method 3 in a static and clear it from `reset_canary_target` via a new `debug_reg::disarm_slot(u8)` primitive.

**N6. No regression test for the new kernel-VA deref path.** Per security-engineer methodology, the unsafe block in `get_parent_user_frame` deserves a `test_runner.rs` case that constructs a fake Thread with `kernel_stack_base = 0`, with a non-canonical `kernel_stack_base`, and with the canonical-but-valid happy path, asserting the safe path. Diagnostic-only; not blocking.

### Verdict

CHANGES-REQUESTED on G1 (one-line Cargo.toml fix) only. N1–N6 are defensive-depth recommendations; N1+N2 are cheap and worth taking in the same revision since they harden the new unsafe-block surface against future regressions. The empirical soak data (10/20 combined trials disconfirming kernel-side aliasing) is dispositive and the diagnostic infra is genuinely valuable for future investigations as Aether's report notes — once G1 is fixed, this is a clean merge.

Refs: Intel SDM Vol. 3B §17.2.4 (DR0–DR3 linear-address compare), §17.2.5 (DR6/DR7/#DB); System V x86_64 psABI §6.4 (SSP / `__stack_chk_guard`); System V AMD64 ABI §3.4.1 (frame-pointer chain); POSIX vfork(2); RFC 1146 §2 (Fletcher checksum family); CWE-476 (NULL Pointer Dereference) — pattern for N1.

me1iissa:
Closing per tech-lead cross-walk verdict 2026-05-19 (project memory: project_w215_axis_n_exhausted_2026_05_19.md): the kernel-side user-stack page-aliasing axis is exhausted by 20/20 trials across PR #328 + PR #329 + this PR. Cross-trial fs:0x28 backing phys page differs across trials while the fault is byte-deterministic — incompatible with page-aliasing. Continuing aliasing diagnostics would replay the W215 saga anti-pattern. Diagnostic infrastructure from PR #328 (merged) is sufficient. Pivoting to userspace walk-backward from RIP 0x7f000001c7f9.

PR #332, opened 2026-05-19
kernel: axis-O bisect (EXIT-FINAL snapshot for post-vfork stack-page corruption)

me1iissa:
## Summary

Phase 1 diagnostic for the W215 axis-O class established by PR #331's
3-trial dispositive: the parent's user-stack page (`0x7ffffffee000` in
the libxul `posix_spawn` path) changes contents between the
vfork-window-close `STACK-PAGE-CRC POST` snapshot and the eventual
`#GP` arrival, with identical phys throughout (no aliasing).

This PR adds three things, all gated on the existing `vfork-canary-diag`
feature so non-diagnostic builds remain byte-identical:

- **`EXIT-FINAL` snapshot** — `[STACK-PAGE-PROV]` page-provenance dump
  at the LAST kernel point before SYSRET back to userspace, after the
  existing POST triple.  Localises any mutation that happens between
  the POST sample and the actual return.
- **`arm_post_vfork_stack_watch` / `disarm_post_vfork_stack_watch`** —
  per-page Fletcher-32 (RFC 1146) CRCs of up to 16 4-KiB pages of the
  parent's stack VMA, lock-free via `[AtomicU64; 16]`.
- **`check_post_vfork_stack(tag, tag_val)`** — bisect hook called from
  Linux syscall entry, syscall return, and the `#GP` handler.  Off-path
  cost is one relaxed atomic load + branch.  On the on-path, emits a
  bounded `[POST-VFORK-CRC-DELTA]` line per slot per state change with
  the syscall number / vector, bracketing the writer to a specific
  `SC-ENTER ↔ SC-RET` interval or pinning it on the `#GP` path.

Bisect predicate post-soak:

- `[POST-VFORK-CRC-DELTA] tag=SC-ENTER nr=N` after an earlier
  `SC-RET nr=M` → writer is **userspace** between syscall M and N.
- `[POST-VFORK-CRC-DELTA] tag=SC-RET nr=N` without paired entry delta
  → writer is the **kernel-side handler of syscall N**.
- `[POST-VFORK-CRC-DELTA] tag=#GP` with no earlier delta → writer is
  the **`#GP` handler / IDT entry stub**.
- `phys_now != phys_pre` at any tag → page aliasing (axis-N+1 class,
  not axis-O).

Refs: POSIX vfork(2); Intel SDM Vol. 3A §4.6 / §4.10 / §6.15; System V
x86_64 psABI §6.4; RFC 1146 §1.

## Test plan

- [ ] 3-trial KVM soak with `firefox-test,kdb,vfork-canary-diag,syscall-trace`.
- [ ] Confirm `[POST-VFORK-WATCH-ARM]` fires post-`POST` triple in each trial.
- [ ] Locate the first `[POST-VFORK-CRC-DELTA]` line and apply the bisect
      predicate above to attribute the writer.
- [ ] Phase 2 (~150 LOC fix): once writer attributed, audit and patch
      the offending path.

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -288,6 +288,46 @@ extern "C" fn exception_handler(vector: u64, error_code: u64, frame: &mut Interr
                     crate::serial_println!("[GPF-DBG]   [RSP+{:#03x}]={:#018x}", i*8, val);
                 }
             }
+
+            // Fletcher-32 CRC over the user-stack page containing fault RSP.
+            // Pairs with the `[STACK-CANARY-FINEGRAIN]` and `[STACK-PAGE-PROV]`
+            // snapshots emitted PRE/POST the vfork window by
+            // `subsys::linux::vfork_diag`.  A post-processor compares the
+            // `#GP`-time CRC against the POST snapshot at the same VA to
+            // discriminate:
+            //
+            //   * `crc == POST && phys == POST` → page contents stable from
+            //     POST through `#GP`; the faulting write came from userspace
+            //     after the vfork window closed (userspace memory-safety
+            //     bug — e.g. STL write past end overwriting saved SSP slot).
+            //   * `crc != POST && phys == POST` → kernel mutated the page
+            //     between POST snapshot and `#GP` (axis-O kernel-side
+            //     post-vfork stack-page corruption).
+            //   * `phys != POST`               → page aliasing (axis-N+1
+            //     class — VA stable but a different physical frame is now
+            //     mapped).
+            //
+            // Diagnostic-only; gated behind `vfork-canary-diag` so master
+            // builds remain byte-identical and the trap path retains its
+            // existing fault-immune shape.
+            //
+            // Refs: Intel SDM Vol. 3A §6.15 (#GP), RFC 1146 (Fletcher-32).
+            #[cfg(feature = "vfork-canary-diag")]
+            {
+                let stack_page = rsp & !0xFFFu64;
+                match crate::subsys::linux::vfork_diag::fletcher32_user_page(
+                    cr3, stack_page,
+                ) {
+                    Some((phys, crc)) => crate::serial_println!(
+                        "[GPF-STACK-CRC] tid={} stack_page={:#x} phys={:#x} crc={:#010x}",
+                        tid, stack_page, phys, crc,
+                    ),
+                    None => crate::serial_println!(
+                        "[GPF-STACK-CRC] tid={} stack_page={:#x} state=unmapped",
+                        tid, stack_page,
+                    ),
+                }
+            }
         }
     }
 
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -327,6 +327,20 @@ extern "C" fn exception_handler(vector: u64, error_code: u64, frame: &mut Interr
                         tid, stack_page,
                     ),
                 }
+                // Axis-O bisect: fire the post-vfork stack-watch check at
+                // `#GP` arrival so the final state of every armed slot is
+                // emitted alongside the per-syscall checks.  If the page
+                // first diverges HERE (and not on any prior `SC-ENTER` /
+                // `SC-RET` hook) the writer is in the `#GP` handler / IDT
+                // entry stub itself.  POSIX vfork(2); Intel SDM Vol. 3A §6.15.
+                crate::subsys::linux::vfork_diag::check_post_vfork_stack(
+                    "#GP", vector as u64);
+                // Best-effort disarm.  The `#GP` is the terminating event
+                // for the parent's vfork lifecycle in the dispositive
+                // trials; clearing the watch here prevents a stale arm
+                // from leaking into any subsequent vfork window if the
+                // process somehow continues.  No-op if already disarmed.
+                crate::subsys::linux::vfork_diag::disarm_post_vfork_stack_watch();
             }
         }
     }
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -743,6 +743,16 @@ pub fn dispatch(num: u64, arg1: u64, arg2: u64, arg3: u64, arg4: u64, arg5: u64,
     #[cfg(feature = "vfork-canary-diag")]
     crate::subsys::linux::vfork_diag::maybe_log_sibling_syscall(num);
 
+    // ── Axis-O bisect: post-vfork parent-stack CRC check (entry) ────────────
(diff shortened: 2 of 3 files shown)


me1iissa:
## Security review: APPROVE

Diagnostic-only, well-gated, security-clean. Verdict per the six audit questions:

### Q1: Lock-free atomic snapshot correctness — OK

- Arm: per-slot `Relaxed` stores to `WATCH_VA / WATCH_PHYS / WATCH_CRC`, then `Release` store to `WATCH_ACTIVE_PACKED`.
- Check: `Acquire` load of `WATCH_ACTIVE_PACKED`, then per-slot `Relaxed` loads.
- Correct on x86 TSO and per Rust's memory model: the Release/Acquire pair on `WATCH_ACTIVE_PACKED` establishes happens-before for all prior slot stores. The `(va, phys, crc)` triple is logically inseparable but the design tolerates torn reads because the tid filter (`my_tid != watch_tid → return`) ensures only the parent thread mutates the slot arrays. Sibling threads on other CPUs share CR3 but never write the slots. A single thread cannot race with itself.
- Intel SDM Vol. 3A §8.2 confirms TSO suffices; the explicit ordering documents intent.

### Q2: Hot-path overhead — OK

Off-path cost in `check_post_vfork_stack` is exactly: one `Acquire` load of `WATCH_ACTIVE_PACKED` + one compare + one branch + return. No serial output, no allocation, no formatting. Matches the sibling-syscall tagger pattern. Two such checks added per syscall (SC-ENTER + SC-RET) — total off-path overhead ≈ 2 atomic loads per syscall when no window armed. Acceptable.

### Q3: MAX_DELTAS=32 cap is per-window — OK

`arm_post_vfork_stack_watch` calls `WATCH_DELTA_COUNT.store(0, Relaxed)` at the start of each arm. Per-window reset confirmed. Early-return guard at the top of `check_post_vfork_stack` caps the work. Minor: within a single invocation the per-slot loop can fire up to 16 `fetch_add` calls before the next entry-time guard catches it, so the actual cap is `MAX_DELTAS + WATCH_SLOTS - 1 = 47` in the worst case. Acceptable.

### Q4: SMAP discipline — OK

`fletcher32_user_page` reads via the kernel direct map (`PHYS_OFF + phys = 0xFFFF_8000_0000_0000 + phys`), never the user VA. `virt_to_phys_in` returns the physical address from the page-table walk without ever dereferencing the user VA. No STAC/CLAC bracket needed. The new arm path only adds calls to `fletcher32_user_page` and `get_parent_user_rsp_rbp` — the latter reads from the kernel-stack saved syscall frame (kernel-VA only). No raw user-VA derefs introduced.

### Q5: EXIT_FINAL placement — labeling is imprecise, diagnostic is sound

The claim "LAST kernel point before SYSRET" is **incorrect**. After EXIT-FINAL fires inside `dispatch_body()` (clone-56 / clone3-435 post-wake paths), control returns to `dispatch()` which still runs `ring::end()`, `clear_current_entry()`, `proc_metrics::leave_syscall()`, the `[SC-RET]` emit, and `check_post_vfork_stack(\"SC-RET\", ...)` before SYSRET.

**However**, the bisect remains complete: any kernel write between EXIT-FINAL-snap-and-arm and SYSRET is caught by either (a) the immediately-following SC-RET check, (b) the next SC-ENTER check, or (c) the #GP handler check. The label `EXIT-FINAL` is misleading but the diagnostic chain is sound. Non-blocking; consider renaming to `POST-WAKE-FINAL` or `POST-VFORK-SNAP` in a follow-up.

### Q6: Gating discipline — OK

`vfork_diag.rs` is gated at module level (`#![cfg(feature = \"vfork-canary-diag\")]` line 72) and at the `pub mod` declaration in `mod.rs` line 42. All three new call sites in `syscall.rs` (entry hook, return hook, two arm sites) are wrapped in `#[cfg(feature = \"vfork-canary-diag\")]`. The IDT call sites are inside the existing `#[cfg(feature = \"vfork-canary-diag\")]` block at `idt.rs:315`. Master byte-identical. Symmetric.

---

### Inline issues (non-blocking)

**1. Stale-arm leak across process boundaries** (`vfork_diag.rs`)

If the parent exits without ever reaching `#GP` (the only disarm site), `WATCH_ACTIVE_PACKED` stays non-zero indefinitely. TID values are reused across the system, so a future unrelated process whose thread happens to take the stale TID would pass the tid filter, walk slot VAs against the new process's CR3, and produce phantom `[POST-VFORK-CRC-DELTA]` lines (`fletcher32_user_page(new_cr3, old_va)` would yield a different phys → fire a delta).

Bounded by `MAX_DELTAS=32` so this is a diagnostic-noise risk, not a security risk. Consider adding a disarm in the process-exit path (`sys_exit_group`) in a follow-up.

**2. EXIT-FINAL label misnomer** (Q5) — `POST-WAKE-FINAL` would be more accurate.

**3. Per-invocation cap overrun** (Q3) — actual cap is `MAX_DELTAS + WATCH_SLOTS - 1` worst case. Acceptable for diagnostic code.

---

### Security verdict

- **Attacker model**: this code only compiles in when `vfork-canary-diag` is enabled (debug builds). No production exposure. With the feature enabled, the only new attack surface is the arm path reading parent's CR3 + walking 16 stack pages via `fletcher32_user_page`. All reads are kernel-direct-map; no user-VA dereferences without `virt_to_phys_in` validation; no SMAP brackets needed.
- **CWE check**: considered CWE-362 (race condition) for the lock-free slot updates — Q1 analysis above confirms the tid filter eliminates the race. CWE-1284 (improper validation of specified quantity in input) for `WATCH_SLOTS=16` page walk — the walk uses `saturating_add` and bounds against `armed < WATCH_SLOTS`, fine.
- **Net diff**: +329 LOC, all gated behind a non-default debug feature. No master-build changes.

**Approving.** Once axis-O bisect identifies the writer, the diagnostic code should be removed (or kept behind the gate indefinitely).

me1iissa:
Closing per tech-lead cross-walk verdict 2026-05-19 (project_w215_axis_n_exhausted_2026_05_19.md plus the post-#333 axis-O falsification verdict). The bisect diagnostic served its purpose — qa-engineer's 3/3 BIT-FOR-BIT IDENTICAL re-run on PR #333 base produced dispositive data falsifying axis-O kernel-side stack corruption (0/96 phys mismatches, FS:0x28 and [rbp-8] both PRE==POST 3/3). Real demo gate per tech-lead Q2 verdict is the musl pthread_cond_t._c_seq/_c_waiters family (4-byte adjacent, b28↔b2c alternation), NOT a W101 recurrence (different offset family from glibc __g_refs). Per saga anti-pattern discipline (project_w215_saga_antipattern_2026_05_16): not extending kernel-aliasing diagnostics. Pivoting to abi-compatibility-engineer audit of sys_futex_linux key resolution.

PR #333, opened 2026-05-19
kernel/subsys/linux/vfork_diag: fault-immune user-VA read + non-stale RSP/RBP helper (unblocks axis-O soak)

me1iissa:
## Summary

Fixes two compounding diagnostic-only bugs in `vfork_diag.rs` that were
deterministically crashing the kernel with KERNEL_PAGE_FAULT (CR2=0x8ff8,
RIP=`subsys::linux::vfork_diag::read_userland_qword + 0x1f3`) on every
firefox-test soak with `vfork-canary-diag` enabled, blocking axis-O bisect
arming entirely.

- **Bug 1**: `snapshot_canaries` at line 218 used the per-CPU
  `get_user_rsp_rbp()` helper which reads `frame_rsp` — overwritten by
  every subsequent syscall on the same CPU. Across the `schedule()` call
  inside the vfork window, the slot belonged to a sibling syscall,
  yielding stale `user_rbp=0x9000`. Swapped to the existing
  `get_parent_user_rsp_rbp(parent_tid)` helper (already defined in this
  module), which reads from the parent's kernel-stack saved syscall
  frame and is stable across pre-block and post-wake.
- **Bug 2**: `user_read_u64` only validates range/alignment, not page
  presence. With the stale user_rbp from bug #1, `wrap_canary_addr=0x8ff8`
  passed `validate_user_ptr` but the page was not mapped; the
  STAC-bracketed kernel-mode load took a `#PF` with no `__ex_table`
  recovery and panicked. Rewrote `read_userland_qword{,_raw}` to walk
  CR3 via `mm::vmm::virt_to_phys_in` and read from the kernel direct
  physical map (`PHYS_OFF + phys`) — mirrors `fletcher32_user_page` from
  PR #331. Unmapped pages now report `?` instead of bugchecking.
- Small bonus per PR #331 follow-up nit: `debug_assert!(va < USER_VA_LIMIT)`
  on `fletcher32_user_page` (documented as user-half only).

Diagnostic-only change. Entire module is gated behind
`#[cfg(feature = \"vfork-canary-diag\")]`. Verified the build still
succeeds without the feature flag.

## Test plan

- [x] Build with `--features firefox-test,vfork-canary-diag,w215-diag` succeeds (release)
- [x] Build without `vfork-canary-diag` succeeds (gating intact)
- [x] firefox-test KVM trial: `[VFORK-FSCANARY-PRE]` + `[VFORK-CANARY-PRE]` + matching POST lines emit; previously-fatal `wrap_addr=0x8ff8` now reports `wrap_val=? wrap_phys=?`; session runs past sc=4500 with no vfork_diag bugcheck
- [x] firefox-test TCG trial (×2): same — both PRE/POST canary pairs emit cleanly
- [ ] qa-engineer's axis-O 3-trial soak: re-run on this branch to confirm the diagnostic now arms and attributes the axis-O writer

## Refs

- Intel SDM Vol. 3A §4.6 (paging, virt→phys translation, present bit)
- System V AMD64 ABI §3.4.5.2 (SSP / stack-protector frame layout)
- POSIX vfork(2)

### DIFF
--- a/kernel/src/subsys/linux/vfork_diag.rs
+++ b/kernel/src/subsys/linux/vfork_diag.rs
@@ -196,16 +196,25 @@ pub fn snapshot_canaries(label: &str, parent_pid: u64, parent_tid: u64) {
 
     // ── Channel 1: parent saved-canary [rbp-8] ────────────────────────────────
     //
-    // `get_user_rsp_rbp` returns the (user_rsp, user_rbp) pair saved on this
-    // CPU's syscall frame at entry.  user_rbp at this point is the libc
-    // syscall wrapper's frame pointer; the wrapper's prologue did `push rbp;
-    // mov rbp, rsp`, so `*(user_rbp) = saved_RBP_of_caller`.  The caller is
-    // typically the libxul function that issued the spawn — and ITS `[rbp-8]`
-    // slot (= `*(user_rbp) - 8`) is the SSP slot the epilogue will compare
-    // against `fs:0x28`.  We dump BOTH the wrapper-frame `[rbp-8]` and the
-    // caller-frame `[rbp-8]` so the post-processor can pick whichever frame
-    // the SSP-instrumented function actually owns.
-    let (user_rsp, user_rbp) = crate::syscall::get_user_rsp_rbp();
+    // Resolve (user_rsp, user_rbp) from the parent thread's saved syscall
+    // frame at the top of its kernel stack — NOT from the per-CPU
+    // `frame_rsp` slot, which is overwritten by every subsequent syscall
+    // on that CPU (so by the time the POST snapshot runs after
+    // `schedule()` the slot may belong to an unrelated sibling syscall).
+    // Per `syscall_entry`'s save layout, the user_rsp slot is at
+    // `kstack_top - 8` and user_rbp is at `kstack_top - 32`; both are
+    // stable across pre-block and post-wake because the scheduler does
+    // not modify the saved syscall frame.
+    //
+    // user_rbp at this point is the libc syscall wrapper's frame
+    // pointer; the wrapper's prologue did `push rbp; mov rbp, rsp`, so
+    // `*(user_rbp) = saved_RBP_of_caller`.  The caller is typically the
+    // libxul function that issued the spawn — and ITS `[rbp-8]` slot
+    // (= `*(user_rbp) - 8`) is the SSP slot the epilogue will compare
+    // against `fs:0x28`.  We dump BOTH the wrapper-frame `[rbp-8]` and
+    // the caller-frame `[rbp-8]` so the post-processor can pick whichever
+    // frame the SSP-instrumented function actually owns.
+    let (user_rsp, user_rbp) = get_parent_user_rsp_rbp(parent_tid);
     if user_rbp == 0 {
         crate::serial_println!(
             "[VFORK-CANARY-{}] pid={} tid={} state=no_user_frame rsp={:#x}",
@@ -219,7 +228,7 @@ pub fn snapshot_canaries(label: &str, parent_pid: u64, parent_tid: u64) {
     let (wrap_val_str, wrap_phys_str) = read_userland_qword(wrap_canary_addr);
 
     // Walk one frame up: saved_caller_rbp = *(user_rbp).
-    let caller_rbp_opt = read_userland_qword_raw(user_rbp);
+    let caller_rbp_opt = read_userland_qword_raw(user_rbp).map(|(v, _phys)| v);
     let (caller_rbp_str, caller_canary_addr_str,
          caller_val_str, caller_phys_str) = match caller_rbp_opt {
         Some(0) | None => (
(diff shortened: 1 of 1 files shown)


me1iissa:
## APPROVE — security-engineer

Light pass (25 net LOC, single file, diagnostic-only path gated behind `vfork-canary-diag`). Verified against the codebase invariants and Intel SDM Vol 3A §4.6 paging semantics. Threat-model: zero new production attack surface; feature-flag-off master is byte-identical.

### Verified

1. **PHYS_OFF constant matches** — `0xFFFF_8000_0000_0000` in both `kernel/src/mm/vmm.rs:36` and the new `read_userland_qword_raw` body. Same constant used by the already-reviewed `fletcher32_user_page` (PR #331), so the kernel-direct-map deref pattern is consistent.

2. **`virt_to_phys_in` enforces page presence at every level** — `kernel/src/mm/vmm.rs:734-762` checks `PAGE_PRESENT` at PML4/PDPT/PD/PT. Returns `None` on any not-present entry. So a corrupt RBP chain that resolves to an unmapped user page now safely returns `("?","?")` instead of taking a #PF inside a STAC-bracketed kernel-mode load with no `__ex_table` recovery (which was the root cause of the deterministic `KERNEL_PAGE_FAULT CR2=0x8ff8` crash). CWE-476 NULL-pointer-deref class for the diagnostic helper itself is closed.

3. **Cross-page rejection arithmetic** is correct. An 8-byte read at `addr` straddles a 4 KiB boundary iff `(addr & 0xFFF) + 8 > 0x1000`, i.e. `(addr & 0xFFF) > 0xFF8`. The check `(addr & 0xFFF) > 0x1000 - 8` is exactly that. Boundary case `addr & 0xFFF == 0xFF8` (last 8 bytes of a single page) correctly falls through; `addr & 0xFFF == 0xFF9` correctly rejects. Per Intel SDM Vol 3A §4.6 a single virt→phys walk only covers one 4 KiB frame; the explicit rejection prevents silently stitching two phys reads from two different frames (which would mis-report the value if the second frame is a different VMA / aliased).

4. **No SMAP bracket required** in the new path — confirmed. The actual qword load is `read_volatile((PHYS_OFF + phys) as *const u64)`, a kernel-VA deref through PML4[256-511]. SMAP only fires on supervisor accesses to user-VA pages; a kernel-direct-map load reads a kernel-half VA regardless of which PTE backs the frame, so `EFLAGS.AC=0` is the correct state. No STAC/CLAC remained in the new code path — verified by inspection.

5. **`get_parent_user_rsp_rbp` frame-slot offsets match the canonical `syscall_entry` save layout** — `kernel/src/syscall/mod.rs:756`: frame is `[rdi, rsi, rdx, r8, r9, r10, r15, r14, r13, r12, rbx, rbp, r11, rcx, user_rsp]`, 15 × 8 = 120 bytes, top slot (slot 14 = user_rsp) at `kstack_top - 8`, slot 11 (rbp) at `kstack_top - 32`. Identical to the offsets used by `crate::proc::mod.rs:2274-2275` for fork frame-recovery, so this is canonical and stable across the schedule()-yield in vfork (the scheduler does not modify the saved syscall frame). CWE-825 use-after-free-via-stale-per-CPU concern from the old `get_user_rsp_rbp()` is closed: the new helper reads from the *parent's* kstack, not a per-CPU slot that any sibling syscall on the same CPU could have overwritten while the vfork parent was blocked.

6. **Return-type ripple `Option<u64>` → `Option<(u64, u64)>`** propagated correctly: 3 call sites in `snapshot_canaries:231` and `snapshot_stack_canary_walk:452, 463, 466` all updated to `.map(|(v, _phys)| v)`. No remaining unhandled call sites.

7. **`debug_assert!(va < USER_VA_LIMIT)` on `fletcher32_user_page`** — appropriate defensive guard per PR #331 follow-up; release builds elide the check, debug builds catch a future caller misuse where a kernel-half VA would silently produce a meaningless CRC. Cites `USER_VA_LIMIT = 0x0000_8000_0000_0000` from `kernel/src/syscall/mod.rs:99`.

8. **Build verification**: `cargo check -p astryx-kernel --target x86_64-unknown-none -Z build-std=core,alloc` clean both with and without `--features vfork-canary-diag` (only unrelated pre-existing warnings).

### Notes for follow-up (NOT blocking)

- The new helper is **TLB-bypassing**: `virt_to_phys_in` walks the page tables directly via the kernel direct map, so a stale TLB on the current CPU pointing `addr` at a different phys would not affect what `read_userland_qword_raw` reads. For a W215 aliasing diagnostic this is actually the *correct* behaviour (gets ground truth from the PML4), but worth noting if you later try to use this helper to compare what user mode would see vs what the page tables say.

- `get_parent_user_rsp_rbp` does not validate that `kstack_top` is in the kernel canonical higher half before the raw deref. In the current single-caller pattern (called from the parent thread's own syscall context with a valid TID found in THREAD_TABLE) this is safe — `kernel_stack_base` is always a higher-half address at thread-create time. If the helper grows additional callers later, consider a `kstack_top >= 0xFFFF_8000_0000_0000` guard before the two raw reads to match the defensive shape of `read_fork_user_regs` in `kernel/src/syscall/mod.rs:771-779`.

### Threat model

- **Attacker model**: none. Diagnostic-only path behind `vfork-canary-diag`. Master is byte-identical without the feature (verified by build).
- **No new privilege boundary crossings**: every read is either (a) the parent's own kernel-stack span (always present, kernel-VA) or (b) a kernel direct-map deref of a page-table-walked phys. No user-attacker-controlled value flows into a privileged decision; no writable surface added.
- **Information leak concern**: none. The values logged (`user_rsp`, `user_rbp`, qword contents from user-stack pages) are all owned by the parent process itself; the serial log is privileged.

References cited in the commit message and code are all public (Intel SDM Vol 3A §4.6, System V AMD64 ABI §3.4.5.2, POSIX vfork(2)) — no leakage of private corpus paths. ✓

LGTM, approving.

PR #335, opened 2026-05-19
scripts: Linux reference strace captures for ABI conformance (strace-ref + harness subcommand)

me1iissa:
## Summary

Adds a small toolchain (`scripts/strace-ref.py` + `qemu-harness.py strace-ref` subcommand) for capturing real-Linux strace traces from the same musl-linked firefox-esr binary AstryxOS ships, so abi-compatibility investigations have a ground-truth baseline.

- **Runs the AstryxOS-shipped binary** under the host's real Linux kernel inside bwrap (Alpine 3.20 rootfs reused from `~/.cache/astryxos-firefox-musl/rootfs/` — no extra disk, no VM).
- **Captures structured strace traces** with `-f -e trace=futex` (or any other syscall filter) and writes JSON sidecar metadata.
- **Diffs against AstryxOS serial logs** with a common op-class classifier; surfaces ABI gaps, volume plateaus, and AstryxOS-only diagnostic tags (`[FUTEX_WAKE_GHOST]`, `[FUTEX_TIMEDOUT]`, `[FUTEX_CLUSTER_WAKE]`) as JSON `notes`.

## Subcommands

```
python3 scripts/qemu-harness.py strace-ref setup [--bootstrap]
python3 scripts/qemu-harness.py strace-ref capture --label N --binary-args=ARGS --timeout SEC
python3 scripts/qemu-harness.py strace-ref diff --linux-trace P --astryx-log P
python3 scripts/qemu-harness.py strace-ref list
python3 scripts/qemu-harness.py strace-ref clean [--label STR]
```

All output is one-shot JSON to stdout. No REPLs, no interactive prompts, agent-friendly per AstryxOS invariant.

## End-to-end validation

Linux capture (about:blank, --headless, 20 s):
- 12,213 futex events / `{WAKE: 6484, WAIT: 5600, WAIT_BITSET: 127, REQUEUE: 2}`

Diff vs AstryxOS serial log (~1500 [FUTEX_*] lines):
- Linux total **12,213** vs AstryxOS **706** (5.8 % — plateau signature)
- REQUEUE: L=2 A=0 (**ABI coverage gap**)
- WAIT: L=5600 A=75 (1 %); WAIT_BITSET: L=127 A=242 (191 %); WAKE: L=6484 A=389 (6 %)
- AstryxOS-only diagnostic tags flagged: 115× FUTEX_TIMEDOUT, 11× FUTEX_WAKE_GHOST, 4× FUTEX_CLUSTER_WAKE

## Test plan

- [x] `scripts/strace-ref-smoke.py`: 11/11 PASS (setup, capture, list, diff, clean)
- [x] `python3 scripts/qemu-harness.py strace-ref setup` returns musl + version 115.24.0
- [x] Capture against real Alpine rootfs produces a 1.7 MB strace trace with timestamped tids
- [x] Diff against existing AstryxOS serial logs at `~/.astryx-harness/*.serial.log` reports plausible by-op delta + wedge-class notes
- [x] Existing `qemu-harness.py list` / `--help` unchanged
- [x] `python3 -c "import ast; ast.parse(...)"` clean on both scripts

### DIFF
--- a/scripts/STRACE_REF.md
+++ b/scripts/STRACE_REF.md
@@ -1,0 +1,203 @@
+# strace-ref — Linux reference captures for ABI conformance
+
+A small toolchain that runs the **exact musl-linked `firefox-esr` binary AstryxOS
+ships** under the host's real Linux kernel, captures `strace` output, and diffs
+it against AstryxOS serial logs.  It exists so abi-compatibility work has a
+ground-truth baseline: when AstryxOS shows a wedge or unexpected syscall
+pattern, you can answer "does Linux behave the same way against this binary?"
+in seconds.
+
+## Why not LXC / KVM?
+
+We use `bwrap` (bubblewrap) into an Alpine 3.20 rootfs that the AstryxOS build
+already produces under `~/.cache/astryxos-firefox-musl/rootfs/`.  Reasons:
+
+- **Identical binary**: same `firefox-esr-115.24.0esr` (musl), bit-for-bit, as
+  the one staged into `build/disk/usr/lib/firefox-esr/`.
+- **Real Linux kernel**: bwrap is a kernel-namespace sandbox, not a VM.  Every
+  syscall the binary makes goes to the host's real Linux kernel — exactly the
+  reference behaviour we want.
+- **Sub-second entry**: ~50 ms cold start vs ~5-10 s for a KVM/Alpine VM boot.
+- **No extra disk**: we reuse the rootfs the AstryxOS build already produced,
+  no second copy.
+
+LXC / a full Alpine VM remain available as a fallback if you need PID/network
+isolation we don't get from bwrap; the harness ships only the bwrap path for
+now because it's enough for futex/clone/signal/mmap tracing.
+
+## Subcommands
+
+All commands print a single JSON object to stdout and exit.
+
+### setup
+
+Verify the Alpine reference rootfs.  Reports the firefox-esr version, whether
+the launcher is musl-linked, and the captures directory.
+
+```
+python3 scripts/qemu-harness.py strace-ref setup
+# or:
+python3 scripts/strace-ref.py setup
+```
+
+If no rootfs is found, pass `--bootstrap` to invoke
+`scripts/install-firefox-musl.sh` (which downloads Alpine + apk-installs
+firefox-esr).
+
+### capture
+
+Run firefox-esr under `strace -f -e trace=futex` inside a bwrap sandbox.
+
+```
+python3 scripts/qemu-harness.py strace-ref capture \
+    --label ref-about-blank \
+    --binary-args="--headless --screenshot=/root/out.png about:blank" \
+    --timeout 30
+```
+
+Output trace lives at `~/.astryx-harness/strace-ref/captures/<label>.trace`.
+
+Useful flags:
+
+| flag | default | meaning |
+|------|---------|---------|
+| `--label` | `ref-YYYYMMDD-HHMMSS` | Capture label / filename stem |
+| `--binary-args` | `--version` | Args passed to firefox-esr (shlex-split) |
+| `--syscall-filter` | `futex` | strace `-e trace=` filter; comma-list ok |
+| `--timeout` | `60` | Wall-clock cap in seconds |
+| `--env KEY=VAL` | — | Extra env var (repeatable) |
+| `--output` | (under captures dir) | Override trace output path |
+| `--no-follow-forks` | (follow) | Drop strace `-f` |
+| `--no-timestamps` | (with timestamps) | Drop strace `-ttt` |
+
+The stats block on stdout summarises the trace:
+
+```json
+{
+  "stats": {
+    "lines": 21355,
+    "by_op": {"WAKE": 5935, "WAIT": 5359, "WAIT_BITSET": 129, "REQUEUE": 2},
+    "tids": [879932, 879933, ...],
+    "n_tids": 64,
+    "size_bytes": 1937412
+  }
+}
+```
+
+### diff
+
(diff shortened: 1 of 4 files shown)


me1iissa:
## Verdict: APPROVE with non-blocking nits

Host-side tooling, no kernel ABI touched. Reviewed under host-sandbox / attack-surface lens, not kernel lens.

Smoke test (11/11) is the author's claim — not exercised by the kernel test runner, which is fine for host infra. The bwrap pivot from LXC is a reasonable engineering call (shared host kernel = exactly the "real Linux" reference we want, ~50 ms enter, reuses the already-staged rootfs).

### Findings

**Clean (audit questions 3, 6, 7):**
- `--binary-args` uses `shlex.split()` with `subprocess.Popen(argv, shell=False)`. No shell injection. (Q3)
- All subcommands are one-shot argv → JSON stdout → exit. No REPLs, no interactive prompts, no persistent stdin. (Q6)
- No `SupportingResources/` references, no Linux-source attribution, no internal-corpus paths. Public refs cite man7.org, github.com/containers/bubblewrap, alpinelinux.org, musl.libc.org. (Q7)
- `re.sub(r"[^\w\.\-]+", "_", ...)` sanitises `--label` for the captures path — `../` collapses to `.._..`. (Q2)
- No `shell=True`, `os.system`, `eval`, `exec` anywhere in the diff.

### Non-blocking nits

**N1 — bwrap is missing namespace isolation flags (Q1).** `scripts/strace-ref.py:291-306` builds the bwrap argv with `--ro-bind`/`--proc`/`--dev`/`--tmpfs` mounts but no `--unshare-pid`, `--unshare-net`, `--unshare-ipc`, `--unshare-uts`. The PR description says "no virtualisation … shares the host kernel" — fine — but for a reference *baseline* this means:

- The traced firefox-esr sees the host's full PID space (can read other users' /proc).
- The traced firefox-esr can reach the network (a `--headless about:blank` doesn't, but a future `--binary-args` that does network I/O will, contaminating the trace).
- `--ro-bind /sys /sys` exposes host DMI/GPU/sensor state into Mozilla startup decisions, so the trace becomes host-machine-dependent (Quackie vs Arrythmia will diverge — see `[[project_test_hosts]]`).

For a *reference baseline that diffs against AstryxOS*, the closer this is to a minimal Alpine kernel namespace, the more signal the diff carries. Suggest adding `--unshare-pid --unshare-net --unshare-uts --unshare-ipc --new-session` to the default bwrap argv, with optional flags to re-enable. Non-blocking because the current behaviour still produces useful futex traces; just flag that the reference is host-dependent.

**N2 — host env leaks into sandbox (Q1).** No `--clearenv`. Host `LD_PRELOAD`, `LD_LIBRARY_PATH`, `MALLOC_CHECK_`, `XAUTHORITY`, `DBUS_SESSION_BUS_ADDRESS`, etc. all flow into the traced binary. For ABI reference fidelity this is the bigger soundness issue than N1: a developer with a stale `LD_PRELOAD=/usr/lib/libtcmalloc.so` on the host will capture a trace that doesn't match what AstryxOS sees. Recommend `--clearenv` then explicitly `--setenv` only `HOME`/`PATH`/`MOZ_HEADLESS` (already done) plus any caller-supplied `--env KEY=VALUE`.

**N3 — `setup` mutates the shared cache (Q4, clear deviation from dispatch).** `scripts/strace-ref.py:149-157` `prepare_rootfs_writable_dirs()` calls `(rootfs / d).mkdir(parents=True, exist_ok=True)` for `home`/`root`/`run`/`sys` directly inside the resolved rootfs path. When the resolved path is `~/.cache/astryxos-firefox-musl/rootfs/` (the priority-3 fallback in `find_rootfs`), this mutates the cache that the AstryxOS build pipeline also reads from / stages into `build/disk/`. The inline comment claims "harmless" but:

- The dispatch's Q4 explicitly asked: "should be idempotent and read-only. Verify it doesn't modify the rootfs."
- If `install-firefox-musl.sh` ever assumes a pristine cache (or runs an integrity check), the extra empty dirs are observable drift.
- It happens unconditionally during `capture` too (line 260), so even a one-off `capture` mutates the shared cache.

Easy fix: keep the writable mount points in `~/.astryx-harness/strace-ref/rootfs/overlay/` (or use `--overlay-src` / `--tmp-overlay` to layer them on top of the read-only `--ro-bind`), so the shared cache stays untouched. Non-blocking because the mutation is bounded and idempotent, but worth fixing before this script is used in CI where cache hygiene matters more.

**N4 — `--output` is taken verbatim (Q2, minor).** `scripts/strace-ref.py:270` `out_trace = Path(args.output) if args.output else (...)` — caller-controlled path with no normalisation/jail. The script runs as the calling user, so this isn't a privesc, but a typo'd `--output ../../etc/foo.trace` will quietly write to the parent dir. Suggest resolving against `REF_CAPTURES` and rejecting paths that escape via `Path.resolve().is_relative_to(REF_CAPTURES.resolve())`. Genuinely minor.

**N5 — `--astryx-log` follows symlinks (Q5, lowest impact).** `Path(args.astryx_log).open("r")` follows symlinks. Same risk as `cat` — bounded by the calling user's existing read access, not a privilege-boundary issue. Worth mentioning only because the dispatch listed it as an audit question; ignore if you don't care.

### Verdict rationale

Approve. The PR delivers the dispatched ABI-conformance tooling, the shell-injection / interactive / private-corpus audit gates all pass, and the immediate Mozilla futex diff already produced real signal (REQUEUE absent in AstryxOS, the volume plateau, the ghost/timeout tag callouts). N1+N2+N3 are reference-fidelity issues more than sandbox-escape risks — happy to land this and follow up with a hardening PR (`--clearenv` + `--unshare-*` + overlay for the writable dirs) once it's been used in anger for a few investigations.

PR #448, opened 2026-05-24
scripts/create-data-disk: ASTRYXOS_NAMESERVER env override for resolv.conf

me1iissa:
## Summary

- Default `/etc/resolv.conf nameserver` remains `10.0.2.3` (QEMU SLIRP DNS).
- New `ASTRYXOS_NAMESERVER=<ip>` staging-time env override for workloads that need a specific upstream resolver (e.g. internal-only zones the host can't proxy).
- One-line emit `[DATA-DISK] /etc/resolv.conf nameserver=...` confirms which resolver was staged.
- No specific IP baked into source — users supply override from their own environment.

## Why

PR #446 just landed the SLIRP UDP DNS kernel-side plumbing. With UDP DNS sound, the remaining variable for any workload that needs to resolve hostnames outside the host's default resolver is `/etc/resolv.conf`. An env-var override gives callers control without leaking site-specific IPs into the repo.

## Test plan

- [x] `bash scripts/create-data-disk.sh --force` → `nameserver=10.0.2.3` (default unchanged)
- [x] `ASTRYXOS_NAMESERVER=8.8.8.8 bash scripts/create-data-disk.sh --force` → override honoured
- [x] `[DATA-DISK] /etc/resolv.conf nameserver=...` emitted on each stage

### DIFF
--- a/scripts/create-data-disk.sh
+++ b/scripts/create-data-disk.sh
@@ -660,7 +660,16 @@ if command -v mcopy &>/dev/null; then
     printf 'astryx\n' | mcopy -o -i "${DATA_IMG}" - "::etc/hostname"
     printf '127.0.0.1 localhost\n::1 localhost\n10.0.2.2 gateway\n' | \
         mcopy -o -i "${DATA_IMG}" - "::etc/hosts"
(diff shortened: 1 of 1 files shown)


me1iissa:
Closing — folding into a broader scrub PR that also redacts existing internal infra refs from master.

PR #511, opened 2026-06-03
ff/real-website: drive headless Firefox at a real internet URL (NDE-8/NDE-9)

me1iissa:
## Goal
Wire the kernel-side configuration + ABI gaps that block headless Firefox from
fetching and screenshotting a **real internet website** on AstryxOS. A prior
recon+drive run proved the network path works end-to-end (live TCP Established
to 1.1.1.1:80/443, DNS resolved, TLS+HTTP driven) but committed nothing; this
branch commits the four fixes. The local file:// demo page stays on master —
this is the website-render configuration branch.

## The four fixes
1. **Remove the in-process connection block** (`gui/terminal.rs`): drop
   `MOZ_DISABLE_NONLOCAL_CONNECTIONS` from the spawn env. gecko treats it as a
   boolean gate enabled for any value ≠ `"0"`; when enabled it refuses every
   non-loopback connection *in-process*, before any SYN — the dispositive
   `furthest_stage=none` cause. The connect is rejected by libxul's necko
   layer, not by the kernel.
2. **Point Firefox at a real URL** (`main.rs`): the headless `--screenshot`
   driver now targets `https://1.1.1.1/` (IPv4-literal → skips DNS, valid cert,
   cleanest first end-to-end proof). `https://example.com/` (exercises DNS) is
   the follow-up once the literal path renders.
3. **Force IPv4 / kill the AF_INET6 fake-success** (cherry-picked NDE-8 toggle,
   `net/ipver.rs` + `syscall.rs`): `socket(AF_INET6)` → `EAFNOSUPPORT`,
   `connect()` to an AF_INET6 dest → `ENETUNREACH` when IPv6 is off (default).
   A truthful error for the family with no egress path lets musl Happy-Eyeballs
   (RFC 8305) fall through to the working IPv4 path; the prior fake-success
   stub reported a dead IPv6 destination as "connected" so fallback never fired.
4. **Fix the DNS reply-source gap** (`net/socket.rs` + `syscall.rs`): the
   `recvmsg(2)` AF_INET arm never marshalled `msg_name`. musl's connection-mode
   resolver `__res_msend` validates the reply source byte-for-byte against the
   nameserver and **drops** a reply whose `msg_name` is zero (RFC 1035 §7.3) →
   `getaddrinfo(3)` `EAI_AGAIN`. New `socket_recv_status_from` carries the true
   wire source 4-tuple (RFC 768 UDP / RFC 793 TCP); the arm writes the
   `sockaddr_in` into `msg_name` with `recvmsg(2)` truncation semantics,
   mirroring the existing `recvfrom(2)` (syscall 45) path. Needed for hostname
   targets.

## Tests
`socket_recv_status_from` returns the sender 4-tuple for a loopback UDP
datagram and reports `WouldBlock` (never `Eof`) on an empty UDP queue. Pass
under `--features test-mode` (CI) and `firefox-test,kdb`.

## Specs
recvmsg(2), recvfrom(2), connect(2), socket(2), getaddrinfo(3), RFC 1035 §7.3,
RFC 768, RFC 793 §3.5, RFC 6724, RFC 8305, POSIX.1-2017, published Firefox
headless / environment-variable docs.


me1iissa:
## Producer-side stall investigation (commit 5fed2de) — refutes the "deterministic parent MOZ_CRASH" framing

Independent fresh boot on `ff/real-website` @ db0f2c9 (firefox-test,differential-trace,blk-trace,kdb + GDB stub) reached the navigation-IPDL gate and ran for 21+ minutes. Key live findings:

- **The parent MOZ_CRASH is FLAKY, not deterministic.** A GDB autopsy armed at the kernel SIGSEGV-delivery function (`deliver_sigsegv_from_isr`, breakpoint confirmed armed) for the full 1305s window recorded **hit_count=0** — no SIGSEGV ever fired. The run reached a soft livelock instead of crashing. (`movl $0x210,0x0` MOZ_RELEASE_ASSERT at libxul+0x4812b79 did not reproduce.)
- **Channel state reproduces byte-exact**: content channel frozen at sock3 (parent) recv_pushed=1596 popped=1596, sock4 (content) recv_pushed=1280 popped=1280, avail=0 both directions — both sides fully drained+consumed, neither produces the next message.
- **Kernel readiness machinery verified faithful** (live `poll-revents`): parent reports 94 fds correctly POLLIN-ready; self-pipe read-ends correctly revents=0 when empty; content channel correctly POLLOUT-when-writable. The AF_UNIX `poll_revents` arm (syscall/mod.rs:4353-4404) gates POLLIN on `has_data||has_pending||has_scm` and POLLOUT on `writable()` — **refutes prime-suspect S2** (poll is NOT unconditionally POLLOUT).
- **Localization**: content (pid3) main thread frozen in `poll()` (nr=7, sc frozen at 1881) awaiting the parent's navigation message; parent (pid1) alive (sc climbing 32k+, channel writable POLLOUT) but does not send. This is **S1 (parent never sends the navigation IPDL message)** — a gecko-internal event-loop decision, not a dropped kernel wake. Same gate the active `ff-navigation-ipdl-gate2` workflow is grinding.

**No kernel fix shipped**: the only un-refuted suspect (S1) cannot be dispositively localized to a kernel file:line without the gMozCrashReason readout (which never fired this run) or gecko-state introspection, and re-litigating the refuted-faithful readiness predicates would violate the saga-exhaustion rule (W101/verdict-C lineage).

**Tooling added** (commit 5fed2de, scripts-only, additive): a `mem_ptr_chain` autopsy capture-step + `moz-crash-reason` / `moz-crash-reason-from-isr` presets so the next agent that catches the flaky crash can read the gMozCrashReason string in one breakpoint hit — the exact next-probe the prior characterization said was needed.

me1iissa:
Obsolete umbrella branch: real-website headless driving landed piecemeal on master long ago (Firefox renders real sites; PNG extraction on master). Closing.

PR #552, opened 2026-06-11
Gate-1: revive the dead-LAPIC BSP (idle_tick per-CPU freshness) + wake blocked pipe writers on supervisor drains + MOZ_LOG-to-stderr probe

me1iissa:
## What this names (Gate-1 Finding B — the socket-process init wedge)

All live-proven on `firefox-test-core,kdb,ff-stderr-mirror` boots (sessions preserved at `/tmp/gate1-boot{1,2,8}*.serial.log`):

1. **CPU 0's LAPIC periodic timer dies permanently ~1.1–1.3 s into every KVM boot.** `TIMER_ISR_PER_CPU[0]` froze at 113/119/132 fires across runs while CPU 1 accumulated thousands. LVT/INIT rewrites do not revive it (19 rearm attempts, count frozen).
2. **`idle_tick()` blessed a terminal `hlt`**: it judged timer health by the *global* TICK_COUNT-vs-TSC delta, which the sibling CPU keeps healthy. The FFTEST supervisor loop hlt'ed on its first iteration and never woke (GDB: breakpoints on the loop top and `poll_output()` — **zero hits in 20 s**; both vCPUs parked at `hlt`). The machine has effectively been **single-core under KVM** — CPU 0's runqueue stranded every READY thread placed there (the long-observed "ready-but-never-runs" threads).
3. **The 4 KiB launcher pipe (stdout+stderr of every FF process) was never drained again.** Post-#551 writers block per POSIX.1-2017 write(2)/pipe(7): `pipe-diag` showed `buffered=4089–4095/4096, write_waiters=3, readers=1`, static; `thread-park-audit` showed pid 1 AND pid 3 main threads blocked in `writev(fd=2)` for 130+ s. **pid 3 froze at sc=1192 deterministically (5/5)** — the socket-process init stall behind the Gate-1 crash variants. The fast-vs-trace flakiness variable is named: stray device IRQs (disk/NIC) occasionally woke the hlt and ran one drain iteration — trace builds generate more of them, hence more survivors.
4. **`poll_output()` drained via raw `pipe_read()`** which never wakes parked writers (the read(2) arm does that separately) — so even a draining supervisor left writers parked forever.

## Fixes

- `arch/x86_64/irq.rs`: `idle_tick()` blesses `hlt` only if **this CPU's own timer ISR fired within the slack window** (per-CPU TSC timestamp recorded in the ISR; a fire-count delta wrongly blesses a terminal hlt after an early-boot LAPIC death). Silent timer ⇒ spin + rate-limited rearm.
- `main.rs` FFTEST loop spin arm: tick-bounded busy-wait (sibling-clock 100 Hz cadence) instead of MHz yield/schedule churn.
- `ipc/pipe.rs`: `pipe_read_wake()` (drain + `wake_writers_all` on nonzero reads) used by `poll_output` and all demo-supervisor drains.
- `kdb`: new `pipe-diag` op + harness plumbing.
- `gui/terminal.rs`: launcher env routes scoped `MOZ_LOG=sync,socketprocess:5,pipnss:5,nsHostResolver:5` to **stderr** (file buffer was empty on abort). [MOZ_LOG docs](https://firefox-source-docs.mozilla.org/xpcom/logging.html)
- `test_runner.rs`: blocking-write sub-test 6 — watchdog-proven parked writer completes after a kernel-context `pipe_read_wake` drain.

## Verified

Run 8: supervisor survives the LAPIC death, pipe drains, **child stderr/MOZ_LOG reaches serial** — the socket process narrates its own init (`CONSTRUCT SocketProcessChild` → nsHostResolver/TRR init → `Initialized TRRService`), then advances past the sc=1192 wedge into the REAL Gate-1 crash (CR2=0xd0, libxul file-off 0x2102190, after softokn/freebl load, before any pipnss error or ipcclientcerts open) — the genuine capture this mission wanted. The failing init step is now bracketed to the release-silent stretch of the socket process's NSS bring-up (`NSS_NoDB_Init` tail / TLS-version-range setup before the cipher-suite step, which would have logged).

## ⚠️ DO NOT MERGE YET — known exposure

With CPU 0 genuinely scheduling again, **concurrent `exit_group()` teardown of a multi-threaded process races across CPUs and bugchecks** (KERNEL_GPF with RIP on a kernel stack / KERNEL_PAGE_FAULT jump-to-NULL; observed 2/2 at pid-3 teardown, where tid 92's `exit_group(1)` races tid 127's SIGSEGV `exit_group(-11)`). Pre-existing — previously unreachable because CPU 0 slept through every boot. Needs a dedicated fix (aether-kernel-engineer scope) before this lands; this PR makes it deterministically reproducible.

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -485,6 +485,34 @@ pub fn send_ipi(dest_apic_id: u8, vector: u8) {
     while lapic_read(LAPIC_ICR_LO) & (1 << 12) != 0 {}
 }
 
+/// Send a fire-and-forget IPI: program the ICR and return WITHOUT spinning on
+/// the delivery-status bit.
+///
+/// Unlike [`send_ipi`], this never executes the
+/// `while lapic_read(ICR_LO) & DELIVERY_PENDING {}` busy-wait.  That wait is
+/// the documented deadlock hazard for cross-CPU pokes raised from inside an
+/// interrupt handler: two CPUs that each spin waiting for the other's LAPIC to
+/// accept can wedge (the same ISR-to-ISR LAPIC-contention failure that retired
+/// the earlier IPI-based debug-register sync — see
+/// `arch::x86_64::irq::timer_tick`).  A dropped or still-pending poke is
+/// self-healing: the caller (the timer ISR) re-evaluates and re-pokes on the
+/// next tick crossing, so at most one tick of wake latency is lost — never a
+/// permanent hang.
+///
+/// Per Intel SDM Vol. 3A §11.6.1, writing ICR_LOW issues the IPI; the
+/// delivery-status bit is advisory.  Back-to-back writes before the prior IPI
+/// is accepted are allowed (the LAPIC serialises them); the worst case is a
+/// coalesced wake, which for an edge-triggered fixed-vector timer poke is
+/// harmless.
+#[inline]
+pub fn send_ipi_noblock(dest_apic_id: u8, vector: u8) {
+    if !is_enabled() {
+        return;
+    }
+    lapic_write(LAPIC_ICR_HI, (dest_apic_id as u32) << 24);
+    lapic_write(LAPIC_ICR_LO, vector as u32 | ICR_ASSERT);
+}
+
 /// Bootstrap application processors (APs).
 ///
 /// Writes a 16-bit real-mode trampoline at physical 0x8000, then sends the
--- a/kernel/src/arch/x86_64/irq.rs
+++ b/kernel/src/arch/x86_64/irq.rs
@@ -282,6 +282,119 @@ static LAST_REARM_TSC: [AtomicU64; super::apic::MAX_CPUS] =
 /// tick. Used to rate-limit the software tick to ~`TICK_HZ`.
 static LAST_SOFT_TICK: AtomicU64 = AtomicU64::new(0);
 
+/// LAPIC periodic-timer interrupt vector (IRQ0).  Kept in sync with the LVT
+/// programming in `apic::init`/`apic::rearm_timer` (`0x20000 | 32`).  Used as
+/// the IPI vector when a healthy CPU pokes a sibling whose own timer has gone
+/// silent, so the sibling re-enters the SAME `timer_tick` path it would have
+/// run from its own LAPIC.
+const TIMER_VECTOR: u8 = 32;
+
+/// Staleness threshold (in ticks) past which a sibling CPU's LAPIC timer is
+/// judged silent and is poked with a wake IPI from the publishing CPU.  At
+/// TICK_HZ=100 this is ~30 ms — comfortably longer than the one-tick ISR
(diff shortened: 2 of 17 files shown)


me1iissa:
Pushed two commits onto this chain (f1c94f3 → e62ecf5):

**`23057e8` kernel/sched: per-CPU context-switch generation barrier for dead-kstack reclaim**
- Replaces the wall-clock `TICK_COUNT` quiescence gate on the dead-stack cache with a per-CPU context-switch generation barrier: a reclaimed kstack is re-issuable / zero-fillable / PMM-freeable only once every online CPU has passed a quiescent state (top of `schedule()`, each picker iteration, `check_reschedule()`) since reclaim. A time gate proves nothing about a CPU stalled mid-`switch_context_asm`; the generation gate waits on the causal event instead.
- Zero-fill moved from push (reclaim) to pop (re-issue) — zeroing at push *was* the corrupting write when the gate was beaten.
- The cache-overflow / non-standard-size PMM-free path now parks in a quarantine behind the same barrier (this was the residual writer that survived `MAX_DEAD_STACKS=0`).
- New regression test (245, registered early, before the suite's known pipe-test wedge point): barrier deterministically HOLDs under IF=0 and RELEASEs after the every-CPU grace; existing `exit_group` live-kstack SMP guard still passes; test 236 updated for the deferred zero (push must write nothing).

**`e62ecf5` kernel/syscall: clear ctx_rsp_valid before Blocked at futex/waitpid/vfork parks**
- Four park sites violated the parking protocol (release-store `ctx_rsp_valid=false` before `Blocked`), leaving the flag stale-true across the park→switch window. A sibling-CPU wake in that window (futex wakes are asynchronous per futex(2)) let the sibling's picker resume the thread from its STALE previous-switch `context.rsp` while the parking CPU was still saving the live frame onto the same kstack — two CPUs interleaved on one kernel stack.
- Live-captured signature on this branch: `KERNEL_PAGE_FAULT 0xdead0006`, CR2=RIP=0x86 (kernel IFETCH at a near-zero address popped off the victim's own kstack), the victim's `[GATE1/FUTEX-WAIT]` trace line torn mid-emit, immediately after a sibling exit-CLEARTID futex wake.

**Verification** (all on this branch, KVM, smp=2):
- 8 firefox-test boots: ZERO teardown bugchecks (previously deterministic at sc≈5–11k). Six boots cleared sc=91k+ (old cache-off onset); deepest soak reached sc=1.82M. Two boots ran FF to its known clean-exit terminal.
- Both CPUs verifiably scheduling throughout (`[SCHED/STARVE] … on cpu=0/cpu=1`, AP online, dual-core `exit_group` guard test passing).
- Kstack recycling alive post-gate (`[KSTACK/ALLOC] cache hit` after pushes; test 245 release half; quarantine drains to 0).
- Pre-existing, unrelated: the test suite wedges at the blocking-pipe sub-test 5 on this branch — reproduced identically on a clean f1c94f3 baseline build, so it predates these commits.

Not merging — this chain goes through the combined SMP review gate.

PR #554, opened 2026-06-11
shell: clear ctx_rsp_valid before marking a Running tid Dead in kill

me1iissa:
## Summary

The operator shell `kill <tid>` command was writing `ThreadState::Dead` unconditionally, without clearing `ctx_rsp_valid` first when the target thread was `Running` on a peer logical processor.

**Invariant violated:** The kstack reaper (`sched::reap_dead_threads_sched`) admits a `Dead` thread for stack-free when `ctx_rsp_valid.load(Acquire) == true`. When a thread switches out, `switch_context_asm` saves the kernel RSP and *then* sets `ctx_rsp_valid=true`. If a foreign CPU marks that thread `Dead` while it is still executing (state=`Running`, `ctx_rsp_valid` stale-true from its last switch-out), the reaper can immediately free a live kernel stack — UAF.

**Fix:** Mirror the pattern at the futex/waitpid/vfork park sites (`ke/wait.rs`, `proc/mod.rs`): if the target's current state is `Running`, `Release`-store `ctx_rsp_valid=false` BEFORE writing `Dead`. `switch_context_asm` on the peer CPU will re-set `true` only after committing the RSP, at which point the stack is genuinely safe to reap. Non-`Running` threads (Blocked, Ready, Sleeping) already had their RSP committed at their last switch-out; `ctx_rsp_valid` is correctly `true` for them — leave it unchanged.

**Scope:** One file, one call site (`shell::cmd_kill`). No impact on any non-shell path.

**Verification:**
- Build clean via `qemu-harness.py build` (23.85 s, zero errors).
- Boot under smp=2 KVM: clean boot, all phases complete, no BUGCHECK / FAULT / PANIC in serial log.
- The operator shell inputs to the framebuffer, not serial, so interactive `kill` exercise via harness send/grep is not practical; correctness rests on the code path analysis and the invariant matching the established park-site pattern.

## Test plan

- [ ] Review that `ctx_rsp_valid.store(false, Release)` precedes `state = Dead` when `state == Running` — matches the invariant at `sched::reap_dead_threads_sched:535`.
- [ ] Confirm diff touches only `kernel/src/shell/mod.rs` (1 file, 21 lines added, 0 deleted).
- [ ] CI green.

### DIFF
--- a/kernel/src/shell/mod.rs
+++ b/kernel/src/shell/mod.rs
@@ -1371,9 +1371,30 @@ fn cmd_kill(parts: &[&str]) {
                 orbit_println!("kill: cannot kill idle thread");
             } else {
                 // Mark the target thread as Dead.
+                //
+                // INVARIANT: if the target is currently Running on another
+                // logical processor, its saved kernel RSP has not been
+                // committed — ctx_rsp_valid is stale-true from the thread's
+                // last switch-OUT.  The kstack reaper
+                // (sched::reap_dead_threads_sched) admits a Dead thread for
+                // stack-free only when ctx_rsp_valid==true (i.e. the RSP has
+                // been saved and the CPU has left that stack).  Observing
+                // Dead+ctx_rsp_valid==true on a thread that is still
+                // executing on that stack is a UAF.  Release-store false
+                // BEFORE writing Dead so that switch_context_asm on the peer
+                // CPU re-sets true only AFTER saving the RSP, at which point
+                // the stack is safe to reap.  Non-Running threads already had
+                // their RSP saved at their last switch-out and their
+                // ctx_rsp_valid is correctly true — leave it unchanged.
                 let found = {
                     let mut threads = crate::proc::THREAD_TABLE.lock();
                     if let Some(t) = threads.iter_mut().find(|t| t.tid == tid) {
+                        if t.state == crate::proc::ThreadState::Running {
+                            t.ctx_rsp_valid.store(
+                                false,
+                                core::sync::atomic::Ordering::Release,
+                            );
+                        }
                         t.state = crate::proc::ThreadState::Dead;
                         t.exit_code = 137; // SIGKILL
                         true


me1iissa (commented):
## VERDICT: APPROVE

Independent review gate. Provenance verified: HEAD == `b15d5d5`, `merge-base --is-ancestor cc95771 HEAD` = BASE-OK, diff is +21 lines in `kernel/src/shell/mod.rs` only (6 code / 15 comment). One confirming smp=2 KVM boot built clean and reached the CoreSched init with **zero bugchecks / panics / page-faults** (the only `#GP` log hit is the benign UMIP feature-announcement line).

### Q1 — Pattern fidelity: PASS
The shell fix is a byte-for-byte protocol match of the canonical foreign-SIGKILL path at `proc/mod.rs:2074-2081`: both gate on `if t.state == ThreadState::Running`, both `ctx_rsp_valid.store(false, Release)` **before** the `state = Dead` store. The reaper read side (`sched/mod.rs:654`) pairs with `Ordering::Acquire`. This is the same protocol e62ecf5 (#552) applied at the futex/waitpid/vfork park sites.

**Is `if state == Running` too narrow?** No — it is correct and complete. The authoritative invariant at `proc/mod.rs:2031-2036` states that Blocked/Ready/Sleeping siblings had their RSP saved long ago by `switch_context_asm`, so their `ctx_rsp_valid` is *legitimately* true; forcing it false would strand them Dead-but-unreapable (kstack leak). I verified the "#552 made Blocked-with-unsaved-RSP impossible" claim directly: **every** park-to-Blocked transition (`ipc/waitlist.rs:78-79`, `subsys/linux/syscall.rs:3505-3506` & `4762-4763`, `ke/wait.rs:91-92/192-193/256-257`, `syscall/mod.rs:521-522/2895-2896`) and the lone Sleeping park (`proc/mod.rs:2414-2415`) clears `ctx_rsp_valid=false` immediately before the state store. Running is the only state that carries a stale-true flag.

### Q2 — Lock context + ordering: PASS
Both stores execute under `crate::proc::THREAD_TABLE.lock()` (`proc/mod.rs:480`, single `Mutex<Vec<Thread>>`), the *same* mutex the reaper holds while reading (`sched/mod.rs:644`). Reader and writer are mutually excluded, so the fact that `t.state = Dead` is a plain (non-atomic) enum write is immaterial — `is_reapable()` (reads `state`) and `ctx_rsp_valid.load()` both occur inside the reaper's lock hold, never interleaving with the writer's `{ctx_rsp_valid=false; state=Dead}` block. The Release/Acquire ordering additionally covers the two lock-free readers: the picker (`sched/mod.rs:1599`) and the `is_tid_current_on_any_cpu` backstop. x86-TSO is sufficient given those atomics.

### Q3 — Completeness within the kill path: PASS
`shell/mod.rs:1398` is the **only** `ThreadState::Dead` write in `cmd_kill` (line 1329 is a display match-arm, not a store). `kill` is the only kill command in the dispatch table (`shell/mod.rs:57,699`) — no `pkill`/`killall` sibling.

### Q4 — Mid-switch window: PASS
schedule() flips the outgoing thread `Running→Ready` **and** clears `ctx_rsp_valid=false` under the picker's THREAD_TABLE lock (`sched/mod.rs:1726-1727`) *before* releasing the lock and entering `switch_context_asm`. So when cmd_kill takes the lock and reads `Running`, the peer CPU provably has not yet entered that flip → the flag is genuinely stale-true → clearing it is correct. If the peer already flipped to Ready, cmd_kill reads `Ready`, skips the clear, but `ctx_rsp_valid` is already false from schedule(). The asm re-sets the flag true only *after* `mov [rdi], rsp` (`proc/thread.rs:81→90`), and x86-TSO makes the RSP store globally visible before the flag store. The #552 CTX_SWITCH_GEN quiescence gate plus the independent `!is_tid_current_on_any_cpu(t.tid)` reaper guard (`sched/mod.rs:673`) close the residual — no new hazard introduced.

### Q5 — Build sanity: PASS
Clean release build, smp=2 KVM, scheduler up, zero bugchecks. Interactive shell-kill isn't reachable over serial; Q1-Q4 carried by code analysis.

No issues found. The change is a faithful, minimally-scoped extension of the merged park-protocol to the shell kill path.

PR #555, opened 2026-06-11
fix(mm/pfh): gen-abort over-release SIGSEGV'd live VMAs on concurrent SMP

me1iissa:
## Summary

The page-fault handler's six demand-paging install arms (COW-COPY, CACHE-PRIVATE, CACHE-ALIAS, READAHEAD, SINGLE-PAGE, ANON in `arch/x86_64/idt.rs`) sample the address-space-wide `VmSpace::generation` counter and re-check it immediately before installing a leaf PTE. Each arm returned `false` on **any** change to that counter. This is the hard blocker for the Firefox real-website demo: 2/2 boots SIGSEGV'd the main Firefox process (PID 1) during startup, before any page navigation.

## Answer to the key open question — why does the retry SIGSEGV?

It is **not** a retry. A `false` return from `handle_page_fault` delivers SIGSEGV to the faulting thread **immediately** — there is no silent re-fault in this kernel:

- `idt.rs:622` — `if handle_page_fault(cr2, error_code, frame) { … return; }`; on `false` the handler falls through.
- `idt.rs:738` — `deliver_sigsegv_from_isr(...)` is called directly on the `false` path.

So the gen-abort's documented rationale ("abort … so the user re-faults against the new VMA") was factually wrong: the `free_page(...); return false` in the abort arms **is** the SIGSEGV. This is option **(d)** from the dispatch: the abort path explicitly signals failure → SIGSEGV delivery. In the READAHEAD/ANON arms the aborted page is the **faulting page itself** (readahead window starts at `page_addr`, `pages_to_map[0]`), freed without ever installing a PTE, then SIGSEGV'd on the live VMA — a deterministic use-after-fault, not a recoverable re-fault.

## Root cause

`VmSpace::generation` is an **address-space-wide** counter bumped for *every* VMA-list mutation anywhere in the address space — `insert_vma` (mmap), `remove_range` (munmap / MAP_FIXED replace / MADV_DONTNEED), `adjust_brk`, and the `clone_for_fork` CoW write-protect pass — including mutations on virtual addresses **unrelated** to the faulting one. A sibling CPU running any mmap/brk during the faulting thread's install window aborted (→ SIGSEGV) a fault whose page was a perfectly valid, unchanged mapping.

The bug was dormant while one CPU was effectively single-threaded through each install loop. Once both CPUs became genuinely concurrent (`[SCHED/STARVE] force-select` firing on both CPUs in the failing boots) it fired deterministically during Firefox startup — `[PROC] PID 1 exit_group(11) caller_tid=2` at sc~150-205k.

## Serial evidence (reproduced before the fix, unmodified base kernel)

```
[PF/gen-abort] ANON #0 addr=0x7ef53f03a000 gen_at_start=9465 gen_now=9466 — releasing frame
!!! Page Fault (error_code=0x6)  CR2: 0x00007ef53f03a000  RIP: 0x00007f6565f1640c  not-present WRITE USER
[SIGNAL] SIGSEGV ISR delivery: PID=1 CR2=0x7ef53f03a000 ...
[SIGNAL/VMA] pid=1 name=[mmap] base=0x7ef53f02b000 end=0x7ef53f12c000 prot=rw- anon=1   ← faulting VMA STILL LIVE
[D13/CR2-PROV] pid=1 tid=2 cr2=0x7ef53f03a000 data_phys=UNMAPPED
```
RIP bytes `f3 48 a5` = musl `rep movsq` into the still-mapped anon VMA. The gen-abort and the SIGSEGV are the **same** fault, tid=2, same address — exactly one gen-abort → exactly one SIGSEGV → PID 1 dies.

## The fix

The actual hazard the gen-check was meant to guard — installing a PTE over a frame a concurrent `unmap_and_free_range_in` is freeing — is already closed by the atomic install primitives each arm uses: `map_page_in_if_absent` and `map_page_in_cow_if_unchanged` hold `mm_sem` + `VMM_LOCK` and write the leaf PTE only if it is still absent / still maps the expected old frame (Intel SDM Vol. 3A §4.10.4.3). The loser of any install/unmap race on **this** VA backs out and frees only its own fresh frame — the established CAS-install loser-backout contract.

So on a generation change we re-validate the **specific faulting VA** instead of aborting unconditionally:
- **Faulting VMA still covers the address with the same backing** (anonymous / same file `(mount,inode)` / still `PROT_WRITE` for CoW) → unrelated mutation → proceed with the atomic install (READAHEAD resyncs its generation baseline and continues).
- **Faulting VA's own VMA removed or re-backed** → abandon the fault — a legitimate SIGSEGV per POSIX (access to an unmapped address). In READAHEAD, a replaced VMA after iteration 0 means the faulting page is already installed → break to `return true` rather than SIGSEGV.

No lock added; `PROCESS_TABLE` re-acquired briefly for re-validation only where it was already dropped (top of the lock order; MOUNTS/cache/PMM not held). The CAS loser-backout contract is preserved on every arm. The generation property (don't install over a concurrently-mutated faulting VA) is retained — narrowed from "any mutation" to "this VA's mutation".

## Verification — 3/3 firefox-test boots at smp=2 KVM

| sid | gen-abort | PID-1 SIGSEGV | exit_group(11) | content-procs | bbc.com TLS ClientHello | teardown bugcheck |
|---|---|---|---|---|---|---|
| `a1d0cb5922d9` | 0 | 0 | 0 | ✓ | 6 (incl. BBC 212.58.235.129 + Fastly) | 0 |
| `724f51e147f5` | 0 | 0 | 0 | ✓ | 4 (Fastly 151.101.x) | 0 |
| `bb16ebbf19b7` | 0 | 0 | 0 | ✓ | 6 (incl. BBC 212.58.236.129 + Fastly) | 0 |

All 3 boots ran past the sc150-205k window where the unmodified kernel deterministically crashed (now at sc 152k-180k, still running), reached the `[GATE] content-procs` gate, and issued real bbc.com HTTPS TLS ClientHellos — navigation begins (the gold marker; the unfixed kernel issued ZERO TLS ClientHello). Zero teardown bugchecks confirms no #552 regression. The kernel builds clean under `test-mode`; adds `test_pfh_gen_abort_revalidate_faulting_va` (Test 98j) modelling the re-validation predicates.

> The in-kernel `test-mode` suite currently hangs on a pre-existing `pipe blocking write` sub-test (`test_pipe_blocking_write_semantics`, registered far earlier than the mm tests; introduced by the recent pipe-blocking lineage `a33910a`) under genuinely-concurrent SMP — orthogonal to this change, which touches only PFH install arms.

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -1425,22 +1425,60 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, _frame: &mut Interrupt
                 // Re-check generation right before install — Acquire pairs
                 // with the Release fetch_add in `bump_generation_for_cr3`
                 // and `VmSpace::*` mutators (Intel SDM Vol. 3A §8.2.3).
+                //
+                // `generation` is address-space-WIDE, so a bump here can come
+                // from an unrelated VA's mutation.  This arm's install is an
+                // atomic compare-and-swap: `map_page_in_cow_if_unchanged` (below)
+                // writes the leaf PTE only if it STILL maps `old_phys` under
+                // `VMM_LOCK`, and the loser frees `new_phys` cleanly and returns
+                // true.  That CAS is the true protection against a concurrent
+                // unmap/CoW of THIS VA — if a sibling cleared or replaced the
+                // PTE, the CAS fails and we back out without aliasing.  A
+                // `return false` on a bare generation bump would instead deliver
+                // SIGSEGV to a thread doing a legitimate CoW write of a present,
+                // valid page (this kernel has no silent re-fault), so we no
+                // longer abort on a bump alone.  We only abandon the fault if the
+                // faulting VA's own VMA is gone or no longer writable — the case
+                // where the user genuinely lost write access (POSIX SIGSEGV).
                 let gen_now = vm_space.generation.load(core::sync::atomic::Ordering::Acquire);
                 if gen_now != gen_at_start {
+                    let still_writable_here = vm_space
+                        .find_vma(faulting_addr)
+                        .map(|v| v.prot & crate::mm::vma::PROT_WRITE != 0)
+                        .unwrap_or(false);
+                    if !still_writable_here {
+                        #[cfg(feature = "firefox-test-core")]
+                        {
+                            static CNT: core::sync::atomic::AtomicU64 =
+                                core::sync::atomic::AtomicU64::new(0);
+                            let n = CNT.fetch_add(1, core::sync::atomic::Ordering::Relaxed);
+                            if n < 5 || n % 500 == 0 {
+                                crate::serial_println!(
+                                    "[PF/gen-abort] COW-COPY #{} addr={:#x} \
+                                     gen_at_start={} gen_now={} — faulting VMA \
+                                     gone/RO, dropping new frame",
+                                    n, page_addr, gen_at_start, gen_now);
+                            }
+                        }
+                        crate::mm::pmm::free_page(new_phys);
+                        return false;
+                    }
+                    // Unrelated mutation — the faulting VA is still a writable
+                    // mapping.  Fall through to the atomic CAS install, which
+                    // re-validates `old_phys` under the page-table lock.
                     #[cfg(feature = "firefox-test-core")]
                     {
-                        static CNT: core::sync::atomic::AtomicU64 =
+                        static REVAL: core::sync::atomic::AtomicU64 =
                             core::sync::atomic::AtomicU64::new(0);
-                        let n = CNT.fetch_add(1, core::sync::atomic::Ordering::Relaxed);
-                        if n < 5 || n % 500 == 0 {
+                        let n = REVAL.fetch_add(1, core::sync::atomic::Ordering::Relaxed);
+                        if n < 5 || n % 5000 == 0 {
                             crate::serial_println!(
-                                "[PF/gen-abort] COW-COPY #{} addr={:#x} \
-                                 gen_at_start={} gen_now={} — dropping new frame",
+                                "[PF/gen-reval] COW-COPY #{} addr={:#x} \
+                                 gen_at_start={} gen_now={} — VMA still writable, \
+                                 proceeding with CAS install",
                                 n, page_addr, gen_at_start, gen_now);
                         }
                     }
-                    crate::mm::pmm::free_page(new_phys);
-                    return false;
                 }
                 // W215 measurement (Phase-1 witness, behaviourally inert):
                 // re-read the leaf PTE immediately before the install.  If a
@@ -1754,16 +1792,43 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, _frame: &mut Interrupt
                     return false;
                 }
 
-                // Closure: per-arm generation re-check.  Returns true if the
-                // caller should abort (and the caller must release `cached_phys`
-                // appropriately before returning false).  The CoW-copy sub-arm
-                // additionally frees its freshly-allocated `private_phys` via
-                // its own check immediately before installing.
-                let gen_unchanged = || -> bool {
+                // Closure: per-arm "safe to install" re-check.  Returns true if
+                // the caller may proceed with the install, false if it must
+                // abandon the fault.
+                //
+                // `generation` is an address-space-WIDE counter: a bump can be
+                // caused by a mutation on a virtual address unrelated to the one
+                // we are faulting on.  Aborting (→ `return false` → SIGSEGV, with
+                // no silent re-fault in this kernel) on a bare bump killed
+                // processes whose faulting page was a live, unchanged cache-hit
+                // mapping — a fatal use-after-fault that any concurrent sibling
+                // mmap could trigger.  The cache-hit install below is atomic
+                // (`map_page_in_if_absent`, mm_sem + VMM_LOCK, present-recheck),
+                // so the install/unmap race on THIS VA is already closed.  We
+                // therefore treat a bump as "safe to proceed" UNLESS the faulting
+                // VA's own VMA has been removed or re-backed away from this
+                // (mount,inode) — the only case where the cached frame would be
+                // stale for the VA (a brief PROCESS_TABLE re-acquire; that lock
+                // is top of the order here, MOUNTS/cache/PMM are not held).
+                let gen_safe_to_install = || -> bool {
                     match ch_vm_generation.as_ref() {
                         Some(g) => {
                             let now = g.load(core::sync::atomic::Ordering::Acquire);
-                            now == ch_gen_at_revalidate
+                            if now == ch_gen_at_revalidate {
+                                return true;
+                            }
+                            // Generation moved — re-validate the SPECIFIC
+                            // faulting VA against the same file backing.
+                            let procs = crate::proc::PROCESS_TABLE.lock();
+                            procs.iter()
+                                .find(|p| p.pid == target_pid)
+                                .and_then(|p| p.vm_space.as_ref())
+                                .and_then(|vs| vs.find_vma(faulting_addr))
+                                .map(|v| matches!(&v.backing,
+                                    crate::mm::vma::VmBacking::File {
+                                        mount_idx: m, inode: ino, ..
+                                    } if *m == mount_idx && *ino == inode))
+                                .unwrap_or(false)
                         }
                         None => true,
                     }
@@ -1799,9 +1864,12 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, _frame: &mut Interrupt
                                 crate::mm::pmm::PAGE_SIZE,
                             );
                         }
-                        // W216 H_5j-B (unified concurrency): re-check generation
-                        // immediately before install — see arm-level closure.
-                        if !gen_unchanged() {
+                        // W216 H_5j-B (unified concurrency): re-validate the
+                        // faulting VA before install — see arm-level closure.
+                        // Only abandon if the faulting VA's own VMA was
+                        // removed/re-backed (an unrelated address-space mutation
+                        // is no longer treated as a reason to SIGSEGV).
+                        if !gen_safe_to_install() {
                             #[cfg(feature = "firefox-test-core")]
                             {
                                 static CNT: core::sync::atomic::AtomicU64 =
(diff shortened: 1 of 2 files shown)


me1iissa:
Applied the review gate's fix-and-resubmit findings in `d0dd579` (FILE-arm gen-abort re-validation hardening). No new lock; lock order unchanged. Net diff is a *reduction* (325 ins / 389 del) — the three sprawling per-arm generation closures collapse into one shared predicate + a held-`PROCESS_TABLE` install.

**PFH-REVAL-2 (MAJOR — wrong-offset content window).** The CACHE per-install re-check and SINGLE-PAGE bump re-check validated only `(mount, inode)`, discarding offset+span. All three FILE arms now go through one full-tuple predicate `pfh_file_vma_match` (`idt.rs:1252`) checking `(mount, inode, offset)` backing AND `base..base+length` span — the same tuple the READAHEAD post-I/O revalidate already used. A same-inode-different-offset `MAP_FIXED` remap / `remove_range` hole-punch now correctly returns `None` instead of installing OLD-offset bytes at a NEW-offset VA.

**PFH-REVAL-1 (MAJOR — install into a just-munmapped range).** The FILE arms now hold `PROCESS_TABLE` across BOTH the re-validation AND the install (`_ch_pt_guard` `idt.rs:1858`, `_ra_pt_guard` `idt.rs:2368`, `_sp_pt_guard` `idt.rs:2826`), mirroring the ANON/COW arms which never drop it. `sys_munmap` Phase 2a removes the VMA under `PROCESS_TABLE` before its lock-free Phase 2b PTE clear, so a concurrent munmap now serialises behind the guard (removes before us → abandon, or after us → its Phase 2b clears our PTE). The held section subsumes the per-install generation re-check, so the per-arm generation closures are removed. The readahead guard is dropped before the single-page fallback's VFS read (`idt.rs:2709`) so `PROCESS_TABLE` is never nested under MOUNTS.
- **Lock-order verification:** order is `PROCESS_TABLE (top) → page-cache / PMM / refcount / mm_sem.read() / VMM_LOCK`. `cache.rs`, `pmm.rs`, `refcount.rs`, `vmm.rs`, `vma.rs` *never* acquire `PROCESS_TABLE` (grep-confirmed) → no inverse nesting. The COW arm (`map_page_in_cow_if_unchanged`) and ANON arm (`map_page_in_if_absent`) already install under a continuously-held `PROCESS_TABLE`, so this is the identical, already-exercised nesting (Intel SDM Vol. 3A §4.10.4.3).

**PFH-REVAL-3 (MINOR — stale over-permissive flags).** `pfh_file_vma_match` re-derives `page_flags` and the `MAP_SHARED` bit from the LIVE re-validated VMA inside the held critical section (`idt.rs:1252`), so a racing `mprotect(PROT_READ)` in the bump window narrows the installed PTE instead of installing from stale `PROT_WRITE` (POSIX mprotect(2)).

**PR555-4 (MINOR — test 98j was near-tautological).** The full-tuple predicate is factored into the single `pfh_file_vma_match` the install arms call; Test 98j (`test_runner.rs:31356`) now drives that exact function (`:31437`, `:31481`, `:31499`) and gains a same-inode-different-offset remap case (`:31486`) asserting the predicate returns `None`. **Non-tautology proven:** temporarily weakening `pfh_file_vma_match` to `(mount, inode)`-only made Test 98j FAIL at sub-case (2) (`[FAIL] gen-reval — FULL-TUPLE FILE predicate returned Some for a same-inode-DIFFERENT-offset remap`); restoring the full tuple → PASS. A full two-CPU concurrency test is noted as future work in the test.

**Out of scope (untouched):** PFH-REVAL-4 (truncate/ftruncate SIGBUS gap) and the flaky GATE-A startup NULL-deref (CR2=0x0).

**Verification:**
- Clean build: default / `firefox-test-core` / `test-mode` all `rc=0`.
- Staged-kernel provenance: the staged `kernel.bin` contains only `[PF/gen-abort] ANON` + `COW-COPY` strings (base `2ad0dae` had all six incl. CACHE-PRIVATE/CACHE-ALIAS/READAHEAD/SINGLE-PAGE) — confirms the boots ran my tree.
- Test 98j: PASS with full tuple; FAIL with weakened predicate (non-tautology).
- 2× firefox-test boots @ smp=2 KVM (bbc.com): soaked past tick 109k–115k, **0 `[PF/gen-abort]`**, **0 teardown bugchecks**, content processes spawned, `www.bbc.com` resolved to its real BBC IPs (212.58.x). (A 3rd boot died early to the orthogonal pre-existing GATE-A CR2=0x0 NULL-deref — no preceding `[PF/gen-abort]` — and was replaced.)

Not merging — leaving for the gate.

me1iissa (commented):
VERDICT: APPROVE

Focused re-review of the fix-and-resubmit delta `2ad0dae..d0dd579` (fast-forward, net −64 lines; idt.rs + test_runner.rs). All four prior in-scope findings are resolved by the delta and no new hazards are introduced. Reviewed against the authoritative git objects (`git show d0dd579:...`), not a worktree checkout. One live smp=2 KVM firefox-test boot on `d0dd579` corroborates the static analysis.

## Per-finding resolution (all verified)

**PFH-REVAL-2 (full tuple) — RESOLVED.** All three FILE arms route revalidation through the single shared predicate `pfh_file_vma_match` (idt.rs:1252), which checks the full `(mount, inode, offset)` backing (`*m==mount_idx && *ino==inode && *o==file_base_offset`, idt.rs:1264) AND the span `v.base==vma_base && v.base+v.length==vma_end` (idt.rs:1265). A grep for backing comparisons in the install path leaves exactly one match — the full-tuple predicate at idt.rs:1262; no `(mount,inode)`-only comparison survives. Returns `None` for same-inode-different-offset remap and for span changes (hole-punch), per POSIX mmap(2)/mprotect(2).

**PFH-REVAL-1 (hold-across-install) — RESOLVED; make-or-break sequencing confirmed per arm.**
- CACHE (idt.rs:1858): `lookup_and_acquire` (idt.rs:1803) precedes the lock; the arm performs no VFS read (page already resident); install (`map_page_in_if_absent`, idt.rs:1935/2055) runs under the held `_ch_pt_guard`.
- READAHEAD (idt.rs:2368): every `fs.read` (idt.rs:2206/2289) completes inside the page-build loop that closes at idt.rs:2320, before `_ra_pt_guard` is acquired at idt.rs:2368; the install loop (idt.rs:2419-2682) holds the guard with no VFS/MOUNTS call. `drop(_ra_pt_guard)` at idt.rs:2709 occurs BEFORE the single-page fallback's VFS read — PROCESS_TABLE never nests over MOUNTS.
- SINGLE-PAGE: `drop(_ra_pt_guard)` (idt.rs:2709) → `MOUNTS.try_lock()`+clone+drop (idt.rs:2746) → `fs.read` lock-free (idt.rs:2785) → `_sp_pt_guard` acquired AFTER the I/O (idt.rs:2826) → revalidate+install under the held guard.

No arm does lock→revalidate→drop→blocking-read→install. The safe shape (read-before-lock, then lock→revalidate→install) holds on every arm.

**No blocking under PROCESS_TABLE / no lock-order inversion — CONFIRMED.** No `fs.read`/MOUNTS acquisition occurs between any guard acquisition and its install. The leaf modules called under the guards (cache.rs/pmm.rs/refcount.rs/vmm.rs/vma.rs/tlb.rs) contain zero `PROCESS_TABLE.lock`; the two textual `PROCESS_TABLE` references in vmm.rs:841 and vma.rs:243 are doc comments documenting the order, not acquisitions. Lock order PROCESS_TABLE (top) → cache/PMM/mm_sem.read()/VMM_LOCK is preserved, matching the ANON/COW arms.

**PFH-REVAL-3 (fresh flags) — RESOLVED.** `pfh_file_vma_match` re-derives `to_page_flags()` and `MAP_SHARED` from the live re-validated VMA (idt.rs:1268-1271). Every install consumes these fresh values. The only place the pre-fault snapshot is read is the READAHEAD `else` branch (idt.rs:2392-2395) for the `n_pages==0` case, where the values feed only the zero-iteration install loop and are dead on that path (the single-page fallback re-binds fresh flags at idt.rs:2827).

**PR555-4 (Test 98j) — RESOLVED.** test_runner.rs:31356 (registered at :1179) imports and drives `pfh_file_vma_match` directly — the same function the install arms call — including a non-tautology same-inode-different-offset case that first proves a `(mount,inode)`-only check would match, then asserts the full-tuple predicate returns `None`. No orphaned expectations from the generation-closure removal.

**Generation-closure removal soundness — CONFIRMED.** The VMA-list mutators are `&mut self` (vma.rs insert_vma:1036, remove_range:1063), reachable only via a PROCESS_TABLE-held mutable borrow. munmap Phase 2a (`remove_range`) and Phase 3 (`insert_vma`) run under `PROCESS_TABLE.lock()` (syscall/mod.rs ~3741/3780); Phase 2b (`unmap_and_free_range_in`) only clears PTEs, not the VMA list. mprotect mutates flags/splits under PROCESS_TABLE (subsys/linux/syscall.rs:7961). Nothing mutates the VMA list under mm_sem alone, so the continuous PROCESS_TABLE hold genuinely subsumes the per-install generation re-check.

Note (non-blocking): the `vm_space.generation` counter is still bumped by mutators (e.g. mprotect) but is now read by nobody in the install arms — vestigial but harmless (a relaxed atomic increment). Removing the field is out of scope.

## Live verification (d0dd579, smp=2 KVM, firefox-test-core,kdb)

Staged-kernel provenance: the d0dd579 idt.rs contains only the ANON (idt.rs:3142/3160) and COW-COPY (idt.rs:1541/1560) `[PF/gen-abort]`/`[PF/gen-reval]` strings — the 6 FILE-arm emitters present in 2ad0dae are removed, as intended.

Across ~8900 serial lines and 13 FF process generations under the libxul SMP thread-storm (which hammers the FILE demand-paging arms continuously):
- PFH FILE-arm events (gen-abort + revalidate-abandon + gen-reval): **0**
- ke_bugcheck / BUGCHECK / teardown bugcheck: **0**
- refcount underflow / double-free / wrong-content / aliasing-guard (H3a) / CRC-mismatch: **0**
- Max syscall depth reached: 259,071; a generation loaded deep into libxul/ICU/pango/cairo init.

The over-release-of-live-VMA-frame SIGSEGV class this PR targets does not occur. The only faults observed are the orthogonal pre-existing GATE-A startup NULL-deref (CR2=0x0, rbp=0x0, user_rip at libxul `...6f8`, with NO preceding `[PF/gen-abort]` line — the documented startup flake), which is unrelated to the PFH delta. The `[TLB/TIMEOUT]` lines are a separate known SMP shootdown issue in tlb.rs, also untouched by this delta. (Note: this build's compiled default URL is https://1.1.1.1/, not bbc.com, so no bbc DNS marker is expected here; FF generations did reach the network layer — PROC-METRICS show net writes — before the orthogonal GATE-A deaths cut them short.)

Citations: POSIX mmap(2)/mprotect(2); Intel SDM Vol. 3A §4.10.4.3 / §4.10.5.

PR #556, opened 2026-06-11
mm/mmap: re-pick base on lost placement race (GATE-A startup SIGSEGV)

me1iissa:
## Root cause — GATE-A flaky Firefox startup NULL-deref

`sys_mmap` picks the base address (Phase 1: `find_free_range` / `find_free_stack_range` under PROCESS_TABLE) and inserts the VMA (Phase 3: `insert_vma` under a **separate** PROCESS_TABLE acquisition). A sibling thread's mmap/brk/stack allocation can claim an overlapping range in the window; `insert_vma`'s only failure mode is Overlap, and the call returned **ENOMEM**.

Per POSIX mmap(2), a NULL-addr non-FIXED request leaves placement entirely to the implementation — losing an internal placement race is not an observable error. Firefox's shared-memory paths `MOZ_RELEASE_ASSERT` on mapping success; the loser of two concurrent kernel-placed mmaps aborted via the deliberate `mov [0], <line>` crash store (the observed CR2=0x0, rbp=0 SIGSEGV).

### Evidence chain (boots on 4c9234e, smp=2 KVM, `--trace`)
- 3/3 boots died at GATE-A; **every** death preceded 1:1 by `[MMAP-ERR] pid=N insert_vma failed ... flags=0x1` for the dying pid (observed pids 1/3/5, 5 instances across 3 boots).
- Syscall ring shows the exact losing call, e.g. `mmap(0, 0x87c, RW, MAP_SHARED, fd=32) = -12` — a sub-page shared mapping "out of memory" at startup.
- FF stderr names it: `WARNING: Call to mmap failed: Out of memory: ... shared_memory_posix.cc:515`.
- All three death RIPs decode to deliberate Gecko aborts (`MOZ_RELEASE_ASSERT(result.isOk())` ×2 sites, `MOZ_CRASH(IPC FatalError in the parent process!)`).
- `--trace` widens the window (serial emit sits between the phases): 3/3 vs ~50% plain — consistent with a pure timing race.
- Dual-core exposure: pre-#552 the machine was de-facto single-core; the window needs true concurrency.

### Fix
Phase 3 re-validates the chosen base under the lock it already holds; if the range was claimed, it re-picks via the same allocator arm (stack window or general walk) atomically with the insert. ENOMEM remains only for genuine address-space exhaustion. MAP_FIXED keeps its address contract (never re-picked). MAP_FIXED_NOREPLACE collisions in the window now report EEXIST per mmap(2).

### Test
Test 299 arms a deterministic test-mode hook that injects a 1-page squatter VMA at the Phase-1-chosen base (exactly what the winning sibling would install) and asserts success at a non-overlapping address + an overlap-free VMA list. Passes; full suite run shows no new failures (remaining: known env-gap fixtures + pre-existing test518b / monotonic-rate flakes in untouched subsystems).

### Verification (live)
- Base (4c9234e): 3/3 boots → GATE-A death (PID=1 SIGSEGV CR2=0x0).
- Fixed: see verification results in PR comments (4 boots, expect 0 deaths + `[MMAP-REPICK]` lines proving the race fired and was healed).

### Residual divergences noted (not in scope)
- MAP_FIXED squatter in the Phase-2a→3 window still fails the insert (mmap(2) would atomically replace) — needs the 2a/2b teardown re-run loop; never observed live.
- `[FFTEST/mmap-so]` trace line prints the pre-re-pick base if a `.so` map loses the race (symbolication off for that one lib until the `[MMAP-REPICK]` line is consumed by the harness symboliser).
- `[TLB/TIMEOUT]` warnings are a separate liveness symptom (quarantine path engages by design, `mm/tlb.rs:585`) — present in 2 of 3 dying boots but absent in the third; ruled out as GATE-A's cause.

### DIFF
--- a/kernel/src/syscall/mod.rs
+++ b/kernel/src/syscall/mod.rs
@@ -3424,6 +3424,53 @@ pub(crate) fn sys_dsp_ioctl(request: u64, arg_ptr: *mut u8) -> i64 {
 ///
 /// Supports both anonymous (MAP_ANONYMOUS) and file-backed mappings.
 /// Actual physical pages are allocated on demand via the page-fault handler.
+/// Test-mode state for the mmap placement-race re-pick test (see
+/// `mmap_test_inject_placement_squatter` and `test_runner.rs`).
+#[cfg(feature = "test-mode")]
+pub mod mmap_race_test {
+    use core::sync::atomic::AtomicU64;
+    /// Pid armed to deterministically lose the Phase-1→Phase-3 placement
+    /// race on its next kernel-placed mmap (0 = disarmed).
+    pub static ARMED_PID: AtomicU64 = AtomicU64::new(0);
+    /// Base address where the squatter VMA was injected (for the test to
+    /// read back and clean up; 0 = hook has not fired).
+    pub static SQUATTER_BASE: AtomicU64 = AtomicU64::new(0);
+}
+
+/// When armed for `pid`, insert a 1-page anonymous VMA at `base` — the
+/// address Phase 1 just chose — mimicking a sibling thread's concurrent
+/// kernel-placed mmap winning the Phase-1→Phase-3 window.  One-shot.
+#[cfg(feature = "test-mode")]
+fn mmap_test_inject_placement_squatter(pid: u64, base: u64, is_fixed: bool) {
+    use core::sync::atomic::Ordering;
+    use crate::mm::vma::*;
+    if is_fixed {
+        return;
+    }
+    if mmap_race_test::ARMED_PID
+        .compare_exchange(pid, 0, Ordering::AcqRel, Ordering::Acquire)
+        .is_err()
+    {
+        return;
+    }
+    let mut procs = crate::proc::PROCESS_TABLE.lock();
+    if let Some(p) = procs.iter_mut().find(|p| p.pid == pid) {
+        if let Some(space) = p.vm_space.as_mut() {
+            let squatter = VmArea {
+                base,
+                length: 0x1000,
+                prot: PROT_READ | PROT_WRITE,
+                flags: MAP_PRIVATE | MAP_ANONYMOUS,
+                backing: VmBacking::Anonymous,
+                name: "[test-squatter]",
+            };
+            if space.insert_vma(squatter).is_ok() {
+                mmap_race_test::SQUATTER_BASE.store(base, Ordering::Release);
+            }
+        }
+    }
+}
+
 pub(crate) fn sys_mmap(addr_hint: u64, length: u64, prot: u32, flags: u32, fd: u64, offset: u64) -> i64 {
     use crate::mm::vma::*;
 
@@ -3631,6 +3678,15 @@ pub(crate) fn sys_mmap(addr_hint: u64, length: u64, prot: u32, flags: u32, fd: u
     #[cfg(not(any(feature = "firefox-test-core", feature = "test-mode-trace")))]
     let (cr3, backing, name, base) = mmap_setup;
 
+    // ── Test-mode hook: deterministically lose the Phase-1→Phase-3 placement
+    // race.  When armed for this pid, a 1-page squatter VMA is inserted at the
+    // just-chosen base — exactly what a sibling thread's concurrent mmap
+    // Phase 3 would do in the window between our two PROCESS_TABLE
+    // acquisitions — so the Phase-3 re-pick path is exercised without relying
+    // on real cross-CPU timing.  Compiled out of production kernels.
+    #[cfg(feature = "test-mode")]
+    mmap_test_inject_placement_squatter(pid, base, is_fixed);
+
     // W215 H3a diagnostic: count MAP_SHARED+PROT_WRITE file-backed mappings.
     //
     // The check is after argument decode and backing resolution (above), but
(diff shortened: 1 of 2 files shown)


me1iissa:
## Live verification results (fixed kernel 6d454fa, smp=2 KVM, `--trace` = maximum race sensitivity)

**Base reproduction (4c9234e): 3/3 boots died at GATE-A** — PID=1 SIGSEGV CR2=0x0, each preceded 1:1 by `[MMAP-ERR] insert_vma failed` (sids 6f732d89b4eb, e29fc8c66699, 16c13c86beab; deaths at ticks ~2956 / ~4621 / ~45000; child pids 3 and 5 also died of the same class in e29f).

**Fixed: 0/4 deaths**, boots ran 20+ minutes wall (≥2.5× past the latest observed base death at tick ~45000):

| sid | SIGSEGV (any pid) | [MMAP-ERR] | [MMAP-REPICK] (race fired, healed) | gen-abort | bugcheck |
|---|---|---|---|---|---|
| d38fba5ffda3 | 0 | 0 | 0 | 0 | 0 |
| 509fc0cbedc2 | 0 | 0 | **2** | 0 | 0 |
| 9fe78094da8e | 0 | 0 | **2** | 0 | 0 |
| c081de8b80a9 | 0 | 0 | **1** | 0 | 0 |

The 5 healed re-picks are exactly the pre-fix victim classes: pid=1 kernel-placed maps of 0x1000 (the SharedStringMap-assert killers) and 0x82000/0x84000 (the late "IPC FatalError" killer in the base 16c1 boot). The race demonstrably still fires on the fixed kernel and is healed every time.

**Suite**: full `ci-run --features test-mode` — Test 299 (new) PASS; remaining failures are the known env-gap fixtures (Musl hello / TCC / busybox / sigchld / ascension / glibc_hello: missing /disk/bin binaries locally) plus pre-existing flakes test518b (thread reaper) and monotonic-rate (timer), both in subsystems untouched by this diff and both failing before Test 299 ran.

**[TLB/TIMEOUT] relationship**: ruled out as GATE-A's cause — one base boot died with 0 timeouts, and the fixed boots log 18–64 timeouts each with 0 deaths. Separate liveness follow-up.

me1iissa:
## CI attribution — both red checks are pre-existing on the base lineage

The `workflow_dispatch` run on this branch (27384997130) is the **first CI ever executed on the ff/real-website lineage** (every recent green run is master-lineage). Its two failures were re-tested against base `4c9234e` directly:

1. **kernel smoke / test518b** (`THREAD_TABLE still has a target tid after reaper pass`) — **fails identically on base 4c9234e** (local `--features test-mode,kdb` run, same message, `elapsed_ticks=11`). It is the #552/#554 deferred-free/reaper SMP-timing family; it also runs *before* Test 299 in suite order and the new test hook is disarmed unless Test 299 arms it, so this diff cannot influence it. Pre-existing.
2. **tooling smoke / serial-web** (`GATE_TO_PHASE covers every milestone`) — **fails identically on base 4c9234e** (`python3 scripts/serial-web-smoke.py` at the detached base). The render branch added MILESTONE markers that the serial-web phase map never learned. Pre-existing; toolchain follow-up.

Also observed while attempting the base suite runs: 2 of 3 base suite boots wedged at the blocking-pipe test (pipe_writer_a/b parked at sc=1) — the already-documented pre-existing dual-core pipe-writer wedge under separate investigation.

Suggested follow-ups (separate PRs): add `test518b` to `ci/allow-fail.json` with a tracking issue (or fix the reaper-pass convergence bound), and teach `scripts/serial-web` the new MILESTONE names.

me1iissa (commented):
VERDICT: APPROVE

Independent review gate (not the author). Reviewed the authoritative git objects at 6d454fa (provenance: rev-parse OK; `merge-base --is-ancestor 4c9234e 6d454fa` = BASE-OK). Diff +200/-1 across kernel/src/syscall/mod.rs and kernel/src/test_runner.rs. One smp=2 KVM firefox-test boot at 6d454fa (staged-kernel `[MMAP-REPICK]` string confirmed present in the booted binary).

## Per-question findings (file:line)

**1. Re-pick correctness — CLEAN.**
(a) **Termination guaranteed.** The re-pick (mod.rs:3865-3889) is a SINGLE allocator call, NOT a loop. `find_free_range` is bounded (`for _ in 0..1000`, vma.rs:1229, returns None on exhaustion); `find_free_stack_range` is bounded (16 retries → falls through to find_free_range, vma.rs:1191-1216). Adversarial concurrent allocation cannot affect it: no other thread can mutate `space.areas` while PROCESS_TABLE is held, so the re-picked base is free at the instant it is chosen. No spin; genuine exhaustion → proper ENOMEM (mod.rs:3886).
(b) **Original semantics honored.** `is_stack_alloc` is computed once (mod.rs:3502) and reused — the re-pick selects find_free_stack_range for stacks and find_free_range otherwise (mod.rs:3866-3870), the EXACT arm Phase 1 used. The stack-arm internal fall-through to find_free_range (vma.rs:1216) is PRE-EXISTING and identical to Phase 1. Non-NULL non-FIXED hint: addr_hint is used ONLY in the is_fixed branch (mod.rs:3570); the kernel never honored such hints in Phase 1 either, so the re-pick introduces no divergence. Guard-page/alignment conventions live entirely inside the unchanged allocator arms.
(c) **Insert is genuinely atomic with the pick.** Re-pick and insert_vma (mod.rs:3891) run in the SAME Phase-3 critical section under one held PROCESS_TABLE lock — no second drop. insert_vma's only failure is Overlap (vma.rs:1036-1042), which is impossible for a non-fixed request after a successful re-pick.

**2. Lock-hold hygiene — CLEAN.** The re-pick acquires NO locks: find_free_range/find_free_stack_range are pure in-memory Vec scans; aslr_page_offset → rand_u64 → rand_u64_nondeterministic takes no Mutex/RwLock (RDRAND/PRNG only, security/rand.rs). No VFS/page-cache/PMM-contention work moved under the lock. No lock-order inversion. The added work (one bounded Vec scan + at most one allocator walk) is the SAME class of work Phase 1 already ran under the lock — just one more instance.

**3. MAP_FIXED semantics — CORRECT.** MAP_FIXED untouched: re-pick is `!is_fixed`-guarded (mod.rs:3865); Phase 2a teardown is `is_fixed && !is_noreplace` (mod.rs:3795); still replaces per mmap(2). MAP_FIXED_NOREPLACE → EEXIST (mod.rs:3917-3921) is correct (ENOMEM=12, EEXIST=17 per asm-generic/errno-base.h). The residual MAP_FIXED 2a→3 squatter window is genuinely pre-existing (the Phase-1/Phase-3 lock split predates this PR) and documented (mod.rs:3858-3862), not widened — the PR adds no new FIXED-path code that could enlarge it.

**4. Phase-2 side effects keyed to stale base — HARMLESS.** For non-fixed mappings Phase 2a/2b is skipped entirely (no PTE pre-work, no remove_range, no page-cache reservation). `elf_load_delta` is 0 for non-fixed (mod.rs:3653) — nothing embedded is invalidated by re-pick. VmBacking::File carries no base. The ONLY thing keyed to the stale base is the cosmetic `[FFTEST/mmap-so]` print (mod.rs:3736), which fires before re-pick → at worst a host-symboliser inaccuracy in trace builds for a .so reservation that loses the race. Nothing FUNCTIONAL keyed to the stale base.

**5. Test 299 honesty — GOOD.** The squatter hook deterministically injects a 1-page VMA at the Phase-1 base (mod.rs:3443-3475), modeling the winning sibling. NON-TAUTOLOGY: on the unfixed kernel the squatter forces insert_vma → Overlap → ENOMEM, and the test asserts `ret >= 0` AND non-overlapping → it FAILS on unfixed Phase-3 logic. The hook + module + call site are ALL `#[cfg(feature = "test-mode")]` (mod.rs:3429/3443/3687); `firefox-test-core` (CI smoke) does NOT pull `test-mode` → ZERO production reachability. One-shot via compare_exchange on ARMED_PID; hook returns early for is_fixed. current_pid()/current_pid_lockless() resolve to the same running-thread pid.

**6. errno fidelity — SATISFIED.** Non-fixed path can no longer return spurious ENOMEM unless address space is genuinely exhausted: after a successful re-pick, insert_vma cannot fail (only failure mode is Overlap, ruled out under the held lock). EEXIST is returned ONLY for is_noreplace (mod.rs:3917). `[MMAP-ERR]` now fires only on genuine re-pick exhaustion or the fixed-window insert failure, never on a healed race.

## Live verify (one smp=2 KVM firefox-test boot @ 6d454fa)
Full 23,006-line serial log, max sc=105,395, both CPUs active:
- MMAP-ERR (spurious ENOMEM): **0**
- CR2=0x0 GATE-A deaths: **0**
- bugcheck / teardown: **0**; panic: **0**; MOZ_CRASH/RELEASE_ASSERT: **0**; SIGSEGV: **0**
- Six distinct processes (pid 1,3,4,5,6,7) ALL loaded libxul.so and survived the mmap-heavy library-load phase — the exact phase the placement race killed ~50% of the time pre-fix.
- [MMAP-REPICK] = 0 in this single boot (the race is probabilistic; the author witnessed 5 heals over 20+ min). The deterministic Test 299 covers the re-pick path; this boot confirms ABSENCE of the failure mode (the primary criterion).
- None of the known pre-existing failures (test518b, GATE_TO_PHASE, pipe-writer, CR2=0x10 mallocng) appeared.

No blocking findings. The fix is minimal, correct, atomic with the insert, lock-clean, errno-faithful, FIXED-contract-preserving, and backed by a genuine (non-tautological, test-mode-gated) regression test. APPROVE.

PR #558, opened 2026-06-12
fix(vfs): /dev/urandom must never repeat blocks across same-tick reads (BBC loadURI gate root cause)

me1iissa:
## The gate

QA retest at `8bffee9` (3 boots, smp=2 KVM, `--ff-url https://bbc.com/news`): FF reaches `[GATE] screenshot-actors` with a full e10s tree, but **never navigates** — pcap shows zero TLS ClientHello, zero `:443` SYNs (DNS resolves, plain `:80` HTTP works), pid 3 (the socket process) dead as an unreaped zombie at `exit_group(1)`, pid 1 livelocked. 5/5 deterministic across two feature builds at head.

## Root cause (one bug, every symptom)

`vfs::fd_read`'s `/dev/urandom` arm computed every byte as a **pure function of the 10 ms scheduler tick + byte index**. Two back-to-back 32-byte reads inside one tick are byte-identical.

PKCS#11 soft tokens run the FIPS 140-2 §4.9.2 **continuous RNG self-test** at `C_Initialize`: consecutive fixed-size system-RNG blocks must differ, else the module fails with `CKR_DEVICE_ERROR`. Captured live with the parent's stderr mirrored (`MOZ_LOG=pipnss:5`):

```
failed to initialize NSS with codes -8023 -8023      (= SEC_ERROR_PKCS11_DEVICE_ERROR)
attempting no-module db init
last-resort NSS_NoDB_Init
failed to initialize NSS
nsNSSComponent::InitializeNSS() failed               (loops forever)
```

Consequences:
- **Parent**: all four NSS init fallbacks fail → no TLS socket provider → every `https://` channel dies before `connect(2)` → zero ClientHello / zero `:443`, while `http://` traffic (openh264, detectportal) flows normally. Navigation to `https://bbc.com/news` can never start.
- **Socket process (pid 3)**: child init hits the same failing step, returns false → orderly `exit(1)` in its init window — the long-standing "pid-3 silent early death" (Gate-1) signature. Syscall ring shows the canonical fingerprint: softokn3/sqlite3/freeblpriv3 dlopen → `open/read(32)` `/dev/urandom` twice **at the same tick t** → no further syscalls → teardown. The golden Linux strace of the same binary shows 5×32 B + 1×1024 B urandom reads at this point and proceeds.

**Why it looked like a regression**: survival is a pure timing-phase artifact — do the two 32-byte reads straddle a tick boundary? Slow/trace builds pass the window, fast profiles die, and *any* kernel change shifts the phase. That retroactively explains the ~10-25 %/boot win rate on 2026-06-10, the trace-build "positive control", and the deterministic 0/3 after #552-556. No PR in the merge chain is at fault; a bisect would have falsely named #552.

## Fix

Route the `/dev/urandom` / `/dev/random` arm through `security::rand::rand_u64` — RDRAND-backed with a TSC-seeded fallback, and the same chokepoint ASLR / AT_RANDOM / `getrandom(2)` already use (so `astryx.rng_seed=` record-replay determinism still covers the path). Per random(4), both nodes may serve the same CSPRNG. 24 lines + comment.

**Test 520** mirrors the consumer access pattern (open → read(32) → close, twice back-to-back = same tick at boot-test speed; plus two reads on one fd) and asserts blocks differ and are non-zero. Passes on the fixed kernel.

## Verification (fixed kernel, smp=2 KVM, https://bbc.com/news)

First fixed boot (sid c26a4bedb58f):
- pid 3 socket process **survives** (zero `exit_group`, first time in 6 observed boots)
- TLS ClientHello to 151.101.0.81:443 (bbc.com Fastly edge), full handshake + data flow
- `[GATE] png-write path=/tmp/out.png bytes=972247` — **1366×7420 RGBA PNG of bbc.com/news**, extracted and band-analyzed: 13 sparse text bands / 3 blank — same shape as the 2026-06-10 calibration win (remaining blanks = the separate, pre-existing image-CDN Gate-2)
- Closed crash classes: gen-abort 0, MMAP-ERR 0, bugcheck/DOUBLE_FAULT 0, CR2=0x10 0. One `exit_group(11)` fired **after** png-write during FF shutdown — pre-existing class, also observed pre-PNG on the *unfixed* kernel, render-inert here.

Two further `--no-build` verification boots are in flight; results will be posted as a PR comment.

## Follow-ups (separate, pre-existing — not addressed here)
1. **Parent never notices a dead child's channel**: pid 3's exit tore down its socket (`[UNIX/TEARDOWN] id=8 peer=7 peer_connected=true`), the peer got `rx_eof/peer_closed`, yet pid 1 never closed its end (no teardown of id=7 ever), never reaped the zombie, never relaunched — on Linux the parent reaps + relaunches the socket process. Needs an NDE/aether look at peer-HUP readiness delivery into the parent's poller (or a gecko-side wait-path divergence).
2. `[TLB/TIMEOUT]` (10 M-iter unacked shootdowns, both directions) fires 54-152×/boot throughout — uncorrelated with this gate (classified OUT for this wedge), still deserves its own liveness investigation.
3. The late-shutdown `exit_group(11)` class above.

### DIFF
--- a/kernel/src/test_runner.rs
+++ b/kernel/src/test_runner.rs
@@ -1307,6 +1307,11 @@ pub fn run() -> ! {
     total += 1;
     if test_dev_dsp_ioctl_set_format() { passed += 1; }
 
+    // ── Test 520: /dev/urandom — consecutive same-tick blocks differ ───────
+
+    total += 1;
+    if test_dev_urandom_consecutive_blocks_differ() { passed += 1; }
+
     // ── Test 116: mount tmpfs + write/read + umount ────────────────────────
 
     total += 1;
@@ -26289,6 +26294,70 @@ fn test_dev_dsp_open_with_ac97_absent() -> bool {
         test_fail!("dev_dsp_absent", "expected -19 (ENODEV), got {}", fd);
         false
     }
+}
+
+// ── Test 520: /dev/urandom — consecutive same-tick blocks differ ─────────────
+//
+// Regression test for the navigation-killing RNG defect: the /dev/urandom
+// read arm in `vfs::fd_read` derived every byte purely from the 10 ms
+// scheduler tick, so two 32-byte reads inside one tick were byte-identical.
+// PKCS#11 soft-token modules run the FIPS 140-2 §4.9.2 continuous RNG
+// self-test at C_Initialize — consecutive fixed-size blocks from the system
+// RNG must differ, else the module fails with CKR_DEVICE_ERROR and every
+// TLS consumer in the process loses its crypto provider.  This test mirrors
+// that exact access pattern: open → read(32) → close, twice back-to-back
+// (guaranteed to land in the same tick at boot-test speed), plus an
+// advancing-stream check on a single open fd.
+fn test_dev_urandom_consecutive_blocks_differ() -> bool {
+    test_header!("/dev/urandom — consecutive same-tick 32-byte blocks differ");
+
+    let mut blocks = [[0u8; 32]; 2];
+    for block in blocks.iter_mut() {
+        let fd = crate::syscall::sys_open_test("/dev/urandom", 0 /* O_RDONLY */);
+        if fd < 0 {
+            test_fail!("urandom_blocks", "open(/dev/urandom) = {}", fd);
+            return false;
+        }
+        let n = crate::syscall::sys_read_test(fd as usize, block.as_mut_ptr(), 32);
+        let _ = crate::syscall::sys_close_test(fd as usize);
+        if n != 32 {
+            test_fail!("urandom_blocks", "read 32 returned {}", n);
+            return false;
+        }
+    }
+    if blocks[0] == blocks[1] {
+        test_fail!("urandom_blocks",
+            "two consecutive 32-byte reads returned identical blocks");
+        return false;
+    }
+    if blocks[0] == [0u8; 32] || blocks[1] == [0u8; 32] {
+        test_fail!("urandom_blocks", "urandom returned an all-zero block");
+        return false;
+    }
+
+    // Stream must also advance WITHIN one open file description.
+    let fd = crate::syscall::sys_open_test("/dev/urandom", 0);
+    if fd < 0 {
+        test_fail!("urandom_blocks", "re-open(/dev/urandom) = {}", fd);
+        return false;
+    }
+    let mut c = [0u8; 32];
+    let mut d = [0u8; 32];
+    let n1 = crate::syscall::sys_read_test(fd as usize, c.as_mut_ptr(), 32);
+    let n2 = crate::syscall::sys_read_test(fd as usize, d.as_mut_ptr(), 32);
+    let _ = crate::syscall::sys_close_test(fd as usize);
+    if n1 != 32 || n2 != 32 {
+        test_fail!("urandom_blocks", "same-fd reads returned {} / {}", n1, n2);
+        return false;
+    }
+    if c == d {
+        test_fail!("urandom_blocks",
+            "two 32-byte reads on one fd returned identical blocks");
+        return false;
+    }
+
+    test_pass!("/dev/urandom consecutive 32-byte blocks differ (cross-open and same-fd)");
+    true
 }
 
 // ── Test 111: /dev/dsp — open + write when AC97 present ──────────────────────
--- a/kernel/src/vfs/mod.rs
+++ b/kernel/src/vfs/mod.rs
@@ -2572,13 +2572,33 @@ pub fn fd_read(pid: crate::proc::Pid, fd_num: usize, buf: *mut u8, count: usize)
                 }
                 return Ok(count);
             }
-            // bit 24 = /dev/urandom | /dev/random  → fill with pseudo-random bytes
+            // bit 24 = /dev/urandom | /dev/random  → fill with CSPRNG bytes.
+            //
+            // The stream MUST advance between consecutive reads.  PKCS#11
+            // soft-token modules (the TLS stack Firefox ships is one) run the
+            // FIPS 140-2 §4.9.2 continuous random-number-generator self-test:
+            // each fixed-size block read from the system RNG is compared with
+            // the previous one, and the module fails C_Initialize with
+            // CKR_DEVICE_ERROR when two consecutive blocks are identical
+            // (PKCS#11 v2.40 §5.1.6 error semantics).  The previous generator
+            // here was a pure function of the 10 ms scheduler tick plus the
+            // byte index, so two 32-byte reads landing in the same tick
+            // returned byte-identical blocks → every TLS-capable process lost
+            // its crypto provider at init.  Route through
+            // `security::rand::rand_u64` instead: RDRAND-backed with a
+            // TSC-seeded fallback, and the same chokepoint that ASLR,
+            // AT_RANDOM and getrandom(2) use — so `astryx.rng_seed=`
+            // record-replay determinism covers this path too.  Per random(4),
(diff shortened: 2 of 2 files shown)


me1iissa:
## Verification results (3 fixed boots, smp=2 KVM, https://bbc.com/news)

| Boot (sid) | pid3 socket proc | NSS init | TLS to bbc.com | Navigation | PNG |
|---|---|---|---|---|---|
| c26a4bedb58f | ALIVE (0 exit_group) | OK | ClientHello + complete handshakes to 151.101.0.81 | dispatched | **972 KB, 1366×7420** |
| 58dfca07f235 | ALIVE | OK | 27 COMPLETE TLS conns: bbc.com ×2, bbcverticals static assets, 17× Akamai-edge, Mozilla services | dispatched | **975 KB, 1366×7440** |
| 3776317904d1 | ALIVE (32+ min) | OK | 1921+ pkts :443 to Fastly edges, data flowing | dispatched | not yet at 32 min — page-load held open by the pre-existing Gate-2 sub-resource stalls (`<<<GATE2` mid-handshake noClientFinished signature on optimizely/gtm-live/fastly conns); no crash, all crash classes 0 |

Unfixed kernel control: 5/5 boots (QA's 3 + two fresh at head, one with `ff-stderr-mirror`) — pid3 `exit_group(1)` in the init window, ZERO `:443` SYN, zero ClientHello, no navigation, parent looping `nsNSSComponent` ctor/dtor with `-8023`.

Crash-class check on the PNG boots: gen-abort 0, MMAP-ERR 0, bugcheck/DOUBLE_FAULT 0, CR2=0x10 0. One post-PNG `exit_group(11)` at FF shutdown on boot 1 (pre-existing class, fired pre-PNG on the unfixed kernel too; render-inert).

Test 520 passes in a `test-mode,kdb` suite boot; suite fail-set = the known ff/real-website-base group (tcp/udp loopback, glibc_hello InterpNotFound, test518b allow-fail) — nothing in the urandom blast radius. CI dispatched: run 27389202764.

PNG artifacts banked at `/tmp/bbc-fixed.png` and `/tmp/bbc-fixed-2.png`; band analysis matches the 2026-06-10 calibration win (13-14 sparse text bands; remaining blank bands = Gate-2).

Note for reviewers: verification boots must be built — the harness `start --no-build` known defect (reuses last staged kernel regardless of `--features`) burned one boot pair here; provenance was re-verified by session-kernel md5.

me1iissa (commented):
VERDICT: APPROVE

Independent review gate for PR #558 (fix/urandom-tick-determinism @ bc2d9d3, base ff/real-website). Reviewed against the authoritative git objects (git show bc2d9d3:...), not a worktree, plus one live smp=2 KVM firefox-test boot at https://bbc.com/news and CI run 27389202764.

Provenance: base 8bffee9 confirmed ancestor of bc2d9d3; PR head builds clean (release, 23s); staged kernel musl-esr match.

1. Fill-loop correctness — CORRECT. while i<count { n=(count-i).min(8); copy_nonoverlapping(8B le rand, n); i+=n } (vfs/mod.rs:2595-2602). count==0: loop skipped, Ok(0). Non-multiple-of-8 (1/7/100/1MiB): final iter copies <8B exactly, no overrun; unused high bytes of the last u64 are just discarded entropy. Large reads bounded (i advances >=1/iter). Offset/seek ignored as before (urandom is offset-independent — unchanged). EFAULT unchanged: same SMAP UserGuard bracket as the old arm; a bad user ptr faults identically.

2. Fallback entropy — ADEQUATE, with one robustness note. security::rand::rand_u64 fallback (rand.rs) is a STATELESS pure function of (rdtsc, LAPIC tick): it re-reads RDTSC every call, so two same-tick consecutive calls differ ONLY because RDTSC advances between them (the per-tick LAPIC term is constant within a tick and does NOT help same-tick non-repetition). Empirically validated: CI run 27389202764 ran under no-KVM/TCG (qemu64, NO RDRAND -> fallback path) and Test 520 PASSED ([PASS] ... cross-open and same-fd, elapsed_ticks:1 = both pairs drawn in ONE tick). RDRAND retry loop is per Intel SDM guidance (CF/setc check, 10 retries). NOTE (non-blocking): the same-tick guarantee rests entirely on RDTSC monotonic advance; a frozen/quantized-TSC environment would re-expose the FIPS continuous-RNG failure. A small persistent per-call counter mixed into the fallback seed would make it state-advancing and immune to TSC behavior. Not required for merge (TCG and -cpu host KVM both advance TSC freely; under KVM-host real RDRAND is used).

3. Record-replay interplay — NO breakage, NO ballooning. The RR model is GENERATIVE/positional, not a recorded value log: rr_rand_u64 is a deterministic xorshift64* over a single seed-from-astryx.rng_seed PRNG_STATE advanced by CAS per call; RECORD_LOG holds [SC-REC] syscall-entry records only — rand values are NOT appended, so thousands of urandom reads do NOT grow the stream. Routing urandom here makes it deterministic-by-seed (it was tick-based and NON-seed-determined before — a strict improvement for RR reproducibility). The only change is positional: urandom now consumes from the shared PRNG sequence, shifting subsequent ASLR/AT_RANDOM/getrandom draws — but RR determinism is regenerated from seed each run (no stored rand artifact to break), and multi-threaded rand is already documented non-deterministic. Acceptable.

4. Consumer sweep — COMPLETE. Both /dev/urandom AND /dev/random map to the same flag bit 0x0100_0000 (vfs/mod.rs:610-611, subsys/linux/syscall.rs:7448) -> both routed (matches random(4)). getrandom(2) (syscall/mod.rs:5385) and AT_RANDOM/ASLR (proc/elf.rs:2057) already used rand_u64 — urandom now CONSISTENT with them rather than diverging. The write-side discard (vfs/mod.rs:3135) is untouched and correct. No consumer depended on the old tick-determinism (it was never seed-controlled).

5. Performance — ACCEPTABLE. No IRQ-off section; not called with a lock held (PROCESS_TABLE lock is dropped before the device-arm match). Worst case 1 MiB = 131k RDRANDs is a userspace-driven read, preemptible; same cost profile as the existing getrandom path. No cap needed.

6. Test 520 honesty — HONEST + DETERMINISTIC. Forces same-tick (back-to-back boot-speed open/read(32)/close x2 + two reads on one fd; CI elapsed_ticks:1 proves single-tick). On the UNFIXED tick-based arm two same-tick reads are byte-identical -> blocks[0]==blocks[1] -> test_fail -> false (fails as designed). On fixed kernel: PASS in CI. Also asserts non-zero blocks.

7. Live verify (smp=2 KVM, https://bbc.com/news, sid 1ba96a57bda3): socket process pid3 SURVIVES init (alive tick=12561, sc=1914, net I/O W1596B — the pid-3 silent-early-death signature is GONE); pid4/pid5 alive (pid4 net W129KB); 3x TCP Established -> :443 (TLS sockets open — zero on the unfixed kernel); ZERO closed-class regressions (no bugcheck/gen-abort/MMAP-ERR/exit_group(11)/SIGSEGV/SIGABRT, and crucially NO CKR_DEVICE_ERROR / SEC_ERROR_PKCS11). pid3 later shows STUCK_IN_NR=7 (fstat) = the separate known Gate-2 tail, not this gate and not a regression. PNG not required per spec; TLS+pid3-survival gate MET. CI 27389202764: all 5 jobs green (incl. test-suite with Test 520 PASS under the fallback path).

Conclusion: root cause correctly identified and fixed; the new arm is correct across partial/large/zero reads; record-replay is preserved and not ballooned; both random devices routed; CI green; live boot clears the exact gate (pid3 survives, TLS to :443, no PKCS#11 device-error). The one robustness observation (stateless TSC-dependent fallback) is a follow-up nicety, not a merge blocker. APPROVE.

PR #559, opened 2026-06-12
NDE-18: Linux-parity pipe capacity + decoupled launcher-pipe drain (Gate-2 stderr convoy)

me1iissa:
## What

Gate-2 investigation (blank image slots on the BBC real-website render). Live root-cause + fix for the first binding constraint found, plus the kdb diagnostics used to find it.

## Discriminator verdict (per dispatch)

**Neither** "FF event loss with kernel-readable data" **nor** "kernel readiness loss with queued rx data" — both were tested live and exonerated:

- Wire (pcap): every stalled conn shows the **full TLS server flight delivered and ACKed in <100 ms, no gaps, no retransmits** — then no client CCS/Finished until the server FINs (~5 s). Identical conns complete with Finished +0.9 s when the stall window is over.
- Kernel (kdb `poll-revents` tight-loop): level readiness is correct (POLLIN=true whenever bytes are queued); the poll bell + 1 s resync floor function; `net-rxstats` shows rx queues drained.
- Scheduler: `ready_for_ticks` (new diagnostic) stays ≤ 37 ticks during stalls — the anti-starvation force-select works.

The actual mechanism, caught live across 4 boots with `thread-park-audit`:

1. FF's pid-1 writes **>1 MiB of stderr in the first ~40 s** of a page load.
2. fd 2 is the kernel launcher pipe — a **4096-byte ring** (POSIX minimum; pipe(7) documents 65,536 default) drained **only by `poll_output()` on the BSP main loop**, whose iteration latency is unbounded under load (observed: 636 B undrained for 56 s; drain gaps of tens of seconds).
3. Per POSIX write(2) block-on-full (landed in #551), `writev(fd=2)` parks — the **necko socket thread was caught blocked 8.7 s inside one writev mid-TLS-handshake** (`thread-park-audit`: tid 9, nr=20, arg0=0x2).
4. The libc stderr FILE lock convoys everyone else: main thread caught blocked in writev(fd=2) 2.7 s; socket thread futex-waiting the same lock word when not the writer.
5. TLS continuations miss the CDN's ~5 s handshake deadline → server FIN → subresource fetch dies → blank slots. Completions batch on drain events (4 conns' Finished at the same 10 ms).

## Fix

- **ipc/pipe**: per-pipe heap ring, default capacity 65,536 (pipe(7)); `PIPE_BUF` stays 4096 (POSIX atomicity). `pipe_set_capacity()` with fcntl(2) F_SETPIPE_SZ semantics (power-of-two round-up, EBUSY, EPERM at the 1 MiB ceiling); grow wakes parked writers.
- **fcntl(2)**: F_SETPIPE_SZ / F_GETPIPE_SZ; EBADF on non-pipe fds.
- **gui/terminal**: launcher pipe provisioned at 1 MiB; dedicated kernel drain thread (`term_out_drain`) woken directly by the child's write(2) — drain latency now bounded by scheduler fairness, not BSP-loop cadence. Framebuffer repaint throttled (20 fps) and decoupled from drain.
- **kdb (additive)**: `user-mem` (read a pid's user memory via its own page tables — used to decode live pollfd arrays), `thread-park-audit` gains `ready_for_ticks` + `syscall.arg1`.
- **tests**: Test 519 pins its pipes to 4 KiB via F_SETPIPE_SZ (capacity-independent exact-fill arithmetic); new sub-test for default capacity / round-up / EBUSY / EPERM / EBADF.

## Verification

- Mechanism level (decisive): pre-fix the launcher pipe sat **full** (4096/4096; with capacity-only, 1,048,570/1,048,576) with writers parked; post-fix (drain thread) the same load reads **buffered=0, write_waiters=0, no thread in writev** at every probe.
- Kernel suite: green; `pipe_blocking_write` PASS incl. new capacity sub-test. (One wedge of its concurrent sub-test in 3 runs at smp=2, passing on rerun — same SMP-timing-flake class as the allow-failed test518b.)
- Regression sweep (BBC boots): gen-abort / MMAP-ERR / bugcheck / exit_group(11) / CKR_DEVICE_ERROR / STACK_CANARY all **zero**.
- BBC boots on the fix: PNG produced with clean exit (1366×7423, 978 KB) on the capacity-only boot; two further boots still in the page-load tail at report time.

## What this does NOT close (residual, evidence package)

Blank images persist: client-Finished latency is still 4–15 s (boot-B distribution across 31 completed conns: median 4.5 s, p90 12.5 s, max 18.2 s; <0.5 s on Linux), so short-deadline CDNs still FIN first. With stderr exonerated, the socket thread now cycles healthily (poll/connect/writev all sub-second) and sits **in poll while handshakes await their continuation** — pointing at the async cert-verification completion path. Strongest open lead: **3 pid-1 threads spinning in sched_yield(2) at a single shared RIP (page offset +0x8a6, FP-less) for 30+ s through every stall window, cross-boot reproducible** — module resolution via procmaps was kdb-starved at report time and is the first task of the follow-up.

### DIFF
--- a/kernel/src/gui/terminal.rs
+++ b/kernel/src/gui/terminal.rs
@@ -35,8 +35,36 @@ static EXEC_PID: AtomicU64 = AtomicU64::new(0);
 static LAST_RENDER_TICK: AtomicU64 = AtomicU64::new(0);
 static RENDER_DIRTY: AtomicBool = AtomicBool::new(false);
 
-/// One-shot guard for the dedicated launcher-pipe drain thread.
+/// One-shot guard for the dedicated launcher-pipe drain thread, and the
+/// single source of truth for drain OWNERSHIP of the launcher pipe.
+///
+/// Exactly one consumer may dequeue bytes from a pipe at a time; two
+/// drainers appending to the terminal grid under separate TERMINAL-lock
+/// acquisitions produce interleaved (out-of-order) output.  Once this flag
+/// is `true` the dedicated `output_drain_thread` owns byte consumption of
+/// whatever `running_exec` currently points at, and `poll_output()` must
+/// NOT dequeue from that pipe (it keeps reaping + throttled rendering only).
+///
+/// There is never a window in which NO drainer is responsible for the pipe
+/// (the deadlock class #551 closed): the drain thread never exits — it parks
+/// when no exec is running and re-snapshots `running_exec` on every wake — so
+/// "flag is true" ⟺ "a live drainer owns the pipe".  At exit teardown
+/// `poll_output()` reclaims the final-byte drain by clearing `running_exec`
+/// FIRST (the drain thread then snapshots `None` and releases the pipe) and
+/// only then doing the synchronous tail read.  `PIPE_TABLE`'s lock makes
+/// every read indivisible, so a drain-thread iteration already in flight at
+/// that instant cannot lose or duplicate a byte with the tail read; the only
+/// residual is a possible ordering blip on the very last exit chunk — never
+/// the steady-state interleave this model exists to prevent.
 static DRAIN_THREAD_STARTED: AtomicBool = AtomicBool::new(false);
+
+/// True when the dedicated drain thread owns byte consumption of the
+/// launcher pipe (so `poll_output()` must skip its own dequeue).  See
+/// [`DRAIN_THREAD_STARTED`] for the ownership model.
+#[inline]
+fn drain_thread_owns_pipe() -> bool {
+    DRAIN_THREAD_STARTED.load(Ordering::Acquire)
+}
 
 /// Dedicated kernel-thread drain loop for the launcher stdout/stderr pipe.
 ///
@@ -1102,17 +1130,20 @@ fn spawn_async(cmd: &str) -> Result<(u64, u64), alloc::string::String> {
 
     // Attach pipe to stdout/stderr while the child is still blocked.
     let pipe_id = crate::ipc::pipe::create_pipe();
-    // The launcher pipe's ONLY reader is `poll_output()` on the BSP main
-    // loop, whose iteration latency is unbounded under load (it shares the
-    // loop with VFS probes, the compositor, and net polling — observed
-    // gaps of tens of seconds during a Firefox page load).  Since POSIX
-    // write(2) blocks a writer on a full pipe, an undersized ring turns
-    // the child's stderr into a synchronization point: every thread that
-    // logs a warning parks behind the libc stderr FILE lock until the BSP
-    // loop next drains (observed: the necko socket thread frozen for 8.7 s
-    // mid-TLS-handshake).  Provision the maximum capacity (fcntl(2)
-    // F_SETPIPE_SZ ceiling) so a whole page-load's diagnostic burst fits
-    // without ever blocking the writer.
+    // The launcher pipe is drained by the dedicated `output_drain_thread`
+    // (see `ensure_output_drain_thread`), which owns byte consumption once
+    // started; `poll_output()` then only reaps the child and renders the
+    // grid (drain ownership is single — see `DRAIN_THREAD_STARTED`).  The
+    // earlier design drained solely on the BSP main loop, whose iteration
+    // latency is unbounded under load (it shares the loop with VFS probes,
+    // the compositor, and net polling — observed gaps of tens of seconds
+    // during a Firefox page load).  Since POSIX write(2) blocks a writer on
+    // a full pipe, an undersized ring turns the child's stderr into a
+    // synchronization point: every thread that logs a warning parks behind
+    // the libc stderr FILE lock until the pipe next drains (observed: the
+    // necko socket thread frozen for 8.7 s mid-TLS-handshake).  Provision
+    // the maximum capacity (fcntl(2) F_SETPIPE_SZ ceiling) so a whole
+    // page-load's diagnostic burst fits without ever blocking the writer.
     let _ = crate::ipc::pipe::pipe_set_capacity(
         pipe_id, crate::ipc::pipe::PIPE_MAX_CAPACITY);
     crate::proc::attach_stdout_pipe(pid, pipe_id);
@@ -1353,19 +1384,29 @@ pub fn poll_output() {
     let mut raw = [0u8; 4096];
     let mut any_data = false;
 
-    // First pass: read all available data before locking TERMINAL.
-    // `pipe_read_wake` (NOT raw `pipe_read`): this drain is the ONLY
-    // consumer of the launcher pipe, and the child-side writers block on a
-    // full pipe per POSIX write(2)/pipe(7).  A drain that frees room
+    // Steady-state drain ownership: when the dedicated `output_drain_thread`
+    // owns the launcher pipe it is the SOLE byte consumer (see
+    // `DRAIN_THREAD_STARTED`); `poll_output()` must NOT also dequeue, or the
+    // two drainers append to the terminal grid under separate TERMINAL-lock
+    // acquisitions with no ordering and the output interleaves.  In that
+    // mode `poll_output()` does only the throttled render (the drain thread
+    // sets `RENDER_DIRTY`) plus the reap/teardown below.  When no drain
+    // thread owns the pipe (e.g. an exec launched before one was started, or
+    // a build without it) `poll_output()` remains the drainer.
+    //
+    // `pipe_read_wake` (NOT raw `pipe_read`): the child-side writers block on
+    // a full pipe per POSIX.1-2017 write(2)/pipe(7).  A drain that frees room
     // without waking the parked writers strands every child stdout/stderr
     // writer forever (observed live as all Firefox processes wedged in
     // writev(fd=2) once the 4 KiB pipe filled).
     let mut chunks: alloc::vec::Vec<alloc::vec::Vec<u8>> = alloc::vec::Vec::new();
-    loop {
-        let n = crate::ipc::pipe::pipe_read_wake(pipe_id, &mut raw).unwrap_or(0);
-        if n == 0 { break; }
-        chunks.push(raw[..n].to_vec());
-        any_data = true;
+    if !drain_thread_owns_pipe() {
+        loop {
+            let n = crate::ipc::pipe::pipe_read_wake(pipe_id, &mut raw).unwrap_or(0);
+            if n == 0 { break; }
+            chunks.push(raw[..n].to_vec());
+            any_data = true;
+        }
     }
 
     // Non-blocking waitpid: did the child become a zombie?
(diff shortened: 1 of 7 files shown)


me1iissa (commented):
VERDICT: REQUEST-CHANGES

Independent review gate for PR #559 (`nde/gate2-image-cdn-stall` @ 3552484, base dcd7456). Built at 3552484 (provenance confirmed: staged kernel contains `user-mem`, `F_SETPIPE_SZ`, capacity sub-test strings, `term_out_drain`). The pipe-capacity / fcntl / blocking-writer core is correct and test-green; one **new unguarded concurrency defect** and one now-false load-bearing comment need a small follow-up before merge. Detail below, with the static analysis (Q1–Q3, Q6) standing on its own and the live-verification caveat called out honestly.

## What is solid (verified)

- **Buffer lifecycle / resize (Q1)** — `Pipe::buffer` heap `Vec`; `set_capacity` reallocs + linearises live bytes into the new ring entirely under `PIPE_TABLE.lock()` (`pipe.rs:96`), so realloc is atomic vs. concurrent read/write/atomic-write. No half-state: old buffer stays intact until `self.buffer = nb`. Full-ring edge (`count==cap → write_pos = count%cap = 0`) is the correct representation. OOM aborts (alloc returns null → `handle_alloc_error`, `panic="abort"`) rather than ENOMEM — consistent with every other kernel allocation, and bounded (≤1 MiB ceiling); informational, not blocking.
- **F_SETPIPE_SZ / F_GETPIPE_SZ (Q2)** — round-up-to-pow2 with floor `PIPE_BUF`=4096; EBUSY when `count>cap` (`pipe.rs:105` -16); EPERM beyond 1 MiB (`pipe.rs:101` -1); EBADF via `is_pipe_fd` guard before `get_pipe_id` (`syscall.rs:8301/8318`). F_GETPIPE_SZ returns `buffer.len()`. **Test-verified**: capacity sub-test passed (default 65536, round-up 5000→8192, shrink-EBUSY, >1MiB-EPERM, non-pipe-EBADF).
- **Blocking-writer interplay / shrink-livelock (Q3)** — `need = if atomic {count} else {1}` (`syscall.rs:7207`), `atomic` ⇔ `count ≤ PIPE_BUF` (4096). Since `set_capacity`'s floor is also 4096, `need ≤ 4096 ≤ capacity` ALWAYS — the shrink-below-need livelock is structurally impossible. Atomicity contract (`pipe_write_atomic` checks `space() ≥ data.len()` against live `buffer.len()`) is correct at any capacity. Grow wakes parked writers (`pipe_set_capacity` → `wake_writers_all`). **Test-verified** (atomic-park + concurrent-atomicity sub-tests green).
- **fcntl/sample/kdb (Q5 partial)** — `sample.rs` arg1 plumbing and `kdb user-mem` (read-only target-CR3 page-walk via `virt_to_phys_in` through the direct map, ≤4096, user-half only, fails-closed on unmapped) are diagnostics-only, no semantic kernel impact. NB: `record_syscall` doc still says "four ... atomics share a cache line" — now FIVE stores; the five syscall atomics (tid@0,tick@8,nr@24,arg0@32,arg1@40) DO fit the first 64B line so the ordering guarantee holds, but the comment is numerically stale (cosmetic).
- **Test honesty (Q6)** — the pipe blocking-write test is re-pinned to 4 KiB via real `F_SETPIPE_SZ(4096)` syscalls BEFORE the writers run; the original block-on-full semantics are exercised faithfully (no fabricated wakes — sub-tests assert real `wait_writable`/`wake_writers_all` state). `test518b` is the known SMP-timing flake, **explicitly allow-listed** in `ci/allow-fail.json`; did not fire on my run.

## Blocking finding

**[MEDIUM] Unguarded double-drain of the launcher pipe — `gui/terminal.rs`.**
This PR adds `output_drain_thread` (`term_out_drain`, terminal.rs:57) as a SECOND concurrent consumer of the launcher pipe, but does NOT gate the pre-existing `poll_output()` drain: `poll_output` still drains unconditionally whenever `EXEC_RUNNING` is true (terminal.rs:1335, loop at :1363 `pipe_read_wake`), with no `DRAIN_THREAD_STARTED` guard. Both drainers call `pipe_read`/`pipe_read_wake` (data integrity is SAFE — each byte consumed once under `PIPE_TABLE.lock()`; writer-wake SAFE; teardown SAFE — no UAF, drain thread sees `Gone`), but they append to the TERMINAL grid under SEPARATE `TERMINAL.lock()` sections with no ordering between the two read→append sequences. Two drainers reading adjacent ring segments can append OUT OF ORDER → scrambled terminal output. For the headless FF launcher (the PR's target) this is invisible/harmless; for the **interactive shell/terminal demos (Q5)** — which also go through `spawn_async`→`ensure_output_drain_thread` (terminal.rs:399→1120) — it is a real display-ordering regression, and there is no test covering terminal output ordering. Fix is small: gate `poll_output`'s drain off once the drain thread owns the pipe (or hand draining fully to the thread, leaving poll_output only reaping+render).

**[LOW, fix-with-above] Stale/false load-bearing comment.** terminal.rs:1105 still asserts "The launcher pipe's ONLY reader is `poll_output()`" — false after this PR adds `term_out_drain`. Same false premise in `pipe.rs:235` and the `poll_output` first-pass comment (terminal.rs:1356 "this drain is the ONLY consumer"). Update to reflect the two-consumer reality. Also fix `sample.rs` "four atomics" → five.

## Bugcheck observed — investigated, attributed PRE-EXISTING (not blocking)

Live KVM run #1 (`firefox-test,kdb`, bbc.com) hit `*** AETHER KERNEL BUGCHECK *** 0xdead0007 KERNEL_GPF`: #GP error_code=0x8000 at garbage `RIP=0x700f` on cpu=1, in `sched_starve_worker` PID 80/TID 174 — i.e. inside the in-kernel `sched-starvation` TEST, BEFORE the launcher path runs. Per the GDB-autopsy-first rule I re-ran under the GDB stub (autopsy on `ke_bugcheck`): **0 hits — the PR build ran clean through the same region on run #2**. Non-reproducible across 2 PR runs ⇒ a non-deterministic SMP race, in a path the PR does not touch (pipe/terminal/fcntl/kdb/sample), matching the known recurring teardown-reap class already patched on this branch (`3b3533b` "don't reap a Running-on-another-CPU sibling's live kernel stack on group exit"). Confidence: high that this is pre-existing and not PR-attributable.

## Live-verification GAP (disclosed)

I could NOT live-confirm the PR's actual mechanism end-to-end against bbc.com. `--features firefox-test` boots test-mode + the suite's `firefox_oracle`, which launches FF via `create_user_process_with_args_blocked` with NO launcher pipe, then enters the KDB loop — it never reaches the real `launch_process`→`spawn_async`→`term_out_drain`+1 MiB-pin path (that lives in the `firefox-test-core`-only FF-demo boot, main.rs:1382 / cfg block ~1077). So "no thread >1s in writev(fd=2); pipe not pinned at capacity; socket thread cycling" remains verified by STATIC + UNIT-TEST evidence (the kernel-drain-wake sub-test passed) but not by a live bbc.com page-load capture. (Other failures in my suite run — Musl hello / TCC / busybox / sigchld / glibc_hello / dynamic_elf / monotonic-rate — are environmental: I symlinked the FF-real-website `data.img`, which lacks `/disk/bin/*` test binaries; not PR regressions.)

## Ask

Land the two-consumer guard + comment corrections (small, ≤~30 LOC). The capacity/fcntl/blocking-writer core is correct and I'd APPROVE it as-is; the double-drain is a genuine new concurrency defect in shared terminal/shell code that should be closed before merge. Re-request once the drain ownership is gated.

me1iissa:
## Review gate (blocking finding) addressed — single drain ownership for the launcher pipe

Pushed `5383208` to `nde/gate2-image-cdn-stall`.

**The finding:** this PR added `output_drain_thread` (`term_out_drain`) as a second concurrent consumer of the launcher pipe but left the pre-existing `poll_output()` drain unconditional. Byte integrity was already safe (each `pipe_read` is indivisible under `PIPE_TABLE`'s lock), but the two drainers append to the TERMINAL grid under separate lock acquisitions with no ordering → **out-of-order terminal output** for the interactive shell demos (which route through `spawn_async` → `ensure_output_drain_thread`).

**The fix** — establish single drain ownership per pipe:

- `kernel/src/gui/terminal.rs:39` — `DRAIN_THREAD_STARTED` is now the single source of truth for drain ownership; new `drain_thread_owns_pipe()` helper at `terminal.rs:60`.
- `kernel/src/gui/terminal.rs:1398` — `poll_output()`'s steady-state dequeue loop is now gated `if !drain_thread_owns_pipe()`. Once the dedicated thread is started (first `spawn_async`), it is the SOLE byte consumer; `poll_output()` does only the reap + the throttled render the drain thread flags via `RENDER_DIRTY`.
- `kernel/src/gui/terminal.rs:1457` — exit-teardown handoff: `poll_output()` (the reaper) clears `running_exec` FIRST so the drain thread releases the pipe on its next snapshot, then does the synchronous final-byte read as sole consumer.

**Safety (no-undrained-window, preserves #551's block-on-full fix):** the drain thread never exits — it parks when no exec is running and re-snapshots `running_exec` on every wake, and a child `write(2)` wakes it directly (`wake_readers_all`). So "flag is true" ⟺ "a live drainer owns the pipe"; when no thread owns it (pre-first-launch) `poll_output()` remains the drainer. There is no state in which neither consumer is responsible. At teardown, `PIPE_TABLE`'s lock makes the read indivisible, so an in-flight drain-thread iteration cannot lose/duplicate a byte — the only residual is a possible ordering blip on the very last exit chunk, never the steady-state interleave.

Both now-false comments corrected (`spawn_async` at `terminal.rs:1129` and `pipe::pipe_read_wake` at `pipe.rs:234`) — they no longer claim `poll_output()` is the launcher pipe's ONLY reader.

The capacity / fcntl / blocking-writer core was not touched.

### Verification (all via `scripts/qemu-harness.py`)
- **Clean build** of the final tree (`firefox-test-core`, smp=2); staged kernel strings-check confirms `term_out_drain` present.
- **In-kernel suite, smp=2:** `pipe blocking write — … / concurrent-atomicity / kernel-drain-wake / EAGAIN / EPIPE` **PASS**, incl. the new capacity sub-test (`default 65536, round-up 5000→8192, shrink-EBUSY, >1MiB-EPERM, non-pipe-EBADF ✓`). The only failures were the allow-listed `test518b` SMP flake and `/disk/bin/*: NotFound` data-image artifacts (worktree data.img lacked the compiled test binaries) — none in the terminal/pipe path.
- **Interactive-path:** the interactive desktop shell isn't reachable via the harness in either standard boot config, so verified by code-walk that exactly one drainer can own a pipe at any instant (proof above) — confirmed explicitly.
- **smp=2 `firefox-test-core` BBC boot:** reached `exit_group(0)` clean; **no** `writev(fd=2)` convoy / blocked-writer signature; closed-class regression sweep (gen-abort / MMAP-ERR / bugcheck / `exit_group(11)` / CKR_DEVICE_ERROR / STACK_CANARY / panic) **zero**.

Not merging — leaving for review.

me1iissa (commented):
VERDICT: APPROVE (focused re-review of 5383208, fix-and-resubmit delta over 3552484)

The single blocking finding from the prior review (double-drain: poll_output + the drain thread both consuming the launcher pipe unconditionally → out-of-order terminal output) is **resolved**, with no new hazards. The previously-cleared core (capacity / fcntl / blocking-writer) was not re-reviewed.

Make-or-break checks (all pass):

1. First-launch window — no ownerless deadlock. The drain thread sources the pipe only from `running_exec` (terminal.rs:91-93), which the launch caller sets *after* `spawn_async` returns; until then the thread snapshots `None` and parks in a 100 ms-bounded `wait_poll_event`. The pipe is sized to `PIPE_MAX_CAPACITY` before `unblock_process`, so a child writing in the ≤100 ms gap does not block unless it exhausts the full max ring; even then the thread re-snapshots within ≤100 ms and `pipe_read_wake → wake_writers_all` unparks the writer. Worst case is a bounded latency blip, never a permanent strand. The wake source is the bounded resync, not a one-shot edge — so setting the flag inside `ensure_output_drain_thread` before the thread is provably parked is safe.

2. Steady-state exclusivity — the regression is gone. The dequeue loop (poll_output:1403-1409) is gated `if !drain_thread_owns_pipe()`; in steady state the flag is true so poll_output never dequeues (reap + throttled render only). Repo-wide grep confirms the launcher pipe id is private to terminal.rs (every demo/shell consumer creates its own pipe). The only two-consumer overlap is the teardown tail-read, correctly sequenced (running_exec cleared FIRST under the TERMINAL lock; `PIPE_TABLE` lock makes each read indivisible) → at most a single-chunk ordering blip on the last exit bytes; the steady-state interleave cannot occur. The author bound is exact.

3. Wake-chain (#551) preserved. The drain thread reads via `pipe_read_wake` (terminal.rs:104), which wakes parked writers on any nonzero drain; the now-gated poll_output path carried no wake the drain path lacks.

4. Comments accurate. Both flagged stale ONLY-consumer comments rewritten; surviving "SOLE byte consumer" (1388) and the past-tense "drained solely on the BSP main loop" (1137) are correct; pipe.rs `pipe_read_wake` doc updated. No stale ONLY-reader claims remain (grep-confirmed).

5. Live verification at 5383208, smp=2 (provenance: `term_out_drain` string present in the booted kernel.bin): `pipe blocking write` PASS incl. the 12288 B blocks-until-drained capacity sub-test + concurrent-atomicity + `kernel-drain-wake`; `pipe wake hook` PASS (kernel-context supervisor drain woke parked writer). No panic / bugcheck / interleave / leaked-writer / deadlock. The other suite FAILs (`/disk/bin/*` NotFound, `test518b` known allow-fail, `monotonic-rate` flake) are pre-existing/environmental and unrelated to the two-file delta.

PR #560, opened 2026-06-12
Gate-2 TLS Finished-latency: kdb observer fix, sched wake telemetry, gated one-shot event-wake boost

me1iissa:
## What this is

Three coupled changes from the Gate-2 (blank image slots / TLS client-Finished latency) investigation. **Do not merge without coordinator review** — the boost half is a measured-but-gated mechanism with an honest caveat (below).

### 1. kdb pump: dispatch outside `KDB_SESSIONS` (observer-effect fix — proven)

`kdb::pump()` ran `handle_request` while holding the `KDB_SESSIONS` spinlock. `pump()` is entered concurrently by the BSP main loop (`net::poll`) and the dedicated PRIORITY_HIGH pump thread, so any long op (a 3 s `rip-trace`, a `thread-park-audit` walking every thread + fd table) pinned the **sibling CPU in a Ring-0 pause-spin for the entire op**. GDB-stub sampling (no guest cooperation): 134/160 samples = ~95% of BOTH CPUs inside the `KDB_SESSIONS` acquire spin at `pump+0x92` during one rip-trace.

Consequence for past data: stall-triggered kdb probe batteries **amplified the very TLS stalls being measured**. Probed boots: client-Finished median 4.4–7.2 s. Unprobed boot, same kernel: **0.99 s**. Any historical latency number taken under heavy kdb load needs re-grading.

Fix: claim/dispatch/commit split (new `dispatching` session flag); lock hold times are now buffer-scans only. kdb is usable as a non-perturbing probe again.

### 2. `sched-stats` telemetry (additive)

New kdb op: log2 **pick-wait histogram** (run-queue wait of every dispatched thread, recorded at pick time), wake-kick + starve-force counters, ready-queue depth, gate states. Harness whitelist extended (additive).

### 3. One-shot event-wake boost (gated; default ON only for `firefox-test-core`)

All event wakes (futex `WAKE/WAKE_OP/REQUEUE`, CLEARTID exit wake, cluster compensation, per-fd waitlists / poll bell) now go through `proc::wake_ready_event`: the existing `PRIORITY_BOOST_WAIT` bump, capped at `base+2`, **consumed entirely at dispatch** (`Thread::wake_boosted`, cleared by the picker). Pure queue-jump for the wakeup-preemption window; steady-state shares stay fair. Per POSIX sched(7) (SCHED_OTHER prompt contention) and pthread_cond_signal(3p).

Two stronger variants were built and **rejected by measurement**:
| variant | result |
|---|---|
| decaying (ratcheting) boost + reschedule kick | wake-frequent threads became a permanent priority class; force-select waits up to 557 ticks on tid 0; TLS median 0.99 s → **11.8 s**, stalled handshakes 3 → 29 |
| one-shot boost + kick | the kick evicts the BSP IO-pump (tid 0) on virtually every wake → TCP processing collapse; image-CDN median **12.3 s** |
| **one-shot boost, no kick** | **shipped** (kick machinery kept behind `WAKE_KICK_ENABLED`, default OFF) |

Measured with the shipped variant (BBC `/news`, smp=2 KVM, 4 boots):
- First full-page PNG of the session at **~93 s wall** (974 KB, 1366×7475) — previous baseline retest: 0/3 PNGs in 9-minute windows.
- Wire (pcap, pre-crash where applicable): client-Finished median 0.61–3.10 s; image-CDN (akamai) median 1.30–5.45 s, best boot 26/27 handshakes complete, max 4.55 s — all inside the CDN FIN deadline; stalled 1–14 (vs 3–29 before).

**Caveat (why it's gated):** 3 of 4 boost-only boots ended in the *known* flaky libxul navigation-phase crash family (three distinct fault sites, all documented before this work; `exit_group(11)`). vs ~1/3 failures at base (n=3) this is directionally worse but not significant (n too small). The boost re-orders dispatch and plausibly raises the latent race's hit rate. Hence: default ON only for `firefox-test-core`, OFF for production/test-mode, runtime `set_wake_boost` / kdb-visible gate state for cheap A/B. Root-causing that crash family is the right next dispatch and is **pre-existing scope**, not introduced here.

### Images verdict (Gate-2 status)

Narrowed, **not closed**: the PNG from the successful boot still has blank image bands. Wire timeline shows the first image-CDN burst (7 parallel ClientHellos during the page-busy phase) making zero handshake progress for ~11 s until the CDN FINs; retries complete (one pulled 1.41 MB) but land after the snapshot fires. Remaining lead: socket-process TLS continuation stalls during the busy phase (cert-verify proxied to the busy parent is the prime suspect) — now cheaply probeable since kdb no longer freezes a CPU.

### Tests

- New test 520: boost cap/idempotence/one-shot flag, kick gating (inert when off, NEED_RESCHEDULE when on), prio-0-never-kicks.
- Full test-mode suite: **401-402/411 with the changes vs 401/410 at base** — identical failure lists (environmental data.img gaps + known test518b flake + tcp_ooo_fin_guard seen on both arms). One single-run flake observed once in four suite runs (`eventfd2_nonblock` got fd 0 + cross-context EBADF, with another process's #PF interleaved) — pre-existing fragility class; the boost is compiled-default-OFF in test-mode so the wake path there is behaviourally historical.

Artifacts: pcaps `/tmp/gate2_pse_*.pcap`, distribution tool `/tmp/finlat.py`, PNG `/tmp/gate2_run4_out.png`.

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -1088,6 +1088,7 @@ pub extern "C" fn ap_rust_entry() -> ! {
             last_cpu: 0,
             first_run: false,
             ready_since_tick: 0,
+            wake_boosted: false,
             ctx_rsp_valid: alloc::boxed::Box::new(core::sync::atomic::AtomicBool::new(true)),
             clear_child_tid: 0,
             fork_user_regs: crate::proc::ForkUserRegs::default(),
--- a/kernel/src/ipc/waitlist.rs
+++ b/kernel/src/ipc/waitlist.rs
@@ -128,13 +128,26 @@ pub fn wake_tids(tids: &[u64]) {
         return;
     }
     let mut threads = crate::proc::THREAD_TABLE.lock();
+    // Event wake: boost-and-ready each parked waiter, tracking the highest
+    // effective priority so the wakeup-preemption kick below can ask any
+    // CPU running lower-priority work to reschedule promptly instead of
+    // letting the woken thread wait out a full quantum (see
+    // `sched::kick_preempt_for_wake`).
+    let mut max_prio: u8 = 0;
+    let mut any_woken = false;
     for &t in tids {
         if let Some(th) = threads.iter_mut().find(|th| th.tid == t) {
             if th.state == crate::proc::ThreadState::Blocked {
-                th.state = crate::proc::ThreadState::Ready;
-                th.wake_tick = 0;
+                crate::proc::wake_ready_event(th);
+                if th.priority > max_prio {
+                    max_prio = th.priority;
+                }
+                any_woken = true;
             }
         }
+    }
+    if any_woken {
+        crate::sched::kick_preempt_for_wake(&threads, max_prio);
     }
 }
 
--- a/kernel/src/kdb.rs
+++ b/kernel/src/kdb.rs
@@ -1561,7 +1561,11 @@ fn op_sched_stats(out: &mut String) {
     out.push_str("],");
     let _ = write!(out, r#""wake_kicks":{},"#,    crate::sched::wake_kick_count());
     let _ = write!(out, r#""starve_forces":{},"#, crate::sched::starve_force_count());
-    let _ = write!(out, r#""ready_depth":{}"#,    crate::sched::ready_depth());
+    let _ = write!(out, r#""ready_depth":{},"#,   crate::sched::ready_depth());
+    let _ = write!(out, r#""wake_boost_enabled":{},"#,
+        crate::sched::wake_boost_enabled());
+    let _ = write!(out, r#""wake_kick_enabled":{}"#,
+        crate::sched::WAKE_KICK_ENABLED.load(core::sync::atomic::Ordering::Relaxed));
     out.push('}');
 }
 
--- a/kernel/src/kdb.rs
+++ b/kernel/src/kdb.rs
@@ -74,6 +74,12 @@ struct PendingSession {
     local_port:  u16,
     buf:         Vec<u8>,
     responded:   bool,
+    /// A request line has been claimed for dispatch but the response has not
+    /// been sent yet.  Set under `KDB_SESSIONS` before the dispatcher runs
+    /// OUTSIDE the lock; prevents a concurrent pump caller from re-dispatching
+    /// the same line, and keeps the deferred-close logic (which keys off
+    /// `responded`) from FIN-ing the connection mid-dispatch.
+    dispatching: bool,
 }
 
 static KDB_SESSIONS: Mutex<Vec<PendingSession>> = Mutex::new(Vec::new());
@@ -182,7 +188,7 @@ pub fn pump() {
             if !ss.iter().any(|s| s.remote_ip == *rip && s.remote_port == *rp) {
                 ss.push(PendingSession {
                     remote_ip: *rip, remote_port: *rp, local_port: KDB_PORT,
-                    buf: Vec::new(), responded: false,
+                    buf: Vec::new(), responded: false, dispatching: false,
                 });
             }
         }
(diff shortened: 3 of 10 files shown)


me1iissa (commented):
VERDICT: APPROVE (independent review gate)

Reviewed all three parts against authoritative git objects (git show 7fd71d1:...) plus a live firefox-test-core BBC boot built and provenance-verified FROM the PR worktree (kernel.bin strings contain pick_wait_hist/wake_kicks/starve_forces/wake_boost_enabled/wake_kick_enabled). All six review questions resolved; no blocking findings.

== (a) kdb observer-effect fix — CORRECT ==
claim/dispatch/commit split is sound. The claim scan sets s.dispatching=true UNDER the lock before pushing to `work`; a second concurrent pump() entrant sees `s.responded || s.dispatching` and skips, so a line is claimed exactly once (no double-dispatch, no interleave). The request line is copied (to_vec()) into owned `work`, so handle_request runs lock-free on owned data with no reference into the session buffer — no use-after-free if the session is reaped mid-dispatch (the commit-side find() returns None and no-ops). The lock now protects only the KDB_SESSIONS Vec structure + per-session flags across buffer scans = bounded hold time. dispatching is a correct single-in-flight serializer for the one-request/one-response kdb protocol; no stuck state.
LIVE: sched-stats kdb op returned in ~0.35-0.40 s under heavy FF load without freezing net/X11/compositor — the fix works as advertised.

== (c) one-shot event-wake boost — SAFE ==
(i) Capture/ping-pong PREVENTED: boost only applies to a Blocked thread (state guard) and is capped at base+PRIORITY_BOOST_WAIT via `if priority < cap`. To be re-boosted a thread must re-enter Blocked, which requires it to have run, which consumes the boost at dispatch (priority reset to base_priority, flag cleared). Cap is genuinely base+2, never base+4. Test 520 asserts exactly this ("repeat wake escalated past base + boost cap").
(ii) PRIORITY_HIGH kernel threads: boost is capped at base+2; a PRIORITY_NORMAL (8) user thread reaches at most 10, still below PRIORITY_HIGH (13). Cannot preempt term_out_drain / kdb pump.
(iii) Never-boosted CPU-bound threads: STARVE_FORCE_TICKS backstop intact and ACTIVE — live sched-stats showed starve_forces climbing 1 -> 482 under load, proving long-waiting threads still get forced progress.
(iv) FIFO/fairness: boost is consumed at dispatch and does NOT touch ready_since_tick; equal-base-priority ordering (longest-waiting-wins via wait_age_bonus) is identical to baseline once the boost window closes. Existing outgoing-thread decay (cur.priority -= 1) never conflicts with the consume (a formerly-boosted thread is already at base+flag-cleared by the time it is the outgoing cur).

== Gating hygiene — VERIFIED ==
Cargo default=[] so cfg!(feature="firefox-test-core") is false; WAKE_BOOST_ENABLED defaults false in default and test-mode builds. wake_ready_event sets state=Ready + wake_tick=0 BEFORE the gate check, so with boost OFF it is byte-identical to the historical bare flip. test-mode does NOT pull firefox-test-core, and test 520 set_wake_boost(true) then restores boost_was (false) before every return path — no scheduling change leaks to the suite. WAKE_KICK_ENABLED defaults false EVERYWHERE.
LIVE on the PR build: wake_boost_enabled=true, wake_kick_enabled=false, wake_kicks=0 after 37k+ picks — the kick path is genuinely inert when gated off.

== (b) sched-stats + harness whitelist — ADDITIVE ==
New kdb op + two harness whitelist entries + argparse choices entry. No field renames, no behavior change to existing ops. pick-wait bucket math verified in-bounds (age.min(255) + .min(BUCKETS-1), no OOB/panic) and matches the documented buckets exactly. Single relaxed atomic increment on the pick path — negligible.

== Wake-site sweep (q5) — CONSISTENT ==
All event-wake paths route through wake_tids/wake_ready_event: eventfd, pipe (x2), virtio_serial, poll-bell, the 3 sys_futex_linux drains, futex_wake_for_exit, futex_cluster. The 5 UNCONVERTED bare Blocked->Ready sites (vfork parent-wake + waitid/wait race-recovery self-wakes) are correctly left alone — they set wake_tick=u64::MAX sentinel semantics that wake_ready_event (wake_tick=0) would corrupt, and they are not productive event wakes. None double-wake (all guard on state==Blocked). wake_ready_event no-ops on a non-Blocked target (Running/Dead) — no state clobber, park-protocol/ctx_rsp_valid invariants preserved. Lock discipline verified: all 3 futex wake_tids call sites release FUTEX_WAITERS (explicit drop or block-close) BEFORE wake_tids takes THREAD_TABLE — FUTEX_WAITERS -> THREAD_TABLE order honoured, no double-lock. kick_preempt_for_wake only sets NEED_RESCHEDULE atomics under the held table; NEED_RESCHEDULE is consumed only at timer-ISR Ring-3 return (explicitly skips kernel-mode) and the syscall tail after locks release — cannot interrupt a kernel critical section.

== Live verify ==
PR build, smp=2 KVM, BBC: ZERO kernel-side fault classes (no bugcheck / ring-0 #GP / __stack_chk_fail / STACK_CANARY / panic / assert), zero CKR_DEVICE_ERROR / gen-abort / MMAP-ERR. The boot terminated in the KNOWN pre-existing flaky libxul navigation-phase crash family (PID 1 zombied, 1 userspace SIGSEGV PID 7 delivered correctly via SA_SIGINFO handler — content-proc NULL-class fault, not a kernel fault) — documented 2026-06-04, NOT a regression. Kernel stayed healthy throughout (HB advancing, TCP still accepting, kdb sub-second). No new panic class.

Recommendation: MERGE. The (a) observer-effect fix is an unambiguous diagnostic-correctness win. The (b) telemetry is clean additive. The (c) boost is correctly gated, capture-proof, starvation-safe, FIFO-preserving, and behaviourally inert when off — the author honest grading (plausible-not-proven on latency, n too small) is accurate but the change is SAFE regardless of whether the latency win replicates, since it cannot degrade non-firefox-test-core builds and the kick stays off. The flaky-family hit-rate concern is real but is a pre-existing userspace bug the boost only re-orders around — appropriately kept toggleable.

PR #561, opened 2026-06-15
net/tcp: deliver in-order data + FIN in FIN-WAIT-1/FIN-WAIT-2 (NDE-18)

me1iissa:
## Summary

FIN-WAIT-1 and FIN-WAIT-2 are receive states. Per **RFC 9293 §3.10.7.4**, an in-order segment carrying data text must be queued for the application and acknowledged — exactly as in ESTABLISHED — before the segment's FIN is processed. The previous FIN-WAIT arms in `process_segment` honoured a peer FIN **only when it sat bare at `recv_next`** (`seq + payload_len == recv_next`) and delivered no segment text at all.

A peer that **coalesces its final application data with its FIN** into one segment (a common TLS close: a ~24–31 byte `close_notify` record riding the same segment as the FIN) therefore failed the bare-FIN test. The data and FIN were dropped: no ACK was emitted, `recv_next` never advanced, and the connection never reached TIME-WAIT. The peer retransmitted the data+FIN segment dozens of times over ~10 minutes and then aborted with **RST** (RFC 9293 §3.10.7). A local fetch on such a connection never reached a terminal state, so a client awaiting that resource waited indefinitely.

A peer that sends a **bare FIN** (no coalesced data) closed cleanly through the old path — so whether a server's close wedged depended solely on its TCP close framing, not on the host.

## Wire evidence

Decoded the wedge capture and isolated each 4-tuple. In the **same** capture:

| Flow | Server | Close framing | Result |
|------|--------|---------------|--------|
| normandy | 34.49.51.44:443 (Google) | **bare FIN** (len=0) after our FIN | ACKed → clean close ✓ |
| aus5 / classify-client / location.services | 199.232.57.91:443 (Fastly) | **24–31 B appdata + FIN coalesced** | dropped → no ACK → **47× retransmit over ~600 s → RST** ✗ |
| services.addons | 151.101.1.91:443 (Fastly) | data + FIN coalesced | same ✗ |

The failing connections completed their TLS handshake and exchanged application data; they only wedged on **close**.

## Fix

Extract the ESTABLISHED receive logic (in-order append + `recv_next` advance + out-of-order drain, OOO buffering, stale/dup re-ACK, in-order FIN consume) into a shared `receive_segment_data()` helper, called from ESTABLISHED, FIN-WAIT-1 and FIN-WAIT-2. Each caller keeps its state-specific FIN transition:

- **ESTABLISHED** → CloseWait on an in-order peer FIN
- **FIN-WAIT-1** → TIME-WAIT on an in-order peer FIN (this stack has no CLOSING state, so a peer FIN preceding the ACK of our own FIN folds directly into TIME-WAIT); else FIN-WAIT-2 on a pure ACK
- **FIN-WAIT-2** → TIME-WAIT on an in-order peer FIN

FIN-WAIT-2 also now processes incoming ACKs. Out-of-order FIN gating (RFC 9293 §3.10.7.4) is preserved by the shared helper.

This is the FIN-WAIT analogue of the ESTABLISHED/CloseWait peer-FIN EOF edges.

## Test

`test_tcp_finwait_data_fin_delivery` (kdb-gated, `test_runner.rs`): injects Established → FIN-WAIT-2 and Established → FIN-WAIT-1, feeds a data+FIN segment in order, and asserts `recv_next` advances past the data and the FIN, the text lands in `recv_buffer`, and the state reaches TIME-WAIT.

## Verification

Unprobed render of `https://en.wikipedia.org/wiki/Firefox` (`firefox-test-core,kdb`, smp 4). FF establishes connections to the exact CDN IPs that wedged in the capture (185/199/151 :443) and renders without the connection-teardown stall (0 RST vs 7 in the wedge). *(Render result appended in a follow-up comment once the PNG gate completes.)*

Spec anchors: RFC 9293 §3.10.7 / §3.10.7.4 (RST on unacknowledged data; FIN-WAIT receive processing), RFC 793 §3.5.


me1iissa:
## Render verification (unprobed, firefox-test-core,kdb, smp 4)

Target: `https://en.wikipedia.org/wiki/Firefox`

**The NDE-18 network-teardown wedge is eliminated.** FF establishes connections to the exact CDN IPs that wedged in the capture and tears them down cleanly:

| Metric | Wedge capture (before) | This render (after) |
|---|---|---|
| Connections to Fastly CDN (199.232.57.91 / 151.101.1.91 :443) | established then **RST after 47–63× retransmit** | **6 established, 0 RST** |
| `RST: closing` total | 7 (one per wedged data+FIN flow) | **0** |
| Clean graceful closes (`LastAck → Closed`) | — | **5** |
| Syscall progress at the resource-load phase | frozen | climbed to **sc ≈ 309 k** |

FF then advanced through all resource fetches to `[GATE] screenshot-actors` (the IPDL screenshot handshake) and ran to `[GATE] ff-exit-clean code=0`.

**The PNG itself was not produced** (`out.png size=0`) — but the blocker is no longer the network teardown. FF wedges at the **separately-tracked, parked screenshot-IPC futex livelock** (a `FUTEX_WAKE_EXIT` storm on the content pid with `key_present=false`, the W101/GATE-B saga where the futex is proven faithful, PR #489). That gate is out of scope for this TCP fix and is a parked investigation — not reopened here.

Net: this PR removes one of the load-blocking divergences (heavy real-site connections that never reached a terminal state on our kernel), surfacing the next, already-known gate.

me1iissa:
Superseded: in-order data + FIN delivery in FIN-WAIT-1/FIN-WAIT-2 landed via #627 (6e56c7c3). Closing.

PR #562, opened 2026-06-16
Headless real-website rendering: 9 OUR-kernel fixes + net hardening + reliable PNG extraction

me1iissa:
## Summary

Real-world websites now render under upstream **headless Firefox** on AstryxOS. This branch stacks the
session's kernel fixes + extraction infrastructure on the proven render base (`ff/real-website` @ #560),
and is the state in which **Hacker News, info.cern.ch, BBC News, example.com, Wikipedia, CNN Lite, and the
full `www.cnn.com` front page** all render to valid PNGs (captured + posted from live boots).

Every fix below is an **OUR-kernel divergence** surfaced by driving real heavy sites — found by
discriminator/autopsy, fixed against the public spec, regression-tested. None is an upstream-gecko change.

## Kernel fixes (CNN/heavy-site-surfaced)

| Area | Fix | Spec |
|---|---|---|
| mm | file-backed `MAP_SHARED`/memfd mappings now hold a counted inode reference (close-while-mapped no longer frees a live inode → SIGSEGV) | POSIX mmap(2)/memfd_create(2) |
| net/udp | bound the per-port receive queue at `SO_RCVBUF` (tail-drop) + O(1) `VecDeque` drain — HTTP/3 QUIC floods no longer OOM/livelock | RFC 768, SO_RCVBUF |
| net/unix | SCM_RIGHTS fd-batches incarnation-guarded against slot recycle **and** a sendmsg's data+fds committed atomically (no IPDL byte/fd desync) | unix(7), recvmsg(2) |
| mm/tlb | self-service own shootdown slot in the ACK-spin to break a cross-CPU TLB-shootdown **deadlock** under IF=0 + tighter ACK bound | Intel SDM Vol. 3A §4.10 / §6.12 |
| mm/heap | size the kernel heap for heavy-page surfaces (ramfs/memfd-backed) — large image surfaces no longer exhaust it | — |
| net/tcp | deliver in-order data+FIN in FIN-WAIT-1/2 (CDN close_notify+FIN) | RFC 9293 §3.10.7.4 |

## Network hardening (audit-confirmed sibling defects)

| Fix | Spec / CVE |
|---|---|
| TCP half-open SYN-flood reaper + per-port backlog cap | RFC 4987 |
| TCP out-of-order reassembly → O(log n) BTreeMap + entry-count cap + coalesce | RFC 9293 §3.10, CVE-2018-5390 (SegmentSmack) |
| TCP `recv_buffer` bound at `SO_RCVBUF` + window shrink (slow-reader OOM) | RFC 9293 §3.8 |
| DNS name-decompression hop cap (cyclic-pointer infinite loop) | RFC 1035 §4.1.4, CWE-835 |

## Extraction / e2e infrastructure

- **Gate the slow/lossy serial B64 PNG emit off by default**; the host extracts `/tmp/out.png` over the
  live kdb/QGA channel instead (kdb read-file: snapshot across chunks + 128 KiB chunks).
- **120 s post-render grace window** in firefox-test mode keeps the VM kdb/QGA-serviceable after the render
  so the host copies the PNG out reliably (lossless, ~seconds) before shutdown — fixing a ~16-min lossy path.
- Snapshot height cap (`--window-size 1366,8000`) bounds the surface for very tall pages.

## Exploratory (additive, off by default)

- `--ff-gui` runtime toggle to launch Firefox non-headless into the native X11 WM (default remains headless).

## Verification

- Per-fix in-kernel regression tests pass (`test-mode`).
- Live headless renders of the sites above produce valid PNGs; full `www.cnn.com` = 1366×8000, 7.6 MB.

## Notes

- Base is `ff/real-website` (#560); the `ff/real-website → master` graft is a separate, coordinator-owned step
  (post-history-purge graph divergence; master is ~17 PRs behind).
- A driver-level virtio-blk completion-stall fix (the residual headless-render flakiness + the GUI-paint
  blocker) is in flight separately.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -49,6 +49,21 @@ firefox-test-core = ["heap-canary"]
 # `dump_futex_wait_stack` rbp-chain walker) — trace is a strict superset of the
 # functional core, never a standalone.
 firefox-test-trace = ["firefox-test-core"]
+# `ff-png-serial-emit` — opt-in for the FULL [FF-OUT-PNG-B64:…] base64 chunk
+# stream that re-encodes Firefox's rendered /tmp/out.png over COM1.  OFF by
+# default because that stream is one synchronous `serial_println!` per 57 input
+# bytes (RFC 2045 §6.8 MIME line), i.e. ~36 800 UART writes for a 2 MB PNG; at
+# ~88 µs per port-I/O VM-exit (Intel SDM Vol. 3C §25 I/O-instruction VM-exits)
+# that is several minutes of CPU0 emit time, and while it runs it starves the
+# kdb pump thread that `qemu-harness.py kdb-read-png` reads through.  With this
+# feature OFF, `ff_out_png::emit_out_png()` still emits the single-line
+# `[FF-OUT-PNG:path=… size=… sig_ok=… complete=…]` summary marker (so the
+# harness knows the PNG is written and complete) but skips the chunk flood; the
+# PNG stays in the guest VFS for the non-baud-bound `kdb-read-png` live read.
+# Enable only when serial is the ONLY extraction channel (no `kdb` feature) or
+# to reproduce the historical serial-stream wire format.  Implies
+# `firefox-test-core` because the emit lives on the firefox-test boot path.
+ff-png-serial-emit = ["firefox-test-core"]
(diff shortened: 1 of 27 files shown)


me1iissa:
Obsolete umbrella branch: the nine render-base fixes and the PNG extraction path all landed on master individually. Closing.

PR #567, opened 2026-06-16
virtio-blk: bounce userspace buffers for read(2) DMA + race-free quarantine (fixes the read-storm guest wedge)

me1iissa:
## Summary

A regular-file `read(2)` issued with a **userspace** buffer wedged the whole guest. The HTTP-server demo (GET `/` → parse → read docroot → send) hit `[VIRTIO-BLK] wait_completion timeout` for every slot, then `slot exhaustion (32/32) — resetting device`, and the cycle repeated until the process died. An in-guest `md5sum` of a file hung identically, while `wc -c` (stat only) returned instantly — isolating the failure to the **read(2) data path**, not mmap (headless-browser mmap demand-paging of hundreds of MiB was unaffected).

## Root cause (GDB-autopsy)

Caught mid-storm with the GDB stub + `qmp-xp` over the live virtqueue:

- **`avail.idx == used.idx == driver-cursor` at every sample** — the device retired *every* request it was given. It was never sick.
- The stuck request's descriptor pointed its **data buffer at a userspace virtual address** (`0x00007eff_…`). ext2's aligned fast path reads whole blocks straight into the caller's buffer with no bounce copy; for `read(2)` that buffer is the user's. `submit_request` derives the DMA address as `virt < PHYS_OFFSET ? virt : virt - PHYS_OFFSET`, so a userspace VA was handed to the device verbatim as a "physical" address far beyond installed RAM. With no IOMMU the device cannot service such a chain (VIRTIO 1.2 §2.7: buffers are addressed physically) — it never lands in the used ring, the waiter trips its no-progress deadline, the slot is quarantined.
- **`QUARANTINED_TIMEOUTS` climbed while `RECLAIMED_QUARANTINES` stayed at 0** — every quarantine leaked, because the quarantine flag was set *after* the final drain/`done` re-check, so a peer walker could consume the slot's used-ring entry in the gap and orphan the quarantine permanently.

The mmap path was unaffected only because it reads into a kernel-physical page-cache frame.

## Fix (three coordinated parts)

1. **Bounce non-DMA-able buffers** (primary). `do_io` routes any buffer that is neither kernel-higher-half nor genuine low-physical RAM (a userspace VA) through a kernel-heap staging buffer. The device only ever sees a real physical descriptor address. Applied at the driver chokepoint, so it covers every block-backed FS (ext2/fat32/ntfs) and every caller.
2. **Race-free quarantine** (`finalize_or_quarantine`). Set `quarantined = true` *before* the final used-ring drain so any concurrent consumer takes `drain_used_ring`'s atomic reclaim branch; re-test `done` and unwind the speculative quarantine if the completion already landed. Closes the `RECLAIMED=0` leak. (VIRTIO 1.2 §2.7.13)
3. **Deschedule-aware no-progress deadline.** A waiter yielded into a deep run queue may not run for seconds while the device drains silently; on wakeup `get_ticks()` has jumped past the deadline though no *observed* device silence occurred. Both wait loops re-arm across a large inter-iteration tick gap, bounding the deadline to observed silence. A genuinely dead device still fails fast (POSIX read(2): I/O error → -1, never short-count-as-EOF).

## Tests

Adds **Test 0e4** (test-mode): the DMA/bounce classifier flags the exact userspace-VA class the autopsy caught, and a multi-block device read completes coherently with **zero spurious no-progress timeouts**.

## Verification

- HTTP-server demo: **10/10 HTTP 200**, `Content-Length: 10489` (the full file), **zero** `wait_completion timeout` / `slot exhaustion` lines. Reproduced on the committed fix (clean cherry-pick verify build).
- Test-mode suite: **Tests 0e2 / 0e3 / 0e4 all PASS**; 219 PASS. The 4 unrelated FAILs (PTY master-read, TCC-missing-artifact, thread-reaper test518b, timer monotonic-rate) pre-date this change and touch no code in this diff.

## Notes

- The HTTP transfer length occasionally varies (1236 vs 1684 vs full) — that is WebXO's own shared-object worker-pool race (the demo runs `--nthreads=1` to mitigate), **not** a kernel issue: the kernel reads the full file and WebXO reports the correct `Content-Length: 10489`.

### DIFF
--- a/kernel/src/drivers/virtio_blk.rs
+++ b/kernel/src/drivers/virtio_blk.rs
@@ -138,6 +138,36 @@ fn phys_to_virt<T>(phys: u64) -> *mut T {
     (PHYS_OFFSET + phys) as *mut T
 }
 
+/// Upper bound on a directly-DMA-able identity-low buffer, in bytes.
+///
+/// A buffer below `PHYS_OFFSET` is normally identity-mapped low physical memory
+/// (the boot stack / early-init scratch), for which `phys == virt`.  But a
+/// *userspace* virtual address is ALSO below `PHYS_OFFSET` (the user half lives
+/// at e.g. `0x0000_7eff_…`), and it is **not** a physical address — the device
+/// has no IOMMU here, so a descriptor pointing at a userspace VA tells the device
+/// to DMA to a bogus "physical" address far beyond installed RAM, which it can
+/// never service (the chain is never retired → the waiter times out → the
+/// slot-exhaustion cascade).  Any genuine identity-low physical buffer is within
+/// installed RAM, which is at most a few GiB; a userspace VA is orders of
+/// magnitude higher.  We treat a low address at or below this ceiling as a real
+/// physical buffer and anything between it and `PHYS_OFFSET` as a non-DMA-able
+/// userspace VA that must be bounced.  4 GiB is comfortably above any guest RAM
+/// size this kernel targets while still excluding the user VA range.
+const IDENTITY_LOW_DMA_CEILING: u64 = 4 * 1024 * 1024 * 1024;
+
+/// Is `buf` a buffer the block device can DMA into/out of directly?
+///
+/// True for kernel higher-half buffers (heap, thread stacks — `phys = virt -
+/// PHYS_OFFSET`) and genuine identity-low physical buffers (`phys = virt`, below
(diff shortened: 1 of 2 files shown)


me1iissa:
Superseded: the virtio-blk user-buffer read(2) DMA fix landed via #568 (881a9b02). Closing.

PR #573, opened 2026-06-17
gui: default to a local file:// URL when non-local connections are disabled

me1iissa:
GUI-mode Firefox pairs `MOZ_DISABLE_NONLOCAL_CONNECTIONS=1` (required so the e10s-disable switch is honoured on the official ESR build) with `DEFAULT_FF_URL = https://1.1.1.1/` — mutually exclusive: necko rejects the non-local load before any SYN, so the demo page never loads. When GUI mode is active with no explicit `astryx.ff_url=`, default the target to `file:///tmp/hello.html` (the page the kernel already stages to the ramdisk at boot), which loads with no network. +19 lines, kernel/src/main.rs. Builds green; confirmed at runtime: `[FFTEST] target URL (GUI compiled default): file:///tmp/hello.html`.

### DIFF
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -1347,10 +1347,29 @@ pub unsafe extern "C" fn _start(boot_info: *const BootInfo) -> ! {
             // compiled default.  Announce which source won so a forensic reader
             // of a render boot can tell at a glance whether the harness's
             // `--ff-url` actually took effect.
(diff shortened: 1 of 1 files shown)


me1iissa:
Superseded: GUI-mode Firefox now defaults to file:///tmp/hello.html on master (3997c20a). Closing.

PR #574, opened 2026-06-17
data-disk: stage gschemas.compiled + mime.cache + built-in pixbuf loaders for GUI Firefox

me1iissa:
## What

Fixes the data-image **staging gaps** that broke windowed (`--ff-gui`) Firefox
at GTK startup. The substantive change is the final commit
(`7112b71f`); the preceding GUI-lineage commits are the repro base this work
builds on (the `--ff-gui` feature, e10s in-process collapse, X11/XKB fixes)
and the `file:///tmp/hello.html` default so a page loads.

## Root cause (corrected)

The windowed path brings up GTK/GDK, which the headless screenshot path never
exercises. Three runtime **index files** that an Alpine apk post-install
trigger normally generates are absent on the data disk, because the image is
built by extracting apks without firing their triggers (same class as the ICU
data-file gap). The earlier hypothesis — "the staged `loaders.cache` omits
png/jpeg, so `GDK_PIXBUF_MODULE_FILE` suppresses the built-in PNG loader" — was
**verified false** at the library level: this build links PNG/JPEG into
`libgdk_pixbuf` and `gdk_pixbuf_io_init_builtin()` runs unconditionally, so
`gdk-pixbuf-pixdata`/`-thumbnailer` decode a PNG correctly with an empty,
missing, or png-less cache alike. The real gaps are the missing compiled
caches:

| Missing file | Effect on `--ff-gui` | Fix |
|---|---|---|
| `gschemas.compiled` | GtkSettings cannot resolve `gtk-theme-name`/`gtk-icon-theme-name`; "Settings schema not installed" warnings + inconsistent theme/icon fallback | `glib-compile-schemas` under qemu-user |
| `mime.cache` | GIO/GTK content-type queries warn | `update-mime-database` under qemu-user |
| `loaders.cache` png/jpeg stanzas | cache advertises a format set that disagrees with what the library decodes | append name-matched built-in stanzas (defense-in-depth) |

The on-disk Adwaita **theme** is intentionally not staged: GTK 3.24 ships
Adwaita as a compiled-in GResource (`/org/gtk/libgtk/theme/Adwaita/...`), and
the crashing titlebar symbolic icons live in the libgtk GResource too.

## Harness

Extends the existing `-vga vmware` auto-injection (was xeyes-test only) to
`--ff-gui`, so the in-kernel Xastryx compositor framebuffer is screendump-able
via QMP. Additive — no field renames.

## Verification

Rebuilt the musl data image: `gschemas.compiled` (2732 B), `mime.cache`
(157560 B), and the png/jpeg stanzas all present in `build/data.img`. A
`firefox-test-core,kdb --ff-gui` boot runs clean through the **entire** serial
log — zero `GDK_IS_PIXBUF` / `Gdk-CRITICAL` / `SIGSEGV` / pixbuf relaunch storm
/ "schema not installed" signatures (vs the prior 5/5 crash at realization).
Firefox survives GTK realization, connects pid1 + pid3 to the X11 server, and
pumps a live event loop.

## Known residual (out of scope — kernel/X11, not staging)

Firefox advances to the 300×200 GTK probe window but does not yet map the real
toplevel or paint (`kdb x11-present` PutImage = 0); the framebuffer shows the
AstryxOS desktop, not Firefox chrome. The event loop is alive (sc climbing,
all threads cycling), so this is the kernel render / X11 event-delivery gate
the headless path also hits — a separate subsystem owner's concern, not a
data-image staging gap.

## Citations

gdk-pixbuf-query-loaders(1) (cache format), ISO/IEC 15948 (PNG signature),
JFIF (JPEG SOI marker), the freedesktop.org Shared MIME-info Database
specification, GSettings (schema cache).

### DIFF
--- a/kernel/src/drivers/virtio_blk.rs
+++ b/kernel/src/drivers/virtio_blk.rs
@@ -138,6 +138,36 @@ fn phys_to_virt<T>(phys: u64) -> *mut T {
     (PHYS_OFFSET + phys) as *mut T
 }
 
+/// Upper bound on a directly-DMA-able identity-low buffer, in bytes.
+///
+/// A buffer below `PHYS_OFFSET` is normally identity-mapped low physical memory
+/// (the boot stack / early-init scratch), for which `phys == virt`.  But a
+/// *userspace* virtual address is ALSO below `PHYS_OFFSET` (the user half lives
+/// at e.g. `0x0000_7eff_…`), and it is **not** a physical address — the device
+/// has no IOMMU here, so a descriptor pointing at a userspace VA tells the device
+/// to DMA to a bogus "physical" address far beyond installed RAM, which it can
+/// never service (the chain is never retired → the waiter times out → the
+/// slot-exhaustion cascade).  Any genuine identity-low physical buffer is within
+/// installed RAM, which is at most a few GiB; a userspace VA is orders of
+/// magnitude higher.  We treat a low address at or below this ceiling as a real
+/// physical buffer and anything between it and `PHYS_OFFSET` as a non-DMA-able
+/// userspace VA that must be bounced.  4 GiB is comfortably above any guest RAM
+/// size this kernel targets while still excluding the user VA range.
+const IDENTITY_LOW_DMA_CEILING: u64 = 4 * 1024 * 1024 * 1024;
+
+/// Is `buf` a buffer the block device can DMA into/out of directly?
+///
+/// True for kernel higher-half buffers (heap, thread stacks — `phys = virt -
+/// PHYS_OFFSET`) and genuine identity-low physical buffers (`phys = virt`, below
+/// [`IDENTITY_LOW_DMA_CEILING`]).  False for userspace virtual addresses, which
+/// must be bounced through a kernel buffer before the device sees them.  Mirrors
+/// the `data_phys` derivation in [`submit_request`].
+#[inline]
+fn is_dma_capable(buf: *const u8) -> bool {
+    let v = buf as u64;
+    v >= PHYS_OFFSET || v < IDENTITY_LOW_DMA_CEILING
+}
+
 // ── Virtqueue Layout Helpers ────────────────────────────────────────────────
 
 /// Calculate the byte offset of the available ring within the virtqueue.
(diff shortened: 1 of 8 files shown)


me1iissa:
## CI note: the 3 unexpected `kernel smoke` failures are inherited from the GUI-lineage base, not from this change

`workflow_dispatch` run [27659858725](https://github.com/me1iissa/AstryxOS/actions/runs/27659858725): cargo check ✓, bootloader+kernel builds ✓, tooling smoke ✓, **qemu-harness smoke (tier 1) ✓**. `kernel smoke` reports 3 unexpected `[FAIL]`s:

- `x11_mitshm — MIT-SHM not present (present=0)`
- `x11_xkb — XkbUseExtension supported=0`
- `275/master-read — master read returned 7 bytes`

These are **not** caused by the staging change. The two staging commits touch only `kernel/src/main.rs` (the 19-line `file://` default), `scripts/install-firefox-musl.sh`, and `scripts/qemu-harness.py` — **zero** kernel x11/pty/test-runner code:

```
$ git diff b3577ffc..HEAD --name-only
kernel/src/main.rs
scripts/install-firefox-musl.sh
scripts/qemu-harness.py
```

They are inherited from the GUI-lineage commits this branch is built on:

- `x11_mitshm` is the **intended** consequence of `cf80e12e` ("x11: stop advertising MIT-SHM … so clients use core XPutImage") — the `x11_mitshm` test now asserts behaviour the lineage deliberately removed and is stale (qa-engineer follow-up to retarget the assertion).
- `x11_xkb` follows the XKB minor-opcode routing change in `b3755b1a` / `24c0a153`.
- `275/master-read` is a PTY echo-timing artifact from the lineage's terminal changes.

CI `build.yml` only auto-runs on PRs targeting `master`; this PR targets `lan/ssh-usable`, so the run above was triggered manually. The staging change itself is data-image only and was verified by a live `firefox-test-core,kdb --ff-gui` boot (clean serial log, crash gone).

me1iissa:
Closing: this branch was pushed off a pre-purge lineage (131 commits diverging from master, merge-base pre-2026-06-10-purge). The valuable staging fix (7112b71f: gschemas.compiled + mime.cache + built-in pixbuf loaders) will be re-grafted onto a clean base. No corpus-path leaks in the pushed content (scanned), but old-history branches must not live on origin.

PR #577, opened 2026-06-17
ipc/pipe: block the writer on a full pipe (POSIX write(2)) — fixes windowed-Firefox writev busy-loop

me1iissa:
## Summary

Windowed Firefox (the `--ff-gui` path) boots clean but **never paints** because PID 1's main thread enters an unbounded `writev(fd=2)` busy-loop (~2.8M writev/tick; the page-fault, disk and mmap counters freeze; the boot is auto-reaped as a livelock at 50M syscalls).

Root cause: the pipe arm of `sys_write_linux` called `pipe_write` once and returned whatever fit. On a **full** pipe that is a no-progress `0` return, which violates POSIX `write(2)` for a blocking fd. Firefox redirects stdout+stderr to a 4 KiB pipe; once its read end stops draining, the main thread's `writev` to that pipe returns `0` and musl stdio / NSPR retry the identical write forever. Because the main thread never returns to its glib main-context `poll()`, the in-kernel X11 server (pumped only from `poll`/`select`) never drains the client socket, so no frame is ever presented — the windowed first-paint gate.

## The fix

`pipe_write_blocking` implements the `write(2)` / `pipe(7)` contract:
- **Atomicity** (`count <= PIPE_BUF`): all-or-block, never a short write.
- **Large writes** (`count > PIPE_BUF`): may split, blocking for space between chunks.
- **O_NONBLOCK**: `-EAGAIN` (not a no-op `0`) when no progress is possible.
- **No reader**: `-EPIPE`.

Blocking reuses the existing per-pipe writer wait list (`wait_writable` / `wake_writers_all` / `waiter_cleanup_writer`) — the same check-then-park primitive the reader path already uses. No pipe lock is held across `schedule()`; the under-lock re-check in `wait_writable` closes the TOCTOU window against a racing reader drain (no lost wakeup, no hold-across-dispatch deadlock).

Adds `pipe::pipe_reader_closed` / `pipe_space` accessors, a `Pipe::reader_closed` predicate, and a public `PIPE_BUF` constant.

## Verification

- **Windowed Firefox before**: PID 1 `file=` syscall counter climbs past 50M (auto-reaped livelock), `cur_nr=20` (writev) pinned, X11 server drain frozen, no toplevel map.
- **Windowed Firefox after**: the writev busy-loop is **gone** — `file=` freezes (e.g. 9462), the main thread parks correctly on the full pipe instead of spinning, the whole process tree goes quiescent (no spinner).
- **Regression test (Test 0a2)**: O_NONBLOCK full pipe → `-EAGAIN`; O_NONBLOCK `<=PIPE_BUF` with insufficient space → `-EAGAIN` (atomicity); blocking write with no reader → `-EPIPE`. Passes; full suite proceeds with no new failures.

SIGPIPE on the no-reader path is a separate, pre-existing gap (`man 2 write`) and out of scope.

Refs: `write(2)`, `pipe(7)`, POSIX.1-2017 §write.

### DIFF
--- a/kernel/src/gui/terminal.rs
+++ b/kernel/src/gui/terminal.rs
@@ -1336,7 +1336,11 @@ pub fn poll_output() {
     // First pass: read all available data before locking TERMINAL
     let mut chunks: alloc::vec::Vec<alloc::vec::Vec<u8>> = alloc::vec::Vec::new();
     loop {
-        let n = crate::ipc::pipe::pipe_read(pipe_id, &mut raw).unwrap_or(0);
+        // `pipe_read_and_wake` wakes any writer (e.g. the Firefox stdout/stderr
+        // producer) parked for buffer space once this drain frees it.  Without
+        // the wake the producer parks forever and never returns to its event
+        // loop — the actual deadlock this consumer must avoid (`man 7 pipe`).
+        let n = crate::ipc::pipe::pipe_read_and_wake(pipe_id, &mut raw).unwrap_or(0);
         if n == 0 { break; }
         chunks.push(raw[..n].to_vec());
         any_data = true;
@@ -1365,7 +1369,7 @@ pub fn poll_output() {
         // Drain any final bytes the child wrote before exiting.
         drop(guard); // release TERMINAL before pipe_read
         let mut tail = [0u8; 4096];
-        let tn = crate::ipc::pipe::pipe_read(pipe_id, &mut tail).unwrap_or(0);
+        let tn = crate::ipc::pipe::pipe_read_and_wake(pipe_id, &mut tail).unwrap_or(0);
         crate::ipc::pipe::pipe_close_reader(pipe_id);
 
         let mut guard2 = TERMINAL.lock();
--- a/kernel/src/ipc/pipe.rs
+++ b/kernel/src/ipc/pipe.rs
@@ -109,6 +109,12 @@ impl Pipe {
         self.writers == 0
     }
 
+    /// Check if every read end is closed.  A blocking `write(2)` on such a
+    /// pipe must fail with `EPIPE` rather than park forever (man 2 write).
+    pub fn reader_closed(&self) -> bool {
+        self.readers == 0
+    }
+
     /// Available bytes (count of unread data).
     pub fn available(&self) -> usize {
         self.count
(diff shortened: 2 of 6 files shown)


me1iissa:
## Lost-wakeup deadlock fixed + end-to-end verified

The independent review was correct: the prior commit parked blocking-pipe writers via `wait_writable` + `schedule()` but **no drain path ever woke them** — the only callers of `wake_writers_all` were the fd-close paths. A parked writer was released only on fd close, never when a reader actually freed space, converting a transient full-pipe stall into a permanent deadlock. The "after" metric (main thread parked at 9462 syscalls) **was** that deadlock.

### Wake mechanism wired
New canonical drain entry point `pipe::pipe_read_and_wake(pipe_id, buf)` — drains via `pipe_read` then calls `wake_writers_all(pipe_id)` on a successful (`n>0`) read. `pipe_read` takes and drops `PIPE_TABLE` before returning, so the wake (`PIPE_WRITE_WAITERS -> THREAD_TABLE`) never nests under `PIPE_TABLE`; documented lock order preserved.

Routed all real drain consumers through it:
- `sys_read_linux` — the `read(2)` pipe branch
- native Aether read path (`subsys/aether/syscall.rs`)
- **`gui::terminal::poll_output`** — the in-kernel terminal that drains a child's stdout/stderr (the actual Firefox producer whose deadlock this prevents), including its final-tail drain

### Doc-comment fix
The false claim on `pipe_write_blocking` that "the reader's drain in `sys_read_linux` already calls `wake_writers_all`" is replaced with an accurate description of the wired mechanism.

### Regression test (Test 0a2, new Sub-test 3)
Single-core deterministic — no second CPU. Enqueues a simulated parked writer, confirms a raw `pipe_read` frees space WITHOUT waking it (reproduces the bug), then confirms `pipe_read_and_wake` drains AND drops the writer-waiter count to 0 (the fix). Suite: **197/202** pass (the 5 fails are pre-existing data.img-staging / monotonic-rate gaps, unrelated to pipes).

### End-to-end: windowed Firefox — deadlock RESOLVED
`--features firefox-test-core,kdb --ff-gui`. The pipe-park no longer deadlocks. FF (pid 32) progressed with **balanced, healthy** metrics rather than the frozen unbounded-`file=` deadlock signature:

```
sc=160   vm=58   file=79   sync=2  proc=7   sig=5
sc=1103  vm=456  file=617  sync=4  proc=7   sig=7
sc=1974  vm=924  file=784  sync=29 proc=78  sig=63
sc=3996  vm=2549 file=988  sync=58 proc=127 sig=89   pf=6422  disk=R315MB
```

`file=` grows modestly and proportionally to everything else (vm/proc/sig/pf all climbing, thread-pool clone children TID 50-54 spawning) — the main thread is **cycling, not parked forever**. This is the deadlock resolving.

### Named next gate (NOT the pipe deadlock, NOT introduced here)
FF then hit a separate **`KERNEL_GPF` bugcheck** during clone-thread spawn: `error_code=0xd`, garbage `RIP=0x7012`, `CR2=0x7efe8130db38` (the TLS base of the just-spawned clone child TID 54), `tid=0 pid=0` (fires during a context-switch/IPI). This is a clone/TLS / cross-CPU class fault on the SMP shared-CR3 path with no pipe involvement — consistent with the known missing cross-CPU TLB-shootdown class, orthogonal to this PR. No paint reached (FF crashed before mapping a toplevel), so I am **not** claiming a paint that isn't there.

Branch rebased onto current `origin/master` (PR #576 had merged the GUI lineage, making 7 of the prior commits duplicates — dropped by the rebase); `git merge-base --is-ancestor origin/master HEAD` passes. Force-with-lease pushed.

PR #580, opened 2026-06-17
x11: rasterise core geometry ops + window background-pixmap (visible client drawing)

me1iissa:
## Summary

Stacked on #579 (spec-correct map sequence). Makes the X server actually *paint*
client drawing instead of silently discarding it — the prerequisite for any
core-protocol-drawing client (xeyes, xclock, Athena/Xaw widgets) to show pixels.

Three gaps were closing the visible-paint path:

### 1. Core geometry ops were silent no-ops
`PolyFillArc` (71), `PolyArc` (68), `PolySegment` (70), `PolyLine` (65),
`PolyPoint` (64) and `PolyRectangle` (67) were `=> {}` stubs in the request
dispatch. A client that draws ellipses/lines via the core protocol issued the
requests and got a spec-correct event stream, yet nothing appeared because the
draws were discarded. Added an integer-only software rasteriser
(`resource::raster`) operating on the shared BGRA pixel buffer used by both
`WindowData` and `PixmapData`, and wired the six ops to it. `fill_arc` fills the
ellipse bounded by the arc rectangle (full-ellipse scanline spans, or a
per-pixel angular-sweep test for partial arcs, via a compile-time integer atan
table); `stroke_arc` strokes the ellipse boundary; `line` is Bresenham. No
floating point (kernel target is `+soft-float`).

### 2. Window `background-pixmap` was parsed-then-discarded
`CreateWindow`/`ChangeWindowAttributes` skipped the `background-pixmap`
attribute, and `ClearArea`/`MapWindow` only ever filled a solid
`background-pixel`. Clients that render static content into an off-screen pixmap
and set it as the window background (relying on the server to paint it on
expose) showed nothing. Track `background_pixmap` in `WindowData` and add
`paint_window_background()`, which tiles the pixmap (origin at the window
top-left, per the core protocol) into `w.pixels` on `ClearArea` and on
`MapWindow`. Solid-background windows are unchanged.

### 3. xeyes soak loop parked the BSP under KVM
The `xeyes-test` soak loop issued `hlt` at its tail; under KVM the BSP's LAPIC
timer is not reliably re-delivered after `hlt`, parking CPU0 and freezing
`compose()` (so no frame reached the screen). Busy-spin like the firefox-test
loop instead, and keep compositing after the client parks (with a bounded soak
budget) so a screendump can capture the painted frame.

## Result

With these fixes the X server composites client windows and rasterises their
core-protocol draws into the framebuffer (verified: an xeyes window now
composites to screen where previously nothing appeared). A residual, separate
**libXt/libXaw** gate remains for xeyes specifically (its Eyes widget completes
realization, receives its Expose, then parks in `poll(timeout=-1)` with no
animation timer armed and issues zero *window* draw ops — it only ever draws
its Shape mask into a pixmap). That gate is downstream of (now-verified-faithful)
event delivery + poll wakeup and is tracked separately.

## Spec references
X Window System core protocol: PolyFillArc, PolyArc, PolySegment, PolyLine,
PolyPoint, PolyRectangle, CreateWindow `background-pixmap`, ClearArea, window
background painting on map.

### DIFF
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -592,8 +592,8 @@ pub unsafe extern "C" fn _start(boot_info: *const BootInfo) -> ! {
                 }
 
                 // xeyes is a draw-and-poll event loop — under the demo soak we
-                // exit when the workload itself exits, OR after 18000 ticks
-                // (~180 s) so a wedged process can't pin the CI watchdog.
+                // keep compositing until the bounded budget below so a wedged
+                // process can't pin the CI watchdog.
                 // Note whether xeyes has terminated, but do NOT break: keep the
                 // BSP composing so the eyes the client painted into its window
                 // remain on screen and a QMP screendump can capture them.  The
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -594,19 +594,37 @@ pub unsafe extern "C" fn _start(boot_info: *const BootInfo) -> ! {
                 // xeyes is a draw-and-poll event loop — under the demo soak we
                 // exit when the workload itself exits, OR after 18000 ticks
                 // (~180 s) so a wedged process can't pin the CI watchdog.
-                if elapsed > 60 && !crate::gui::terminal::is_firefox_running() {
-                    serial_println!("[XEYES] xeyes exited after {} ticks", elapsed);
+                // Note whether xeyes has terminated, but do NOT break: keep the
+                // BSP composing so the eyes the client painted into its window
+                // remain on screen and a QMP screendump can capture them.  The
+                // window's `pixels` (and thus the composited frame) persist after
+                // the client exits because the X resources are reaped lazily.
+                if elapsed > 60 && !xeyes_exited && !crate::gui::terminal::is_firefox_running() {
+                    serial_println!("[XEYES] xeyes exited after {} ticks (exec_pid={} exit={:?}) — continuing to composite",
+                        elapsed, crate::gui::terminal::exec_pid(),
+                        crate::gui::terminal::exec_exit_status());
                     xeyes_exited = true;
-                    break;
                 }
 
-                if elapsed >= 18000 {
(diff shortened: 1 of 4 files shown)


me1iissa:
Added a regression test (`test_x11_poly_fill_arc`, Test 66-adjacent): a full-ellipse PolyFillArc over a 100×100 window must paint the centre with the GC foreground and leave a near-corner pixel (outside the inscribed ellipse) at the window background — distinguishing an ellipse fill from a rectangle fill or a no-op. PASSES.

Note on the broader test run: the other `[FAIL]`s observed locally (`Musl hello`, `dynamic_elf`, `TCC compile`, `sigchld`, `ascension` — all `/disk/bin/* NotFound`; `monotonic-rate`/`serial_fifo_irq` — KVM-timing) are pre-existing environment/data.img issues unrelated to this branch. `x11_render_d` (RENDER Composite → w.pixels) also fails but is pre-existing — this branch touches no RENDER/Composite code (verified `git diff` shows zero changes to op_render_*/window_composite_pixels). CI on a fully-staged image is authoritative.

me1iissa:
Addressed the review (APPROVE w/ non-blocking finding N1):

- **N1 (overflow hardening)**: `fill_arc` half-width `bw²·num` and `stroke_arc` inside-test `bw²·bh²` now use i128 intermediates — a client sending PolyFillArc/PolyArc with CARD16 w/h up to 65535 (65535⁴ > i64::MAX) would have wrapped silently in release (plot() clamping already prevented any OOB write) and panicked an overflow-checks build. Results fit i64 after the divide, so output is unchanged. (commit 3cc6d82c)
- **N2/N3 (cosmetic)**: fixed the stale soak-loop tick-budget comment and the atan-table `i/256` doc mismatch.
- **N4 (perf nit)**: left as-is — negligible for the xeyes ≤2-arc profile; noted for a future high-item-count client.

Regression test re-run after the i128 change: still PASS. test-mode builds clean.

PR #591, opened 2026-06-18
mm/vfs: pin backing inode for live MAP_SHARED file mappings (fixes FF main-proc realize/map SIGSEGV)

me1iissa:
## Summary

Fixes the deterministic SIGSEGV that killed Firefox's main process at the GTK realize/map transition, before it could paint. With this fix the main process no longer crashes there and advances to map a real 1280x972 toplevel (it could not before).

## The bug

A `MAP_SHARED` file mapping is an independent reference on its backing inode (POSIX `mmap(2)`: the mapping adds a reference not removed by a later `close(2)`; `unlink(2)`/`shm_unlink` keep an unlinked file alive while any reference remains). AstryxOS pinned **nothing** for live `MAP_SHARED` file VMAs, so the POSIX shared-memory idiom — `memfd_create`/`shm_open` -> `ftruncate` -> `mmap(MAP_SHARED)` -> `close(fd)` (-> `shm_unlink`) — freed the inode at the `close` while the mapping was still live. The next demand fault then stat'd a dead inode (size 0), populated no page, and returned an unserviced fault that became a silent SIGSEGV.

This is the deterministic crash a cross-process IPC layer that maps a `/dev/shm` region `MAP_SHARED` then `close()`s the fd hits.

## The fix

**1. Pin balance (one pin per `MAP_SHARED` file VMA-list membership):**
- `insert_vma` (+1, the mmap install)
- `clone_for_fork` (+1, the child inherits each VMA)
- `remove_range` fully-contained (-1, munmap / `MAP_FIXED` replacement)
- `remove_range` hole-punch split (-1 original, +1 each surviving half)
- `free_process_memory` / `free_vm_space` (-1 each, address-space teardown)

`PINNED_INODES` is authoritative: `close(2)`'s unlink-on-last-close (C5) and the unpin orphan-check both decline to free an inode that still has a live `MAP_SHARED` mapping.

**2. Deadlock fix:** `unpin_inode` takes `PROCESS_TABLE`, but the unpin sites (`remove_range`) run inside the `mmap`/`munmap` `PROCESS_TABLE` critical section. Split unpin into a lock-free `PINNED_INODES` decrement (`unpin_inode_deferred`) that queues the orphan-check/free, and a reaper (`reap_pending_inodes`) the syscall layer runs after dropping the lock — mirroring the fd-close path's deferred inode free.

**3. Defensive SIGBUS:** a demand fault on a file-backed VMA whose backing inode no longer stats (dead/truncated mapping) now delivers SIGBUS (`BUS_ADRERR`) to a Ring-3 faulter instead of a silent zero page or an unserviced fault — making any residual lifetime bug loud and spec-correct per POSIX `mmap(2)`. `deliver_sigsegv_from_isr` is generalised into `deliver_fault_signal_from_isr`.

## Test

New regression **Test 119b**: `memfd` -> `MAP_SHARED` pin -> `close(fd)` -> the inode data must still read; tear down the mapping -> the orphaned inode is then freed. Also asserts `MAP_PRIVATE` does **not** pin. Passes.

Full test-mode suite: **205 pass**, 3 pre-existing environmental fails (missing `tcc`/`busybox` on the GUI test image; the known KVM HLT-idle timer-rate quirk) — none related to mm/vfs/mmap. No deadlock, no DEC-UNDERFLOW.

## Verification (GUI FF boot, `--ff-gui` against the prebuilt image)

- The PID 1 `/dev/shm` mmap SIGSEGV / `exit_group(11)` is **gone** — PID 1 stays alive and advances (sc 15k -> 3.3M).
- FF maps a **real toplevel** `MapWindow 0x400016 1280x972` (cross-checked against the X11 window list), not the 1x1 synthetic windows.
- No spurious SIGBUS from the defensive change during normal operation.

**Honesty note:** FF does **not** yet paint. A screendump shows the desktop background (30 distinct dark-blue gradient colors), not painted FF chrome. The window maps but issues no present ops — that is the separate, **parked downstream gate** (AstryxOS-specific GObject signal-registration failure killing the GTK frame-clock paint cycle), independent of and downstream from the SIGSEGV fixed here.

## Notes

- Based on `ff/gui-declutter-desktop` (PR #589) for a clean single-client screendump; merge after/with that.
- Refs: POSIX.1-2017 `mmap(2)`, `munmap(2)`, `unlink(2)`; `memfd_create(2)`, `shm_open(3)`; SIGBUS-on-truncated-mapping semantics.

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -1237,7 +1237,7 @@ extern "C" fn exception_handler(vector: u64, error_code: u64, frame: &mut Interr
 /// - Bit 1: Write (1 = write, 0 = read)
 /// - Bit 2: User (1 = user mode, 0 = kernel mode)
 /// - Bit 4: Instruction fetch
-fn handle_page_fault(faulting_addr: u64, error_code: u64, _frame: &mut InterruptFrame) -> bool {
+fn handle_page_fault(faulting_addr: u64, error_code: u64, frame: &mut InterruptFrame) -> bool {
     PAGE_FAULT_TOTAL.fetch_add(1, Ordering::Relaxed);
     // Per-process PF counter.  Lockless: takes neither THREAD_TABLE nor
     // PROCESS_TABLE; one bounds-check + one Acquire load + one Relaxed
@@ -1878,7 +1878,7 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, _frame: &mut Interrupt
                                     cached_phys,
                                     c_mount, c_inode, c_off,
                                     mount_idx, inode, file_page_offset,
-                                    _frame.rip,
+                                    frame.rip,
                                 );
                             }
                         }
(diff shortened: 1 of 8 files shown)


me1iissa:
## mprotect partial-overlap split blocker — fixed (ee629798)

The reviewer-identified refcount imbalance in `sys_mprotect`'s partial-overlap split is fixed. The split now mirrors `VmSpace::remove_range`'s hole-punch arithmetic for the inode pin:

- **Unpin the removed original** — `space.areas.remove(i)` is now followed by `vfs::vma_unpin_if_shared_file(removed.flags, &removed.backing)`. This is the deferred (PROCESS_TABLE-free) variant, correct because `sys_mprotect` runs under `PROCESS_TABLE`; the orphan free is queued to the reaper.
- **Pin each surviving piece** — every piece is `vfs::vma_pin_if_shared_file`'d before its direct `areas.insert` (the inserts bypass `insert_vma`, so the pins are placed inline). `vma_pin_if_shared_file` touches only `PINNED_INODES`, so it is lock-safe.
- **Reap after the lock drops** — `drop(procs)` then `vfs::reap_pending_inodes()` after the VMA-edit loop and before the lock-free PTE retag (the retag/shootdown use only the copied `cr3`). Same wiring as `sys_munmap`'s post-Phase-1 reap.

The result restores the invariant: exactly one inode pin per MAP_SHARED file VMA-list membership. Previously the pin count stayed 1 while 2–3 pieces mapped the inode, so the first piece's teardown freed the inode out from under the others (the UAF this PR fixes, which had resurfaced through the split path as the SIGBUS).

### Sibling audit
Swept every direct `areas.remove` / `areas.insert`:
- `VmSpace::remove_range` fully-covered + hole-punch arms — already balance pins.
- `remove_range` left/right **shrink** arms — keep the *same* VMA object (one membership), so no rebalance is needed/correct.
- `insert_vma` — pins.
- SysV-SHM `shmat`/`shmdt` — `VmBacking::Device` (no inode), correctly out of scope.

Only the mprotect split was unbalanced.

### Test
New **Test 119c** (`test_mprotect_split_inode_pin_balance`, alongside 119b) drives the **real** `sys_mprotect` (via a `sys_mprotect_test` shim + a test-only `inode_pin_count_pub` accessor): mmaps a MAP_SHARED memfd region, `mprotect`s the interior to force a **3-way split**, and asserts:
- single VMA → pin count 1
- after the 3-way split → 3 pieces, **pin count 3**, inode still alive (data reads correctly)
- after tearing the address space down → inode freed **exactly once** (no double-free, no leak)

Pre-fix this fails at the post-split inode-alive check. Verified the test fails against the unfixed split and passes with the fix.

### Verification
- **Test suite (`test-mode,kdb`)**: 205/205 in-suite tests pass, including the new 119c. The 4 remaining failures are all pre-existing/environmental (`dynamic_elf` process-exit, missing `/disk/bin/tcc` + `/disk/bin/busybox` build artifacts, and the known KVM HLT timer-delivery quirk for `monotonic-rate`) — none touch the mm/vfs paths.
- **`--ff-gui` sanity** (firefox-test-core,kdb against the GUI image): clean — **0** SIGBUS/deadlock/bugcheck markers, and FF still maps its real toplevel (`[X11] MapWindow 0x400016 1280x972`). The only exit is the pre-existing content-proc (pid 3) `.bss` thread-start gate, unrelated to this change.

Ready for re-review / merge.

PR #592, opened 2026-06-18
vfs/x11: fix two Firefox paint gates (deadvertise MIT-SHM + reopen /proc/self/fd memfd by inode)

me1iissa:
## Summary

Firefox now maps a real toplevel window but paints 0 ops because of two kernel-side divergences on the paint path. This PR fixes both, named by a golden-diff against an upstream paint sequence.

### Gate 1 — stop advertising MIT-SHM (`kernel/src/x11/mod.rs`)
Our X server advertises the MIT-SHM extension via `QueryExtension`/`ListExtensions` but stubs the shared-memory image transport. A client that believes SHM works routes its paint through `CreatePixmap(shmseg) → CopyArea(shm-pixmap → window)`, which hits the unsupported arm and is silently dropped. Per the **MIT-SHM protocol spec**, a client that gets `present=0` from `QueryExtension` falls back to core `XPutImage(72)`, which `op_put_image` composites correctly.

- `op_query_extension`: `"MIT-SHM"` now falls into the not-present arm (`present=0`).
- `op_list_extensions`: `"MIT-SHM"` removed from the advertised set.
- `op_shm` left in place (harmless — a conforming client never reaches it after `present=0`).

### Gate 2 — `/proc/self/fd/N` reopen of an unlinked memfd no longer ENOENTs (`kernel/src/vfs/mod.rs`)
`memfd_create(2)` returns an anonymous (unlinked) inode whose synthesised name is removed at creation. Mozilla's shared-memory layer re-opens that fd **read-only** via `/proc/self/fd/<N>` to obtain an independently-flagged handle on the same backing. Our `open()` followed the procfs magic symlink to the fd's stored path string and **re-resolved** it; the unlinked memfd has no resolvable name, so the re-resolve returned **ENOENT** → the content-process sandbox channel broke.

Per **proc(5)**, each `/proc/<pid>/fd/<N>` entry is a magic symlink to the *open file* — opening it must **DUP the underlying open file by its backing inode**, not re-resolve a pathname. `vfs::open` now intercepts `/proc/{self|<pid>}/fd/<N>` **before** path resolution and clones the target fd's backing `(inode, mount, type)` into a fresh fd with a new offset and the caller's requested access mode (so a read-only reopen of an `O_RDWR` memfd is honoured). General to any fd type; the unlinked-memfd case is the one that must stop ENOENT'ing. Independent of the #591 MAP_SHARED inode-pin **lifetime** fix — this is a separate **reopen-by-path** bug.

## Tests (`kernel/src/test_runner.rs`)
- **Test 134** rewritten: `QueryExtension(MIT-SHM) → present=0` **and** MIT-SHM absent from `ListExtensions` (regression for the deadvertise).
- **Test 295** (new): `open(/proc/self/fd/N)` of an unlinked memfd succeeds, the new fd reads the memfd contents from a fresh offset, and a read-only reopen records `O_RDONLY`. Placed beside the inode-pin tests (119b/119c) so it runs ahead of the heavy Firefox-launch oracle.

Both pass in `test-mode`:
```
QueryExtension(MIT-SHM): present=0 ✓   ListExtensions: MIT-SHM absent ✓
open(/proc/self/fd/0) → fd 1 ✓ (no ENOENT)   reads payload from fresh offset ✓   recorded O_RDONLY ✓
```

## `--ff-gui` verification (honest)
Booted `--features firefox-test-core,kdb --ff-gui` on the GUI data image:
- ✅ **Gate 2 confirmed fixed**: 0 occurrences of "read-only dup failed / not using memfd". FF advances far past the old gate — libxul loads, content processes spawn, GTK initialises.
- ✅ **Gate 1 confirmed**: MIT-SHM no longer advertised; Test 134 green.
- ⚠️ **FF does NOT paint yet** — a **separate, deeper gate** remains. The content process (pid 3) takes a userspace **SIGSEGV** during libxul early init (`#PF: CR2=0xd0`, NULL+0xd0 read, faulting RIP in libxul's file-backed mapping, with a poisoned pointer `0x9ff49ad74a5a0015` threaded through registers + stack). This is the W215-family CLONE_VM store-visibility / aliasing crash, **outside filesystem scope** (kernel mm). The toplevel (1280×972) never maps; no `PutImage`/`CopyArea` fires. Screendump shows the desktop background only — **not claiming paint**.

This PR removes two genuine paint blockers and moves the failure point from a path-resolution ENOENT to a deeper mm/aliasing SIGSEGV; the remaining gate is for the kernel-mm owner.

Refs: MIT-SHM protocol spec; proc(5); memfd_create(2); POSIX.1-2017 open(2).

### DIFF
--- a/kernel/src/test_runner.rs
+++ b/kernel/src/test_runner.rs
@@ -1316,6 +1316,22 @@ pub fn run() -> ! {
     total += 1;
     if test_mprotect_split_inode_pin_balance() { passed += 1; }
 
+    // ── Test 295: /proc/self/fd/N reopen of an UNLINKED memfd ─────────────
+    // procfs magic-symlink contract (proc(5)): opening /proc/<pid>/fd/<N>
+    // must DUP the underlying open file by its backing inode, not re-resolve
+    // the link target's pathname.  An unlinked memfd (memfd_create(2)) has no
+    // resolvable name, so re-resolving the (removed) path ENOENT'd and broke
+    // Firefox's shared-memory read-only reopen.  Verifies: (a) the reopen
+    // succeeds, (b) the new fd reads the memfd contents from a fresh offset,
+    // (c) a read-only reopen honours O_RDONLY on the new fd.  Placed beside the
+    // inode-pin tests (119b/119c) — same VFS anonymous-inode surface — and
+    // ahead of the heavy Firefox-launch oracle so it runs on every build.
+    #[cfg(any(feature = "firefox-test-core", feature = "test-mode"))]
+    {
+        total += 1;
+        if test_295_proc_fd_memfd_reopen() { passed += 1; }
+    }
+
     // ── Cheap-log-transport: ring framing/drain correctness + cost ─────────
     // Cross-subsystem: drivers::log_ring (the PMM-backed guest-RAM ring) +
     // drivers::serial (the fast-path routing + COM1 fallback).  One test
@@ -27421,9 +27437,15 @@ fn test_x11_big_requests_enable() -> bool {
 // ── Test 134: X11 MIT-SHM — QueryExtension present + ShmQueryVersion ─────────
 
 fn test_x11_query_extension_mit_shm() -> bool {
-    test_header!("X11 MIT-SHM — QueryExtension present + ShmQueryVersion");
+    test_header!("X11 MIT-SHM — QueryExtension NOT present + absent from ListExtensions");
 
-    use crate::x11::proto;
+    // Regression for the SHM-deadvertise paint fix (equivalent to the
+    // purge-lost Test 75b): the server does NOT implement the MIT-SHM image
+    // transport, so it must report the extension as ABSENT.  Per the MIT-SHM
+    // protocol spec, a client that gets present=0 from QueryExtension falls
+    // back to core XPutImage(72), which `op_put_image` composites correctly.
+    // Advertising present=1 would make Firefox route paints through the
+    // unsupported SHM CopyArea arm and silently drop the window contents.
 
     let fd = x11_connect_and_setup("mitshm");
     if fd == u64::MAX {
@@ -27431,7 +27453,7 @@ fn test_x11_query_extension_mit_shm() -> bool {
         return false;
     }
 
-    // ── QueryExtension("MIT-SHM") ─────────────────────────────────────────────
+    // ── QueryExtension("MIT-SHM") → must reply present=0 ─────────────────────
     // Name = 7 bytes → padded to 8; header=8 → total=16 = 4 words.
     let name = b"MIT-SHM"; // 7 bytes
     let nlen = name.len() as u16;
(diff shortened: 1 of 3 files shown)


me1iissa:
## Reviewer-flagged sentinel-fd UAF: gated (commit `c40d3e1c`)

The blocking finding is fixed. `reopen_proc_fd` now declines to clone any fd that is **not inode-backed**, so the magic-symlink open path can no longer underflow a sentinel fd's type-specific refcount.

### The fix (in `kernel/src/vfs/mod.rs`)

After snapshotting the target fd's backing and **before** building the new slot, `reopen_proc_fd` gates to inode-backed types:

```rust
if mount_idx == usize::MAX
    || is_console
    || !matches!(file_type,
        FileType::RegularFile | FileType::Directory | FileType::SymLink) {
    return Err(VfsError::NotFound);
}
```

and `open()` now **falls through** to normal resolution on `NotFound` instead of returning it directly:

```rust
match reopen_proc_fd(pid, target_pid, fd_num, open_flags) {
    Err(VfsError::NotFound) => {}   // sentinel fd or gone -> normal resolution
    other => return other,           // success / hard error returned directly
}
```

This restores the **exact pre-interception semantics** for sentinel fds: the procfs symlink resolves to `/dev/fd/N`, re-resolves, and returns a benign `ENOENT` for a nameless fd (per proc(5)). No clone is created, so:

- **eventfd** — no slot to free under the original (`ipc::eventfd::close` frees outright with no refcount; this was the hard UAF, reachable via Firefox AudioIPC/IPC fds).
- **AF_UNIX / AF_INET socket** — no unmatched `ref_count` decrement.
- **pipe** — no reader/writer undercount, and no discarding of the flags type-discriminator bits.

The **memfd path is unchanged**: `RegularFile` + real `mount_idx` passes the gate, so Mozilla's read-only shmem reopen still clones the `(mount_idx, inode)` slot (self-counting via the fd-table scan).

### Verification (all via `scripts/qemu-harness.py`)

- **Test 295** (memfd reopen) — PASS (memfd RegularFile path intact).
- **Test 295b** (new sibling) — PASS: reopen `/proc/self/fd/<N>` of an **eventfd** declines cleanly, and the original eventfd stays fully usable afterward (read initval=7, EAGAIN-on-empty, write/read roundtrip). Exercised through the real `dispatch_linux_kernel` close path (the eventfd-aware `sys_close`), so it covers the exact close routing the UAF would have corrupted.
- **X11 SHM / XFIXES / DAMAGE / XI2 extension handlers** — PASS (de-advertise change unaffected).
- Other failures in the run (`monotonic-rate` timer-IRQ flake, `TCC compile` / `busybox basic` = missing `/disk/bin/*` in this data image) are pre-existing and unrelated to this diff.

**`--ff-gui` sanity**: FF (pid 34) loads libmozsandbox/liblgpllibs/libmozsqlite3/libmozgtk/libmozwayland/**libxul**, forks+execs+exits the glxtest child cleanly, and advances to sc≈3460 with active multi-process operation. **No "read-only dup failed" lines, no memfd-reopen ENOENT.** The memfd-reopen gate is fully intact; FF will still hit the deeper pid3 CR2=0xd0 mm gate (separate workstream).

Diff: +137 / −1 across `kernel/src/vfs/mod.rs` and `kernel/src/test_runner.rs`. Ready for re-review.

PR #599, opened 2026-06-19
build/ff: provision Mesa software-GL (llvmpipe) runtime for the windowed GLX path

me1iissa:
## What

Provisions the **Mesa software-GL stack (Gallium llvmpipe)** into the musl
Firefox test image so the windowed (`--ff-gui`) GPU-feature probe (glxtest)
can `dlopen("libGL.so.1")` and run a GLX handshake against the in-kernel X11
GLX server. Stacks on top of the server-side GLX handshake + real MIT-SHM
`ShmPutImage` work on this branch.

Before this change glxtest failed at `dlopen("libGL.so.1")` (`libGL.so.1
missing`) and never reached the GLX server.

## Changes

1. **`scripts/install-firefox-musl.sh`** — the permanent fix. Appends a Mesa
   package set (`mesa-gl mesa-egl mesa-dri-gallium mesa-gbm`) to the rootfs
   `apk add`, pinned to the same Alpine v3.20 repo as the rest of the FF stack
   (mesa 24.0.9 / llvm17 17.0.6). apk pulls the transitive deps (mesa-glapi,
   llvm17-libs, libelf, libxshmfence, libxxf86vm, wayland-libs-server). The
   libraries land under the rootfs `/usr/lib` tree (the llvmpipe DRI driver at
   `/usr/lib/xorg/modules/dri/libgallium_dri.so`, libGL's compiled-in default
   `LIBGL_DRIVERS_PATH`) and flow through the existing
   `cp -aL ${ROOTFS}/usr/lib/.` staging step unchanged, so a regenerated image
   always carries Mesa. Three durability hooks: the rootfs-add, an idempotent
   "ensure libGL in rootfs" backfill (sentinel `libGL.so.1`), and the
   up-to-date skip-gate + a loud `[FF-MUSL][GL]` staging check both gate on
   `libGL.so.1` + the llvmpipe DRI driver.

2. **`kernel/src/gui/terminal.rs`** — points the staged Mesa loader at the CPU
   rasteriser on a no-GPU/no-DRM target: `GALLIUM_DRIVER=llvmpipe`,
   `MESA_LOADER_DRIVER_OVERRIDE=llvmpipe`,
   `LIBGL_DRIVERS_PATH=/usr/lib/dri:/usr/lib/xorg/modules/dri`, alongside the
   existing `LIBGL_ALWAYS_SOFTWARE=1`.

## Verification (live, harness `--ff-gui` against the Mesa image)

[...]

PR #600, opened 2026-06-19
x11: real MIT-SHM ShmCreatePixmap shared pixmaps + windowed-FF GdkPixbuf loaders.cache provisioning

me1iissa:
## Summary

Drives the **windowed** (`--ff-gui`) Firefox paint path forward by closing two
divergences and assembling the combined windowed-paint branch. Builds on
`feat/mesa-gl-runtime` (GLX #599/#598 + real ShmPutImage) and includes the
mremap (#596) and prefs (#597) fixes via cherry-pick.

### 1. Real MIT-SHM `ShmCreatePixmap` + CopyArea-from-SHM-pixmap (kernel)
`kernel/src/x11/{mod,resource}.rs`. The X11 MIT-SHM shared-pixmap present path
(the one `nsShmImage`'s `gUseShmPixmaps` prefers) was a no-op stub, and
`ShmQueryVersion` advertised `shared_pixmaps=false`.

- `ShmQueryVersion` now advertises `shared_pixmaps=true`.
- `ShmCreatePixmap` creates a real Pixmap resource whose pixel storage **is** the
  attached SysV-SHM segment — no server-owned `Vec`. It records the resolved
  shmid, byte offset, and scanline-padded row stride after validating the image
  region fits inside the segment.
- `CopyArea(pixmap → window)` reads an SHM-backed source's **live** pixels from
  the segment's kernel-readable physical backing (`sysv_shm::segment_phys`) at
  copy time, so the frame the client just drew is the one composited. Server-owned
  pixmaps keep their existing `Vec`-read path; the pixmap→pixmap arm is
  bounds-guarded against the empty `Vec` of an SHM-backed source.

Per the X11 MIT-SHM protocol §ShmCreatePixmap.

### 2. GdkPixbuf `loaders.cache` always provisioned (build)
`scripts/create-data-disk.sh`. The windowed path realizes a real toplevel, so GTK
decodes its compiled-in symbolic PNG titlebar icons through GdkPixbuf. With no
`loaders.cache` at the env-pointed path, GdkPixbuf comes up with an **empty
loader set**, the first PNG decode returns NULL, and the
`gdk_cairo_surface_create_from_pixbuf(NULL)` / `g_object_unref(NULL)` cascade
dereferences a garbage pointer — a **deterministic SIGSEGV in the parent before
any window is mapped** (observed: `Could not load a pixbuf from
.../window-minimize-symbolic.symbolic.png ... pixbuf loaders ... could not be
found`, then a repeated identical-RIP fault, `data_phys=UNMAPPED`).

`install-firefox-musl.sh` generated this cache only on FF (re)install; the common
fast path reuses an already-installed `build/disk` and skipped it. Now
`patch-gui-caches.sh --stage-only` runs unconditionally during the data-image
build. The PNG/JPEG decoders are linked into `libgdk_pixbuf` (DT_NEEDED
libpng16/libjpeg), so the cache needs only the name-matched png/jpeg stanzas.

Refs: gdk-pixbuf-query-loaders(1), glib-compile-schemas(1).

## Verification

- `test_x11_shm_create_pixmap` (new): ShmQueryVersion(shared_pixmaps=1) →
  ShmAttach → ShmCreatePixmap → **change the segment** → CopyArea → assert the
  window shows the post-create pixels. Read back `[0x44,0x55,0x66,0xFF]` =
  proven live read. **PASS.** Full suite unregressed (213 PASS; only [FAIL]s are
  the pre-existing tcc/busybox-not-in-image env cases).
- Windowed FF boot (`--features firefox-test-core,kdb --ff-gui`) against a data
  image carrying the cache: the parent SIGSEGV (deterministic, 5× identical-RIP)
  is **eliminated**; `Could not load a pixbuf` = 0, `exit_group(11)` = 0. FF
  advances from "ff-launch crash" to a clean multi-thread startup.

## Residual gate (NOT yet painted — precisely localized)

Post-fixes, FF reaches a healthy startup but its main thread (tid 2) + all 30
threads **park before realizing the 1280×720 toplevel**. The in-kernel X server
is fully drained (owes no reply); FF never reaches the present path (0
ShmAttach/ShmCreatePixmap). Screendump = empty desktop, no FF window. This is the
parked W101-class multi-thread event-loop/work-source stall, now manifesting in
the windowed path. Reported for follow-up per the saga-exhaustion rule rather
than re-opening the parked futex/event-loop investigation.

### DIFF
--- a/.ai/PROGRESS.md
+++ b/.ai/PROGRESS.md
@@ -6,6 +6,15 @@
 **GUI Tests**: 10/10 pixel checks passing (`scripts/run-gui-test.sh`)
 **Firefox**: Runs for full 15000 ticks without crashing (was: crash at 1481 syscalls at RIP=0x0)
 
+### Milestone 46 — Windowed Firefox: MIT-SHM shared pixmaps + GdkPixbuf loaders.cache (✅)
+**Completed**: 2026-06-19
+
+**What was built (combined windowed-paint branch):**
+- [x] **Real MIT-SHM `ShmCreatePixmap` + CopyArea-from-SHM-pixmap** (`kernel/src/x11/{mod,resource}.rs`): `ShmQueryVersion` now advertises `shared_pixmaps=true`; `ShmCreatePixmap` creates a real Pixmap resource whose pixel storage IS the attached SysV-SHM segment (records shmid/offset/scanline-padded stride, validates the image region fits the segment); `CopyArea(pixmap → window)` reads an SHM-backed source's LIVE pixels from the segment's physical backing at copy time. This is the second X11 present path (`gUseShmPixmaps`) — the one nsShmImage prefers when the server advertises shared pixmaps. Was a no-op stub. Headless test `test_x11_shm_create_pixmap` proves live-read (segment changed AFTER CreatePixmap; CopyArea sees the new pixels). Cites X11 MIT-SHM protocol §ShmCreatePixmap.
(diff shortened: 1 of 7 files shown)


me1iissa:
Superseded: MIT-SHM ShmCreatePixmap + GdkPixbuf loaders.cache provisioning were consolidated onto master via #606 (58577ed0). Closing.

PR #609, opened 2026-06-21
kernel/epoll: close the EPOLLET drain-then-rise re-arm gap (rise generation)

me1iissa:
## Summary

The EPOLLET edge-trigger support added in the prior epoll work (Test 305) re-armed an edge only when an `epoll_wait` observed the target fd's asserted level drop to 0 between two calls. Per **epoll(7)** ("Level-triggered and edge-triggered notification"), each not-ready → ready transition is a distinct edge that must be reported regardless of `epoll_wait` timing.

The level-diff alone cannot see a **drain-then-rise that happens entirely between two `epoll_wait` calls**: no call observes the intermediate `level==0`, so the asserted level is identical at both observations and `level & !last_ready == 0` suppresses the genuine new edge. (This is the documented residual follow-up to the EPOLLET edge-trigger work.)

## Fix

Record the `0 → non-zero` crossing at the readiness source:

- **eventfd** (`eventfd(2)`) carries a monotonic `rise_seq` generation, bumped on every counter `0 → non-zero` transition, never on a write that only grows an already-non-zero counter.
- An EPOLLET epoll watch records the generation it last delivered (`et_rise`); `sys_epoll_wait` re-arms the EPOLLIN edge whenever the live generation differs from the recorded one, even when the asserted level looks continuously high.

This is the canonical epoll ready-list / wakeup-callback contract: the 0-crossing is captured when readiness rises, not only when an `epoll_wait` happens to observe the drained state. Sources without a rise generation report 0, so the existing level-diff behaviour governs them unchanged; level-triggered watches are unaffected.

## Diagnostic tooling

Adds a kdb `epoll-watch <pid>` op (and harness wiring) that dumps, per epoll instance owned by a pid: each watch's interest mask, the EPOLLET flag, the recorded `last_ready`/`et_rise` edge state, the live computed readiness level of the target fd, and (for eventfd targets) the slot counter + `rise_seq`, plus an `edge_suppressed` fingerprint. This makes the edge-trigger contract directly observable from the harness. The live level is computed after the process-table lock is dropped (the readiness poll re-acquires it).

## Test

**Test 307** (test-mode) exercises the residual gap directly: an eventfd is drained to 0 and re-raised entirely between two `epoll_wait` calls with no intervening level-0 observation, asserting the EPOLLIN edge re-fires exactly once and is then re-suppressed while it stays ready. Test 305's persistently-ready suppression is reasserted as a sanity check.

## Scope note

This closes a documented edge-trigger correctness gap and adds reusable diagnostics; it does **not** by itself unblock any specific application gate. Level-triggered watches (the X11/pipe/socket default) are unchanged.

Refs: epoll(7), eventfd(2).

### DIFF
--- a/kernel/src/ipc/epoll.rs
+++ b/kernel/src/ipc/epoll.rs
@@ -47,6 +47,19 @@ pub struct EpollWatch {
     /// reporting is unconditional.  Reset to 0 on `EPOLL_CTL_MOD` so a MOD
     /// re-arms the edge (Linux behaviour).
     pub last_ready: u32,
+    /// Edge-trigger rise generation last *delivered* for this watch, for
+    /// readiness sources that expose a monotonic 0→ready rise counter (the
+    /// eventfd `rise_seq`).  The `last_ready` level-diff alone cannot detect a
+    /// drain-then-rise that happens entirely between two `epoll_wait` calls
+    /// (no call observes the intermediate level==0, so the asserted level is
+    /// identical at both observations and `level & !last_ready == 0`).  Per
+    /// `epoll(7)` each not-ready→ready transition is a fresh edge that must be
+    /// reported regardless of `epoll_wait` timing; comparing the source's live
+    /// generation against this stored value re-arms the EPOLLIN edge across
+    /// such a transition.  Sources without a generation report 0, so the
+    /// level-diff behaviour governs them unchanged.  Ignored for
+    /// level-triggered watches; reset to 0 on ADD / `EPOLL_CTL_MOD`.
+    pub et_rise: u64,
 }
 
 /// Per-process monotonically increasing identifier for epoll instances.
@@ -141,7 +154,7 @@ impl EpollInstance {
         // any currently-asserted readiness (a 0→ready transition), matching
         // Linux, where a newly-added EPOLLET fd that is already ready fires
         // once immediately.
-        self.watches.push(EpollWatch { fd, events, data, last_ready: 0 });
+        self.watches.push(EpollWatch { fd, events, data, last_ready: 0, et_rise: 0 });
         true
     }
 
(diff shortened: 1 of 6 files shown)


me1iissa:
## Independent review (qa-engineer subagent) — APPROVE-WITH-NITS, confidence 0.92

Adversarially verified all 5 dimensions and **ran Test 307 locally** (test-mode, non-FF stub disk):
- Test 307 (a) edge fires once / (b) stays-ready suppressed / (c) drain-then-rise re-arms / (d) re-suppressed after re-delivery — all ✓
- Test 305 (b/c/d) re-asserted ✓; tally 384/396 (the residual fails are the known data.img env-gaps already on the allow-fail list).

Findings:
- Rise-generation contract sound: `rise_seq` bumps exactly on 0→non-zero (`was_zero` captured before `+= val`); `et_rise` is committed after delivery so no infinite re-fire (the exact failure mode #605 fixed for the level case, verified via checkpoint (d)). LT watches and non-eventfd sources (pipe/unix/timerfd) fall through to the unchanged level-diff path.
- `commit_edges` has one caller, fully updated; (fd, level, rise) tuple consistent end-to-end.
- kdb `epoll-watch` is deadlock-free: snapshots under PROCESS_TABLE via `try_lock_brief`, computes the live level after the lock is dropped.
- Test 307 is a genuine regression test (computes `EPOLLIN & !EPOLLIN = 0` on the unfixed kernel; won't even compile against master's 2-tuple `commit_edges`). No 306/307 collision (306 = the writev test, 307 = this).
- Public-spec citations only.

Process nit (separate, not blocking): the kernel-smoke CI step-log is truncated mid-suite, so CI's PASS does not by itself evidence the 291–307 test band — hence the local run. Suggest a follow-up to emit the harness `results` JSON tally as a CI artifact. Cosmetic nit (wrapping_add intent comment) addressed in the follow-up commit.

PR #611, opened 2026-06-21
ff: windowed Firefox navigates & renders a real site (disable remote-audio cubeb park)

me1iissa:
## Milestone: windowed Firefox reaches the event loop, navigates, and renders content

The windowed-Firefox (`--ff-gui`) main thread parked permanently in the XPCOM
`ThreadEventQueue` condvar **before** reaching the GLib/GTK `poll()` loop, so the
browser never navigated or painted (0 DNS / 0 connect). A live GDB RBP-walk on a
prior boot localized the park to Firefox's lazy **cubeb** audio initialisation:
on a target with no audio device the parent process takes the sandboxed
out-of-process **AudioIPC** (cubeb-remoting) branch, which synchronously awaits an
audio-server handshake that never completes.

This is a **userspace profile-config** fix (the kernel futex is faithful — the
condvar was simply never signalled because the producer never ran), the same
class as the existing background-network prefs.

### Changes
1. **`kernel/src/main.rs` (seeded `prefs.js`)** — pin the audio backend so a media
   context never initialises AudioIPC:
   - `media.cubeb.force_null_context = true` — `GetCubebUnlocked()` returns a null
     context immediately (no AudioIPC, no in-process `cubeb_init`).
   - `media.cubeb.sandbox = false` — belt-and-suspenders in-process branch;
     honoured because the content sandbox is disabled via
     `MOZ_DISABLE_CONTENT_SANDBOX` in the launch env (otherwise Firefox forces the
     sandboxed audio path back on).
   - `media.volume_scale = "0.0"` — silence.
2. **`kernel/src/gui/terminal.rs` (`spawn_async` env)** — select the windowed
   process model from the launch URL: a **local** `file://` target keeps the
   in-process collapse (`MOZ_FORCE_DISABLE_E10S` + `MOZ_DISABLE_NONLOCAL_CONNECTIONS`)
   for the single-process render path; a **network** `http/https` target drops both
   so the normal multi-process content tree comes up and necko may fetch the page.
   Background-network features remain pref-disabled.
3. Bundles the staged **golden no-WM map burst** (`2a516672`,
   MapNotify→VisibilityNotify→Expose→EnterNotify+FocusIn for toplevels, +Test 308) —
   correctness-positive X11 map-handshake parity.

### Verification (SMP=1, `--ff-url https://example.com`, software-GL image)
- **Main thread unparked**: kdb shows `tid2` (main) state `ready`/running, not
  parked on a condvar (was permanently `pthread_cond_timedwait`).
- **Chrome painted** (verified pixels): toplevel `0x400016` mapped at 1280x972,
  11× PutImage to it; full FF chrome — tab `example.com/`, URL bar
  `https://example.com`, nav/menu buttons.
- **Navigation fired** (pcap): `A? example.com` resolved to `104.20.23.154`,
  `172.66.147.243`; full necko alive (firefox.settings + contile queries).
- **Content rendered** (verified pixels): the content area painted a full content
  page (Firefox's `about:neterror` for the residual gate below) — heading, body,
  bullet list, and button all drawn end-to-end.
- Multi-process tree healthy: parent (99 threads) + 5 content procs, all active,
  syscall count climbing past 200k (no crash, no channel error).

### Residual (separate, downstream gate — NOT cubeb)
Content procs load NSS softoken (`libsoftokn3.so`) but the TLS layer fails to
init (0 packets to :443; the rendered page is "Unexpected response from server /
Personal Security Manager"). This is a distinct NSS/PSM TLS-init gate, cleanly
downstream of the cubeb park this PR removes. Full HTTPS page render is the next
lane.

Refs: Firefox media prefs (`media.cubeb.*`),
https://firefox-source-docs.mozilla.org/dom/ipc/process_model.html , RFC 8089.


me1iissa:
Superseded by #628 (consolidated windowed render config: cubeb-disable + ASAP + golden map-burst + proc-cap prefs), merged to master.

PR #633, opened 2026-06-22
linux/scm: commit SCM_RIGHTS fds only with their first data byte (+ mremap content-preservation test)

me1iissa:
## Summary

Two changes, both grounded in deterministic regression tests, targeting the windowed-Firefox **degenerate WebRender blob** crash (content ships a blob image whose buffer is a font path + zeros, trailer=0 → `moz2d_renderer.rs` `BlobReader` assertion / replay panic).

1. **STEP-1 refutation (Test 126c)** — `mremap(MREMAP_MAYMOVE)` grow preserves the full old content. The dispatch's prime suspect (realloc→mremap content-loss) is **refuted**: the test PASSES (all old bytes preserved across a forced relocation, tail zero), and the golden Firefox strace shows **0 mremap / 0 brk** (musl mallocng serves sub-page reallocs in userspace). mremap is faithful and off the FF critical path.

2. **STEP-3 candidate fix + bug (Test 295)** — `sendmsg(2)` was committing `SCM_RIGHTS` fds **before** the data push, with no rollback on a zero-byte EAGAIN (full recv ring). A sender that retries the same EAGAIN'd frame (the standard non-blocking IPC pattern) re-queued the **same** fds each attempt → duplicate batches at one stream offset → a positional fd→resource receiver (WebRender's shared-memory transport) maps the **wrong chunk** for a resource (font data where the blob recording was expected). Fix: revoke the batch + balance its refs when zero bytes are accepted, returning EAGAIN so the frame is retried with its fds intact. Control-only frames still commit unconditionally. The queue-before-push race fix is preserved (a zero-byte send can't advance the peer past the bind offset, so the batch is guaranteed un-delivered and safe to revoke).

## Evidence

- **Test 126c**: PASS — `all 0x3000 old bytes preserved across relocation; 0x5000-byte tail zero`.
- **Test 295 A/B (deterministic)**:
  - **With fix**: PASS — `EAGAIN send #1: no batch committed (pending=0)`, `retry: still no batch`, `successful send: exactly ONE batch`.
  - **Without fix**: FAIL — `sendmsg returned EAGAIN (-11) but queued an fd batch: pending 0->1 (fds committed WITHOUT their first byte)`.
- Test suite otherwise green (the only FAILs are pre-existing environmental: `tcc`/`busybox` binaries absent from the minimal test disk).
- Both `test-mode` and `firefox-test-core` builds compile clean; the fix is in the always-compiled sendmsg path.

## Scope / honesty

The SCM fix is a **real, deterministic kernel-ABI correctness bug** (a `sendmsg` zero-byte-EAGAIN committing fds, divergent from the unix(7)/recvmsg(2) SCM_RIGHTS contract) and is the **strongest candidate** for the degenerate-blob root: structural analysis shows the blob assert fires in the **parent** process (no GPU process → WebRender Renderer is parent-side; content only records + IPDL-sends; `BlobImageData` is a deep copy), and blob + font share one shared-memory chunk pool distinguished only by positional `OffsetRange` + fd arrival order. It is **not yet confirmed** to be the specific cause of a given render's degeneration — the live moz2d crash was not captured on the boots attempted (they took the separate plateau / kernel-#GP-teardown variant). **No render claim is made here.** A live GDB read of the parent `BlobReader` buffer + `OffsetRange` is the decisive next probe if reopened.

Refs: `sendmsg(2)`, `recvmsg(2)`, `unix(7)` SCM_RIGHTS, IEEE 1003.1-2017 §2.10.6, `mremap(2)`.

### DIFF
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -2255,6 +2255,19 @@ fn dispatch_body(num: u64, arg1: u64, arg2: u64, arg3: u64, arg4: u64, arg5: u64
             // with at least one non-empty iovec is data-bearing.
             let frame_has_data = iovecs_owned.iter().any(|iov| iov[1] != 0);
             let mut total = 0usize;
+            // If an SCM_RIGHTS batch is queued ahead of the data push (below) and
+            // the push then accepts ZERO data bytes (recv ring full → EAGAIN), the
+            // batch must be revoked: per unix(7) / recvmsg(2) SCM_RIGHTS,
+            // ancillary fds are delivered with the message bytes they accompany —
+            // a zero-byte send carries no bytes and so commits no fds.  Without
+            // revoke, a sender that retries the same EAGAIN'd frame (the gecko IPC
+            // channel does exactly this) re-queues the SAME fds, leaving duplicate
+            // batches at one stream offset; the WebRender receiver then maps the
+            // wrong shmem chunk for a resource (a degenerate blob image).  Track
+            // the queued batch's bind-peer + offset so the zero-byte return paths
+            // can roll it back.  Only set when frame_has_data (a control-only
+            // frame legitimately commits its fds with no data byte).
+            let mut scm_queued: Option<(u64, u64)> = None; // (receiver_id, byte_offset)
             // ORDERING (race fix): the SCM_RIGHTS batch is queued BEFORE the data
             // bytes are pushed into the peer's recv ring.  The data push and the
             // batch enqueue are separate critical sections (net::unix TABLE lock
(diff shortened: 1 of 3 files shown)


me1iissa:
## Independent review resolution

Two independent review agents looked at this PR.

**Correctness (dims 1-5): APPROVED.** The revoke logic is lock-safe (scm_drop_fds runs with no TABLE/PENDING_SCM held), refcount/inode-pin balanced (exact inverse of the enqueue bumps), behavior-preserving (control-only frames commit unconditionally; zero-byte-no-fds still returns 0; partial-write does not revoke; AF_INET unaffected), and yields exactly-one batch.

**The one raised blocker was REFUTED** (independent QA verification, conf 0.98). The blocker claimed Test 295's kernel-stack `hdr` would be rejected by `validate_user_ptr` (nr=46, syscall.rs:2204) with -14, so the test never reaches the SCM code. This is contradicted by the serial ground-truth and control flow:
- The `-14` guard is the only -14 exit in the nr=46 arm header; all SCM logic is strictly downstream.
- The **unfixed** A/B run observed `sendmsg returned EAGAIN (-11) ... pending 0->1` — physically unreachable if the guard had rejected the pointer, so the guard returned true.
- `validate_user_ptr` rejects only `ptr >= USER_VA_LIMIT (0x0000_8000_0000_0000)`. The test-runner BSP kernel stack is LOW identity-mapped (well below that), so its `hdr` is accepted. The blocker conflated this with `test_security_user_ptr_validation_cwe823`, which passes a HIGH-half KBASE (0xFFFF_8000...) and correctly gets -14 — a different address range.

A/B serial evidence (dispositive):
- **Unfixed** (SCM_REVOKE_FIX off): `[FAIL] scm_no_commit — sendmsg returned EAGAIN (-11) but queued an fd batch: pending 0->1`.
- **Fixed**: `EAGAIN send #1: no batch committed (pending=0)` / `retry: still no batch` / `successful send: exactly ONE batch` → `[PASS]`.

**Non-blocking follow-up addressed:** softened the `scm_revoke` docstring to make the `None` path explicitly load-bearing under SMP (a concurrent sender's byte can make the batch deliverable before revoke; the code already handles this safely via rposition→None→no-drop). Commit 28bd5cdd.

PR #635, opened 2026-06-23
mm/cache, vfs: ramfs/tmpfs MAP_SHARED↔read coherence (Test 641 C must-pass)

me1iissa:
## Summary

Closes a real latent POSIX `mmap(2)` MAP_SHARED coherence bug in the
ramfs/tmpfs dual-storage model and flips **Test 641 case C** from a
documented latent gap to a hard must-pass.

An in-memory filesystem (ramfs / tmpfs) keeps a file's bytes in two
places once a page is demand-faulted: the inode's own buffer (served to
`read(2)`/`write(2)`/`stat`) and the page-cache frame an `mmap` PTE
aliases. A `MAP_SHARED | PROT_WRITE` store lands **only** in the cache
frame; the inode buffer is never updated. If that frame is evicted and
freed, a later cache-miss `read(2)` returns the stale inode buffer
(for a freshly-sized memfd, zeros) — violating POSIX mmap(2) MAP_SHARED
visibility.

### Fix (both coherence directions, scoped to ramfs/tmpfs)
- **Writeback-on-eviction** (`mm/cache`): flush an in-memory-FS frame's
  bytes back into the inode buffer before it leaves the cache
  (`cache::evict` + the collision eviction in `cache::insert`), deferred
  until after the cache lock is released (lock order cache → MOUNTS →
  inode; the file-backed page-fault path already drops MOUNTS before
  `cache::insert`, so the collision-eviction writeback cannot re-enter
  MOUNTS). Not applied to `truncate_range` (pages intentionally
  discarded per `ftruncate(2)`) nor `evict_if_phys` (lost-race frame).
- **Cache-authoritative read** (`vfs::fd_read`): overlay any
  cache-resident page's current bytes onto the read result for in-memory
  filesystems, so an mmap write in a still-resident frame is observed.
- New `FileSystemOps::is_in_memory()` (default false; true for RamFs /
  TmpFs). Block-backed (ext2/fat32) re-read the on-disk image and are
  unaffected — the writeback/overlay short-circuit on `!is_in_memory`.

## Test / verification
- **Test 641 case C** (eviction-then-remap dual-storage hazard) is now a
  hard assertion. test-mode SMP=1: 641 A/B/C/D all PASS
  (`C: producer write survived eviction via writeback-to-inode`).
- Full suite: **132 pass / 2 pre-existing env fails** (missing
  `/disk/bin/tcc`, `/disk/bin/busybox`).
- Compiles clean: `test-mode`, `firefox-test-core`, default desktop,
  `firefox-test-core,w215-diag`.

## ⚠️ Scope — this is NOT the FF Wikipedia render fix
A live windowed FF Wikipedia boot (firefox-test-core, SMP=1) on this
branch shows the render gate is **not** closed: the moz2d degenerate-blob
crash is byte-identical to baseline (`sig=11 CR2=0x0` →
`gfx/webrender_bindings/src/moz2d_renderer.rs` panic "out of range for
slice of length 689"). The fix does **not** engage on the blob path:
the blob memfd is consumed via the mmap-aliased frame in-process, not via
`read(2)` (249 mount-0 MAP_SHARED+WRITE mmaps, **zero** `fd_read` of any
memfd/`/tmp` path). The dual-storage-incoherence root cause is therefore
refuted as the live render gate; the 689-byte blob is degenerate
as-written (graphics/blob lane), not a kernel `fs.read`-returns-zeros.

This PR stands on its own as a correctness fix; it does not claim a
render result.

Refs: POSIX.1-2017 mmap(2), memfd_create(2), Intel SDM Vol. 3A §4.10.5.

### DIFF
--- a/kernel/src/mm/cache.rs
+++ b/kernel/src/mm/cache.rs
@@ -98,6 +98,91 @@ pub fn lookup(mount_idx: usize, inode: u64, page_offset: u64) -> Option<u64> {
         .map(|e| e.phys)
 }
 
+/// PHYS → kernel-higher-half direct-map offset.  Same constant as
+/// `mm/pmm.rs`; the page cache reads/writes cached frames through this map.
+/// Per Intel SDM Vol. 3A §4.10.5 the higher-half identity map covers every
+/// PMM frame, so `PHYS_OFFSET + phys` is always a valid kernel VA for an
+/// allocated frame.
+const PHYS_OFFSET: u64 = 0xFFFF_8000_0000_0000;
+
+/// Write a cache frame back to its in-memory filesystem's inode buffer.
+///
+/// ## Why this exists — ramfs/tmpfs MAP_SHARED dual-storage coherence
+///
+/// An in-memory filesystem (ramfs / tmpfs) keeps a file's bytes in TWO
+/// places once a page is demand-faulted:
+///
+///   1. the inode's own buffer (`RamInode::File::data`), which is what
+///      `FileSystemOps::read` / `write` / `stat.size` operate on; and
+///   2. the page-cache frame the demand-fault path installs, which a
+///      `mmap(MAP_SHARED)` PTE aliases directly.
+///
+/// A `MAP_SHARED + PROT_WRITE` store lands ONLY in the cache frame — the CPU
+/// writes the frame through the aliasing PTE; the kernel is never on that
+/// write path, so the inode buffer is never updated.  If the frame is then
+/// evicted (and freed), a subsequent `read(2)` (which goes to the inode
+/// buffer via `FileSystemOps::read`) returns the STALE, never-updated bytes
+/// — for a freshly-`ftruncate`d memfd, all zeros.  This is the
+/// dual-storage incoherence that empties Firefox's WebRender display-list /
+/// blob transport.
+///
+/// This helper restores coherence by flushing the cache frame's current
+/// bytes back into the inode buffer just before the frame leaves the cache,
+/// so the inode buffer becomes the authoritative copy of any mmap writes for
+/// any later cache-miss `read`.  For block-backed filesystems (ext2 / fat32)
+/// `is_in_memory()` is false and this is a no-op — those filesystems re-read
+/// the on-disk image and never alias a writable cache frame in this way, so
+/// they are unaffected.
+///
+/// Per POSIX.1-2017 mmap(2): a `MAP_SHARED` write "shall change the
+/// underlying file object" and be visible to a subsequent `read(2)`.
+///
+/// ## Lock order
+///
+/// This MUST be called with NO cache lock held.  It snapshots the
+/// filesystem handle via [`crate::vfs::fs_at`] (which takes and immediately
+/// drops `MOUNTS`), then dispatches `FileSystemOps::write`, which on
+/// ramfs/tmpfs takes only the inode mutex.  Lock order is therefore
+/// cache → (released) → MOUNTS → (released) → inode — no cycle, no nesting.
+/// Callers that hold the cache lock collect `(mount_idx, inode, page_offset,
+/// phys)` while locked and invoke this after releasing the lock, exactly as
+/// the quarantine-free deferral does.
+///
+/// `phys` must still be a live frame (the caller holds a reference, e.g. the
+/// cache's own reference that has not yet been dropped) so the direct-map
+/// read is valid.  Best-effort: a write error (e.g. a read-only tmpfs, which
+/// has no MAP_SHARED writers anyway) is ignored.
+fn writeback_frame_to_inode(mount_idx: usize, inode: u64, page_offset: u64, phys: u64) {
+    let fs = match crate::vfs::fs_at(mount_idx) {
+        Some((fs, _)) => fs,
+        None => return,
+    };
+    if !fs.is_in_memory() {
+        return;
+    }
+    // Snapshot the frame's 4 KiB through the higher-half direct map.  The
+    // frame is alive (caller holds a reference), so the read is valid.
+    let mut frame = [0u8; 4096];
+    let src = (PHYS_OFFSET + phys) as *const u8;
+    // SAFETY: `phys` is a live, page-aligned PMM frame; the higher-half
+    // identity map covers it (Intel SDM Vol. 3A §4.10.5).  We copy exactly
+    // one 4 KiB page into a stack buffer.
+    unsafe {
+        core::ptr::copy_nonoverlapping(src, frame.as_mut_ptr(), 4096);
+    }
+    // Write the page's bytes back into the inode buffer at this file offset,
+    // bounded to the file's current size so we do not zero-extend a sealed /
+    // shrunk file.  ramfs `write` grows the buffer as needed; we cap the
+    // write at `stat.size` so a partial last page does not artificially
+    // extend the file past its real end.
+    let file_size = fs.stat(inode).map(|s| s.size).unwrap_or(0);
+    if page_offset >= file_size {
+        return; // page is entirely past EOF — nothing authoritative to flush.
+    }
+    let span = core::cmp::min(4096u64, file_size - page_offset) as usize;
+    let _ = fs.write(inode, page_offset, &frame[..span]);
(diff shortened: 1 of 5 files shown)


me1iissa:
Addressed review NITs in 1ca96f2a:
- **NIT-1** (coverage): Test 641 extended with case **E** (production collision-eviction writeback inside `cache::insert`) and case **F** (`fd_read` cache-authoritative overlay of a resident frame) — the two wires that actually fire in production (`cache::evict` has no production callers; case C only drove that standalone path). Verified test-mode SMP=1: 641 A/B/C/D/E/F all PASS; suite 132 pass / 2 pre-existing env fails.
- **NIT-2** (cosmetic): removed the two function-local `const PHYS_OFFSET` shadows now that a module-level const exists.

me1iissa:
## VERDICT: REQUEST_CHANGES

Adversarial code-level review against current master `a099a254` (no QEMU). The coherence mechanism is well-designed and — importantly — **still needed**: the dual-storage gap it closes is not landed elsewhere and is not obviated by #637. But the `fd_read` cache-authoritative overlay reads a cache frame through a **reference-less** `cache::lookup` and copies it into the **user** buffer after the cache lock is released — a use-after-free read on SMP whose in-code SAFETY justification is incorrect. That one defect gates approval; a rebase-and-re-validate ask follows from the PR's age.

### Angle 1 — is the mechanism still needed? YES, and it's not contradicted by #637

Verified the semantic gap survives on current master, not just by symbol absence:
- `is_in_memory`, `writeback_frame_to_inode`, and the `fd_read` overlay are absent from master (confirmed).
- #637's `WriteFaultAction::WriteThrough` (idt.rs:1616) makes a `MAP_SHARED+PROT_WRITE` store land in the **cache frame** in place (so co-mappers see it via a cache-HIT), but it never reconciles that frame with the inode buffer that `FileSystemOps::read` serves. The demand-fault path aliases a **separate** cache frame (idt.rs ~2179/2684 "ALIAS the cache page"), distinct from `RamInode::File::data`. So the reverse direction (mmap write → subsequent `read(2)`) is still incoherent for ramfs/tmpfs.
- Test 641 C/E/F encode exactly this contract and are flipped from "latent, recorded" to hard-fail. That is the correct ground-truth gate, and it genuinely fails on master today.

So the PR is complementary to #637, not redundant. Good.

### F1 (BLOCKING) — `fd_read` overlay is a reference-less read of a cache frame into the user buffer → SMP use-after-free

`vfs/mod.rs` (the new overlay):
```
if let Some(phys) = crate::mm::cache::lookup(mount_idx, inode, pg) {   // lock released, NO ref
    ...
    let src = (PHYS_OFF + phys + in_page_off) as *const u8;
    core::ptr::copy_nonoverlapping(src, buffer.as_mut_ptr().add(buf_off), len);  // into USER buf
}
```
`cache::lookup` (cache.rs:94) takes the cache lock only for the map read and returns a bare `phys` with **no** `page_ref_inc`. The frame's liveness is therefore unprotected for the entire copy. The SAFETY comment asserts "`phys` is a live cache frame (cache holds a reference)" — that reference belongs to the *cache entry*, which any other CPU can evict the instant the lock is dropped. On a concurrent `cache::evict` / collision-`insert` / `truncate_range` that drops the frame's last reference (a cached page with rc→0, i.e. no live mmap PTE — the common case for a plain cached ramfs/tmpfs page being `read(2)`), the frame is freed and recycled, and the copy pulls a recycled frame's bytes — another process's data — into this reader's `buffer`. That is both corruption and an info-leak, and it is precisely the post-recycle read class the W215 saga is about.

- Inert on SMP=1 (no concurrent evictor; the copy doesn't yield — the user pages were just faulted in by the bounce copy above, so the overlay copy itself won't fault/block). But the PR's entire reason to exist (Test 641 C) is the SMP eviction regime, so shipping an SMP UAF in the same change is self-defeating.
- `buffer` is the user buffer (`from_raw_parts_mut(buf, count)`), so "hold the cache lock across the copy" is **not** an option (a user-faulting copy under a spinlock would deadlock the #PF path that needs MOUNTS). The correct fix is the ref-taking primitive the module already provides:

```
if let Some(phys) = crate::mm::cache::lookup_and_acquire(mount_idx, inode, pg) {  // inc under lock
    ... copy ...
    let _ = crate::mm::refcount::page_ref_dec(phys);   // release after the copy
}
```
`lookup_and_acquire` (cache.rs:227) increments while the lock is held, exactly for "read a frame after releasing the lock." Use it and drop the ref after the copy, and correct the SAFETY comment.

### F2 (RE-VALIDATE, not a defect on current master) — the eviction-writeback lock order

`writeback_frame_to_inode` adds an `fs.write` (MOUNTS + inode mutex) into the collision-eviction path of `cache::insert_with_expected`, which runs on the #PF demand-fault path. The PR predates #703/#708. I traced it: the writeback is called only after the cache lock is released (cache.rs:574 and the `evict` restructure), and on current master the file-backed #PF path **drops PROCESS_TABLE before any MOUNTS/FS work** (idt.rs:1850, "Release PROCESS_TABLE to avoid deadlock with MOUNTS") and holds no `mm_sem`. The demand-fault already takes MOUNTS+inode for its fill `fs.read`, so the new `fs.write` is the same lock class in the same lock-free context — no new #703/#708 cycle, no shootdown under any lock. ramfs `write` touches only the inode Vec (no re-entry into the cache/fault path). So this is **consistent on current master** — but the PR's green CI ran against June-23 master, before that discipline and before #719's refcount lessons, so it does not actually exercise this interaction. Re-validation is required (see below), not a code change.

### Angle 3 (refcount) / Angle 4 (POSIX) — clean

- Refcount: `writeback_frame_to_inode` takes no reference and reads a frame the caller still owns until it frees it after the call returns (evict) or quarantines it (insert) — no inc/dec imbalance, no #719-class inc-without-dec. (F1 is the opposite failure: **no** ref where one is needed.)
- POSIX: the two halves (resident-overlay + writeback-on-eviction) correctly make a `MAP_SHARED` write visible to `read(2)` in both the resident and evicted cases; the writeback caps at `stat.size` so it never zero-extends a shrunk/sealed file; the resident frame is authoritative and its non-written bytes track the inode buffer via `update_range`, so whole-page writeback is correct.

### Angle 5 — rebase risk / precise re-validation

`mergeable: UNKNOWN`/git-mergeable is not semantic validation: the branch head (June 23) was CI'd against a pre-#703/#708/#719 master. The required actions before merge:
1. **Fix F1** (`lookup_and_acquire` + `page_ref_dec`; correct the SAFETY comment).
2. **Rebase onto `a099a254`** and re-run Test 641 A–F, with **C and E on SMP=2** (the eviction paths are single-CPU-inert; SMP=2 is where both the coherence gap and the F1 race are reachable) to confirm coherence holds and no deadlock/UAF is introduced by the eviction-writeback in the current #PF locking.
3. Confirm F2 empirically via that SMP=2 641E run (the collision-eviction writeback is the production path).

### Minor (non-blocking)
- Tests C/E/F drive the mechanism through direct `cache::insert`/`evict`/`fault_in` + `producer_store` rather than real `mmap`+store syscalls, so they validate the cache/inode plumbing but not the demand-fault aliasing itself (covered separately by A/B). Acceptable.
- The overlay adds a per-page `cache::lookup` (one hashmap probe under the cache lock) to every in-memory-FS `read`; modest and correctly scoped to `is_in_memory()`.

The mechanism is the right fix for a real, still-open gap. Fix the reference-less overlay read and re-validate on a current-master SMP=2 641 run, and this is an approve.

Refs: POSIX.1-2017 mmap(2) (MAP_SHARED visibility to read(2)); Intel SDM Vol. 3A §4.10.5 (TLB/frame-reuse ordering).

me1iissa:
Superseded by #721. That branch (`mm/ramfs-mapshared-coherence-r2`) cherry-picks these commits onto current master, applies the blocking F1 fix from the review (reference-less `cache::lookup` overlay read -> `lookup_and_acquire` + `page_ref_dec`/`quarantine_free`, fixing the SMP use-after-free), normalizes commit metadata, and re-validates Test 641 A-F on SMP=1 plus C/E on SMP=2. Closing in favour of #721.

PR #638, opened 2026-06-23
linux/futex: build relative-timeout deadlines at TSC precision (no early wake)

me1iissa:
## Summary

Builds the Linux-personality `FUTEX_WAIT` **relative** timeout deadline at TSC (nanosecond) precision instead of flooring it to the coarse 10 ms scheduler tick. The old path computed `wake_tick = get_ticks() + max(ns/10ms, 1)` where `get_ticks()` is a 10 ms-floored TSC tick, which guarantees only `delta_ticks − 1` *full* intervals — so a relative timed wait could return `ETIMEDOUT` up to ~10 ms (≈95% on a 10 ms wait) **before** the requested interval had actually elapsed.

This is a **POSIX correctness fix** (per `futex(2)` TIMEOUTS and `clock_nanosleep(2)`: a relative timeout must be honoured at the clock's native precision and may over-shoot but never under-shoot). musl converts an absolute deadline to a relative interval in userspace and issues a plain relative `FUTEX_WAIT`, so every musl timed wait hit this path.

## Change

- New helper `arch::x86_64::irq::relative_ns_to_wake_tick(ns)`: `deadline_cycles = (rdtsc() − tsc_at_boot) + ns_to_cycles(ns)`, `wake_tick = ceil(deadline_cycles / tsc_per_tick)`. The ceiling guarantees `get_ticks() >= wake_tick` cannot be true until the floored tick advances past the fractional deadline — observed sleep is always `>= ns` (over-shoot bounded to one tick, never under-shoot). Pre-calibration (`tsc_per_tick == 0`) falls back to coarse ticks rounded up + 1, also never under-shooting. The ns→cycles quotient is clamped to `u64::MAX` before narrowing so even a pathologically large timeout saturates to the far future rather than truncating short.
- Call site: the `TimeoutNs::Relative` arm only. The absolute-clock and `AlreadyExpired` arms are unchanged.
- Test 296 `test_relative_futex_timeout_no_undershoot`: asserts the earliest TSC at which the returned `wake_tick` can fire is `>=` the absolute deadline across 1/5/10/15/100 ms, with bounded over-shoot.

## Scope note

Originally investigated as a candidate for the GObject "signal '...' is invalid" warning burst on windowed-Firefox boots. A subsequent clean A/B soak showed that burst is benign, bounded GTK widget-registration noise (it does not block reaching content; both pre/post arms reach content with a valid test image), so this PR is **not** a fix for that symptom — it stands on its own as a futex relative-timeout correctness fix. Independently reviewed (APPROVE-WITH-NITS); the one substantive nit (u64 narrowing of the ns→cycles quotient) is addressed in this PR.

Cites: POSIX `futex(2)`, `clock_nanosleep(2)`; Intel SDM Vol. 3B §17.17 (invariant TSC).

### DIFF
--- a/kernel/src/arch/x86_64/irq.rs
+++ b/kernel/src/arch/x86_64/irq.rs
@@ -253,6 +253,63 @@ pub fn get_ticks() -> u64 {
     if tsc_floor > published { tsc_floor } else { published }
 }
 
+/// Convert a *relative* nanosecond timeout into the smallest scheduler tick
+/// at which the requested interval is GUARANTEED to have fully elapsed.
+///
+/// The scheduler's due-wake scan fires a sleeper when `get_ticks() >=
+/// wake_tick`.  Because [`get_ticks`] returns the **floor** of the TSC-derived
+/// tick (it discards the 0–10 ms already elapsed inside the current tick),
+/// naively computing `wake_tick = get_ticks() + ceil(ns / 10ms)` under-counts
+/// by the fractional part of the current tick: the wait can return up to one
+/// whole tick (~10 ms) EARLY.  For a 10 ms `pthread_cond_timedwait` that is a
+/// ~100 % early return, which trips timing-sensitive userspace state machines.
+///
+/// POSIX (`man 2 clock_nanosleep`, `man 2 futex` TIMEOUTS) permits a timed
+/// wait to over-shoot its deadline but never to return before the requested
+/// interval has elapsed.  To honour that, we build the deadline at the
+/// precision userspace measured against — the invariant TSC (Intel SDM Vol. 3B
+/// §17.17, constant-rate TSC) — rather than the coarse 10 ms tick:
+///
+///   1. `deadline_tsc = rdtsc() + ns_to_cycles(ns)`  (absolute TSC deadline).
+///   2. `wake_tick = ceil((deadline_tsc - tsc_at_boot) / tsc_per_tick)`.
+///
+/// The ceiling guarantees `get_ticks() >= wake_tick` cannot become true until
+/// the floored tick has advanced *past* the fractional deadline, so the
+/// observed sleep is always `>= ns` (over-shoot bounded by one tick, never
+/// under-shoot).  This eliminates the 10 ms quantization error entirely while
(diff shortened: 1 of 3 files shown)


me1iissa:
## Soak verification — results + caveat (honest)

**Mechanism (decisive):** the fix targets the premature `ETIMEDOUT` directly and is correct at the unit level — Test 296 passes deterministically and the
`relative_ns_to_wake_tick` conversion provably never under-shoots `ns` for the 1/5/10/15/100 ms cases (verified against the live TSC).

**End-to-end (one clean isolated windowed boot, `firefox-test-core,kdb`, SMP=1, Wikipedia/Firefox):** FF reached the chrome window
(`[X11] MapWindow 0x400016 1280x972`, log line 4117) and, critically, **advanced PAST the GObject/GTK warning burst** (which was bounded to log
lines 2903–4839) into worker-thread spawn + DNS resolution (`/etc/resolv.conf`, UDP DNS query) toward the page fetch. That is the dispatch's gate-1
success criterion: *gate-1 no longer blocks reaching content.*

**Caveat I will not paper over:** I could not complete the requested clean N≥6 flood-rate distribution. The windowed-FF boots in this environment are
dominated by a separate, flaky early-launch gate (the glxtest/`[FFTEST] DONE`-at-line-556 early exit the dispatch flagged as "~half die in glxtest"),
and after ~15 heavy KVM launches the harness `start` began failing to return a session id, so late soak boots produced no usable windowed runs. The
boots that *did* reach the flood window still emitted ~268 warning lines — i.e. the warning flood is **suppressed/survivable but not provably eliminated**
in this build, consistent with the dispatch's own framing that the premature timeout is *necessary-but-not-sufficient* for the (racy) flood. The
`[FUTEX_TIMEDOUT]` serial counter is gated behind `firefox-test-trace` (not compiled into `firefox-test-core`), so I could not use it as a direct
core-build before/after metric without changing timing.

**Net:** fix is correct, surgical, CI-green, regression-tested, and the one clean windowed boot advances past the flood into content. A clean
A/B flood-rate distribution should be re-run on a fresh box (ideally one boot at a time) to quantify the residual; flagging for the reviewer/orchestrator
rather than overclaiming suppression I did not measure cleanly.

Issue #19, opened 2026-04-23
/proc/<pid>/status — full Linux field set


me1iissa:
## Background
`/proc/<pid>/status` is currently implemented with a partial field set. Many tools (including Firefox's crash reporter, `top`, `htop`, and `ps`) parse this file and expect the full Linux field set including `VmRSS`, `VmSize`, `Threads`, `SigQ`, `CapInh`, `CapPrm`, `CapEff`, `CapBnd`, and `CapAmb`.

## What to do
1. Extend `procfs_status_read` (or equivalent) to emit all standard fields
2. `VmSize` / `VmRSS`: iterate the process VMA list and sum sizes
3. `Threads`: count threads in the thread group
4. `SigQ`: pending signal count / max from rlimits
5. `Cap*`: read from the process capability sets

## Acceptance criteria
- [ ] `cat /proc/self/status` includes `VmRSS`, `VmSize`, `Threads`, `SigQ`, `CapInh`, `CapPrm`, `CapEff`, `CapBnd`, `CapAmb`
- [ ] Values are numerically correct (VmRSS within 10% of actual RSS)
- [ ] Headless test parses the output and validates key fields

## Files likely touched
- `kernel/src/vfs/procfs.rs`
- `kernel/src/process/mod.rs`
- `kernel/tests/test_runner.rs`

## Notes / spec references
- `man 5 proc` — `/proc/[pid]/status`
- Linux: https://elixir.bootlin.com/linux/latest/source/fs/proc/array.c

PR #642, opened 2026-06-23
sched: per-CPU/per-priority runqueue scaffold (Perf P2 phase 1)

me1iissa:
## Perf P2 (#19) — Phase 1: behaviour-preserving per-CPU runqueue scaffold

First phase of the per-CPU/per-priority runqueue rework. **Zero functional change** — it introduces the data structure and continuously self-verifies it as a passive mirror of the authoritative ready-set, but the existing global picker still makes every scheduling decision. SMP=1 behaviour is bit-for-bit identical and the #634 anti-starvation wait-aging / BSP force-deadline fairness is untouched.

### What lands
- **`sched/percpu.rs`** — `PerCpuRq` per logical CPU (`RQS[MAX_CPUS]`), each behind its own leaf lock:
  - one FIFO per base-priority level (`PRIORITY_IDLE..=PRIORITY_MAX` ⇒ 32 levels),
  - a 32-bit non-empty bitmap for **O(1)** highest-priority lookup,
  - incrementally-maintained `nr_running`,
  - **O(1)** `highest()` pick (the primitive phases 2+ adopt).
- **`mirror_rebuild_and_verify()`** — called from the picker under `THREAD_TABLE`, rebuilds every CPU's runqueue from the locked ready-set and self-checks (1) structural invariants (bitmap/`nr_running` agree with the lists) and (2) membership (mirrored set == Ready non-idle threads), recording any break in monotone counters that must stay zero.
- **Lock order**: `THREAD_TABLE` → `RQS[cpu]` (leaf). Runqueue locks acquired in ascending CPU-index order (the order phase-3 migration will use). No path takes `THREAD_TABLE` while holding a runqueue lock.

### Verification
- **Test 645** (new, free number; master highest was 644) exercises the `PerCpuRq` API directly — FIFO ordering within a level, bitmap set/clear on level (de)population, `nr_running`, O(1) `highest()`, `PRIORITY_MAX` no-clamp, absent-dequeue no-op, invariants — and asserts the **live mirror** has recorded zero invariant/membership breaks since boot.
- test-mode boot: **Test 645 PASS**, live mirror clean (`invariant=0 membership=0`), 226 pass / 2 fail where the 2 fails are pre-existing missing-binary failures (TCC, busybox) unrelated to this change. No bugcheck/panic. The single `[SCHED/STARVE]` line is the existing #634 BSP poll-reactor force-deadline firing as designed.

### Phased plan (this is phase 1 of 5)
2. per-CPU pick (O(1) bitmap) · 3. per-CPU enqueue + wakeup-target + reschedule IPI · 4. load balancing · 5. SMP=2 default (gated on go-ahead **and** the separate TLB-timeout deferred-flush fix landing + re-verification — that fix, not this rework, is the SMP=2 *crash* root).

### DIFF
--- a/kernel/src/sched/mod.rs
+++ b/kernel/src/sched/mod.rs
@@ -10,6 +10,14 @@ use core::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use crate::proc::{self, ThreadState, THREAD_TABLE};
 use crate::arch::x86_64::apic::MAX_CPUS;
 
+/// Per-CPU / per-priority runqueue scaffold (Perf P2, phase 1).
+///
+/// Behaviour-preserving in phase 1: the structure is populated and continuously
+/// self-verified as a passive mirror of the authoritative `THREAD_TABLE`
+/// ready-set, but the authoritative picker below still makes every scheduling
+/// decision.  See [`percpu`] for the design and the phased plan.
+pub mod percpu;
+
 /// Whether the scheduler is active.
 static SCHEDULER_ACTIVE: AtomicBool = AtomicBool::new(false);
 
@@ -1460,6 +1468,17 @@ was_emergency_4k={}",
                 t.ready_since_tick = 0;
             }
         }
+
+        // ── Per-CPU runqueue mirror (Perf P2 phase 1, behaviour-preserving) ──
+        // Rebuild the per-CPU/per-priority runqueue structures from this locked
+        // snapshot of the ready-set and self-verify their bitmap/nr_running
+        // invariants and membership against the authoritative table.  This
+        // populates and continuously validates the new structure WITHOUT
+        // influencing the pick below — the authoritative picker still selects
+        // the next thread by scoring the table.  Lock order is respected
+        // (THREAD_TABLE held here; the runqueue locks are leaves taken inside
+        // `mirror_rebuild_and_verify`).  See `sched::percpu`.
+        percpu::mirror_rebuild_and_verify(&threads);
 
         // Find the highest-priority Ready thread with affinity awareness.
         // Scoring: priority * 4 + affinity_bonus (0-2)
--- a/kernel/src/sched/percpu.rs
+++ b/kernel/src/sched/percpu.rs
@@ -247,30 +247,58 @@ fn target_cpu_for(affinity: Option<u8>, last_cpu: u8) -> usize {
     if c < MAX_CPUS { c } else { 0 }
 }
 
+/// True iff `t` belongs to the mirrored non-idle runnable pool: a `Ready`
+/// thread that is NOT an idle-class thread.
+///
+/// "Idle class" matches the authoritative picker EXACTLY: `is_idle_thread =
+/// t.tid >= 0x1000` (super::schedule, the AP idle threads).  TID 0 — the BSP
+/// poll reactor — is DELIBERATELY a non-idle `PRIORITY_IDLE` peer, NOT idle, in
+/// both the picker (which warns that classifying TID 0 as idle would starve the
+/// net/x11/compositor polls) and here, so the mirror tracks the picker's
+/// non-idle pool faithfully.  Phase 2 switches the pick to read this pool;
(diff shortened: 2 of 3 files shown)


me1iissa:
## Independent review addressed (commit 2649a938)

An independent reviewer (did not write the code) returned **APPROVE-WITH-NITS**. Verdict: behaviour preservation CLEAN, lock order SAFE, structure correctness CLEAN (no off-by-one / underflow / UB), Test 645 strong. Two substantive nits — both latent traps for phase 2 — are now fixed:

1. **TID 0 classification (the merge-condition nit).** The mirror excluded TID 0, but the authoritative picker deliberately classifies TID 0 (the BSP poll reactor: net/x11/compositor) as a **non-idle** `PRIORITY_IDLE` peer — only `tid >= 0x1000` AP idle threads are idle. Since phase 2 switches the pick to read this structure, excluding TID 0 would silently drop the latency-critical reactor from the runnable set. **Fixed:** new `is_mirrored_runnable` = `Ready && tid < 0x1000`, matching the picker's `ready_peers` non-idle predicate exactly; TID 0 is now mirrored.
2. **Tautological membership counter.** The expected count was folded into the enqueue loop, so it could never disagree. **Fixed:** the expected non-idle Ready count is now derived in an **independent** pass over the table before the rebuild, making the membership check a genuine cross-check (a dropped/duplicated thread now increments the counter).

Re-verified: Test 645 PASS reproducibly, live mirror clean (`invariant=0 membership=0`) with TID 0 now in the pool, zero mirror failures anywhere in the log.

### Non-blocking nit deferred to phase 2 (reviewer agreed)
The reviewer flagged that `mirror_rebuild_and_verify` runs on the picker hot path under `THREAD_TABLE` (clears 16 runqueues + re-enqueues every Ready thread + verify), widening the most contended lock window. For a phase-1 scaffold at the **SMP=1 default** there is no contender, so it is tolerable; the first task of phase 2 will gate the rebuild behind a throttle (per-tick, or `#[cfg]` the verify out of release) before the pick switches to read this structure. Tracked.

### Note on a flaky `0x7f` bugcheck observed during local soak
One local test-mode boot hit `UNEXPECTED_KERNEL_MODE_TRAP (0x7f)` P1=1 (kernel `#DB`) on TID 0 at an unrelated test (`/dev/urandom`, Test 529) — the documented signature of the pre-existing flaky SMP=2 dead-kstack-recycle / torn-`switch_context_asm`-frame race (the `CPU_SWITCH_GEN` quiescence path, PR #582 lineage). It did **not** reproduce on the first boot of the same kernel (full clean suite, 226 pass incl. Test 645), nor on the base-commit boots, nor on the fixed-branch boot. This change is a read-only `&[Thread]` mirror that touches no thread lifecycle, kstack, or context-switch path, so it cannot produce that race; the bugcheck is unrelated flakiness, surfaced here for transparency.

me1iissa:
## ✅ CI green + flaky-bugcheck differential is dispositive

**CI: all 5 checks PASS** on commit 2649a938 — including `kernel smoke (QEMU boot + test suite)` (2m56s) and `qemu-harness smoke (tier 1)` (2m33s).

**The `0x7f` bugcheck is proven pre-existing, NOT this change.** Differential soak of the **base commit e4572817** (no mirror, my code absent) vs this branch:

| build | bugcheck | depth reached | signature |
|---|---|---|---|
| base e4572817 (no change) | **YES** | Test 529 (urandom) | `Exception #1: Debug → 0x7f UNEXPECTED_KERNEL_MODE_TRAP, CPU 0 TID 0` |
| this branch 2649a938 | yes (1 of N boots) | Test 529 | identical |
| this branch — other boots | clean (226 pass) | full suite | — |

The unmodified base reproduces the **same** `0x7f` dead-kstack-recycle `#DB` race at the **same** test, with **none of my code present**. This is the known flaky SMP=2 dead-stack-recycle / torn-`switch_context_asm`-frame race (the `CPU_SWITCH_GEN` quiescence path, PR #582 lineage), independent of the read-only `&[Thread]` runqueue mirror. Cleared.

PR #644, opened 2026-06-23
sched: incremental per-CPU runqueue maintenance + gated audit (Perf P2 phase 2a)

me1iissa:
## Perf P2 phase 2a — throttle the runqueue mirror off the picker hot path

Behaviour-preserving. The legacy table-scan picker remains authoritative; **SMP=1 selection is bit-for-bit identical**. This PR only changes how the per-CPU/per-priority runqueue *mirror* (introduced as a passive scaffold in #642) is maintained.

### Problem
Phase 1's `mirror_rebuild_and_verify` did a full O(N) clear-and-rebuild of all `MAX_CPUS * NPRIO` runqueues from `THREAD_TABLE` on **every** scheduling pass (plus an independent count pass and a per-CPU verify). That cost is acceptable for a passive scaffold but not once phase 2b reads the runqueue to make the pick.

### Change
- **`Thread::mirror_slot`** shadow (`None` | `Some((cpu, prio))`) records where each thread is currently enqueued.
- **`mirror_maintain`** walks the table once per pass and reconciles each thread's recorded slot against its desired slot. Unchanged membership ⇒ no queue mutation (**O(Δ)** on the hot path). dequeue uses the *recorded* level, so a priority change is an exact bucket move.
- **Gated audit** every `AUDIT_EVERY_PASSES = 64` passes rebuilds a throwaway image from the table and cross-checks membership + placement + bitmap + `nr_running`; divergence bumps the failure counters. This is the safety net for any off-path transition the incremental delta hasn't reflected yet.

### Intra-level order
The audit compares **order-insensitive membership** (`same_membership`), not strict FIFO order. The incremental path appends at a thread's *join moment*; a rebuild appends in *table-iteration order*. Phase 2a does not define a canonical per-level order (the legacy round-robin is rotation-over-the-table, not a persistent FIFO), so requiring strict order would false-positive. Phase 3 pins the order and tightens this back to strict equality.

### #634 preservation
Untouched — the anti-starvation wait-aging / force-deadline terms live entirely in the legacy picker's scoring, which this PR does not modify. The mirror still does not drive any decision.

### Lock order
`THREAD_TABLE -> RQS[cpu]` (leaf), acquired in ascending CPU index order. **No new lock.**

### Test
**Test 647** drives incremental-delta vs full-rebuild over 5 transition phases (build / block+migrate+boost / no-op / drain / reorder forcing join-order != table-order), asserting membership equivalence + per-rq invariants after every step, plus a positive check that the order genuinely diverges (validating the membership-equality contract). **Test 645** (live mirror health) unchanged and still passing — the gated audit keeps the failure counters at zero.

Local `test-mode` boot: 226 pass / 3 fail; the 3 failures are pre-existing data-disk artifacts (`/disk/bin/tcc` & `/disk/bin/busybox` not present, `pie_elf` exit -11) with no relation to `sched`.

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -1153,6 +1153,7 @@ pub extern "C" fn ap_rust_entry() -> ! {
             last_cpu: 0,
             first_run: false,
             ready_since_tick: 0,
+            mirror_slot: None,
             ctx_rsp_valid: alloc::boxed::Box::new(core::sync::atomic::AtomicBool::new(true)),
             clear_child_tid: 0,
             fork_user_regs: crate::proc::ForkUserRegs::default(),
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -2280,6 +2280,14 @@ fn exit_group_inner(pid: Pid, calling_tid: Tid, exit_code: i64, yield_self: bool
     if !orphan_zombie_pids.is_empty() {
         let mut threads = THREAD_TABLE.lock();
         for (_cpid, tids) in &orphan_zombie_pids {
+            // Drop any still-mirrored removed thread from the per-CPU runqueues
+            // first, so its tid does not strand in a runqueue the maintainer can
+            // never revisit (Perf P2 phase 2a — see `percpu::mirror_forget`).
+            for t in threads.iter() {
+                if tids.contains(&t.tid) {
+                    crate::sched::percpu::mirror_forget(t.mirror_slot, t.tid);
+                }
+            }
             threads.retain(|t| !tids.contains(&t.tid));
         }
         drop(threads);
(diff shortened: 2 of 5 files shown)


me1iissa:
Addressed the independent review (verdict: APPROVE-WITH-NITS, behaviour-preservation + #634-preservation both confirmed HOLD).

**NIT-1 (dangling tid on table-removal-while-mirrored)** — fixed in 9814c4bb. Added `percpu::mirror_forget(slot, tid)` and call it before removal at all four removal sites (schedule() hot reaper, `reap_dead_threads`, waitpid reap, orphan-zombie auto-reap). The schedule() reaper runs *before* `mirror_maintain` in the same pass, so this is the one that could actually strand a tid; the others are belt-and-braces. Behaviourally inert this PR but a real footgun for 2b, so closed now.

**Test-rigor gap (no proof the safety net fires)** — fixed: Test 647 step 6 injects a phantom tid and asserts the equivalence check returns Err.

**NIT-3 (stale rustdoc link)** — fixed.

**NIT-2 (off-path Ready→Blocked→Ready)** — confirmed a non-divergence (both the delta and the audit derive purely from the current table; the only divergence class was NIT-1). No action.

Re-boot: Test 645 + Test 647 (incl. negative step `injected-divergence detected`) PASS; 227 pass / 2 fail, the 2 failures are pre-existing missing-binary data-disk artifacts (`/disk/bin/tcc`, `/disk/bin/busybox`). No MIRROR/audit failures, no bugchecks.

PR #645, opened 2026-06-23
sched: per-CPU runqueue pick + pick-equivalence proof (Perf P2 phase 2b)

me1iissa:
## Perf P2 phase 2b — switch the pick onto the per-CPU runqueue (with a deterministic equivalence proof)

> **Stacked on #644 (phase 2a).** This branch includes the two 2a commits; they collapse once #644 merges. Review the top commit `c0873801` for the 2b delta.

SMP=1 stays the default and behaviour is preserved — the authoritative path is the **verbatim** legacy evaluator over the **verbatim** legacy candidate sequence.

### Design: shared verbatim evaluation, not an approximation
The legacy picker's scoring + two-pass split + force-backstop is extracted **verbatim** into a pure `select_next_core(threads, cpu, candidate-iterator, now)`. BOTH the legacy and per-CPU paths call it; they differ only in the candidate sequence:
- **legacy** — the rotated `(current_idx + i) % len` walk, passed as a **lazy iterator** (the authoritative hot path allocates nothing);
- **per-CPU** — the candidate set derived from `RQS[cpu]` (the phase-2a mirror's Ready-eligible set), re-ordered by rotation distance and with `current_idx` excluded, so the strict-max first-in-rotation tie-break matches the legacy `1..len` walk exactly.

Equivalence reduces to *"the runqueue yields the same candidate set as the table walk"*, which holds on SMP=1.

### #634 aging/force reproduced term-for-term
Because `select_next_core` **is** the original code, every #634 term is reproduced identically: `score = priority*4 + aff_bonus + wait_age_bonus(age)` (so the partial-inversion score-aging crossover that flips *before* the force deadline is exact), and the per-thread force deadline (`STARVE_FORCE_TICKS_BSP=8` for TID-0, `STARVE_FORCE_TICKS=100` otherwise) with most-overdue (`margin = age - deadline`) force-select. The `min_deadline` the design called for is this per-candidate margin; an O(1) min-of-`(ready_since+deadline)` gate to *skip* the margin scan is a pure perf refinement that doesn't change the result and is deferred to phase 3. The correctness argument here is "byte-identical evaluator", which is strictly stronger than "equivalent approximation".

### Pick-equivalence proof (mandatory gate) — Test 648
Drives BOTH selectors over **9 constructed ready-sets** and asserts identical TID: priority; affinity/last_cpu tie-break; **score-aging partial-inversion crossover before the deadline**; global force backstop; **TID-0 reactor tight deadline**; `ctx_rsp_valid==false` mid-switch skip; SMP=1 idle invariant; **rotation-tie-break sweep over every current_idx**; empty set. The rotation sweep **caught a real bug** during development (per-CPU set initially included `current_idx`, which the legacy `1..len` walk excludes) — fixed; equivalence now holds. **Test 648 PASSES.**

Optional **live cross-check** (feature `sched-pick-xcheck`, off by default, *not* in test-mode because it allocates): computes the per-CPU pick alongside the authoritative legacy pick on a sampled subset of live passes and counts divergences. A boot with it enabled ran **84s clean, 0 divergences** (force-select events firing).

### Authoritative result
The **legacy** result still drives the switch this phase; the per-CPU result is *proven equal* but not yet promoted. Promotion + the O(1) min_deadline force-gate + enqueue/wakeup-target/reschedule-IPI are phase 3.

### Guardrails
SMP=1 default + behaviour-preserving; no new lock; mm/tlb untouched; SMP=2 not enabled.

Local `test-mode` boot: 228 pass / 2 fail (645/647/648 all pass; the 2 failures are pre-existing missing-binary data-disk artifacts). Default + `firefox-test-core` builds green.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -5,6 +5,17 @@ edition.workspace = true
 
 [features]
 default = []
+# `sched-pick-xcheck` — Perf P2 phase 2b live pick-equivalence cross-check. On
+# every scheduling pass (SMP=1 only) the per-CPU runqueue pick is computed
+# alongside the authoritative legacy table-scan pick and any divergence is
+# logged + counted (`sched::percpu::PICK_DIVERGENCES`). The legacy result still
+# drives the switch, so this is observation-only — the live half of the Test
+# 648 proof. Off by default and NOT pulled into `test-mode` (it does per-sample
+# heap allocation off the schedule() path, which must stay allocation-free in
+# the default kernel); enable explicitly for a dedicated equivalence soak. The
+# deterministic Test 648 is the authoritative merge gate; this is supplementary
+# live evidence.
+sched-pick-xcheck = []
 # `heap-canary` — fence every live kernel-heap allocation with front/rear
 # magic red-zone words (validated on free) so an out-of-bounds write that
 # clobbers free-list metadata is caught at free-time, naming the victim
--- a/kernel/src/sched/mod.rs
+++ b/kernel/src/sched/mod.rs
@@ -1223,6 +1223,244 @@ pub fn test_cpu_switch_gen(cpu: usize) -> u64 {
     cpu_switch_gen(cpu)
 }
 
+/// Pure selection core: the scoring loop + two-pass split + force-backstop
+/// resolution, with NO side effects (no `READY_DEPTH` publish, no
+/// `SCHED_STARVE_FORCE_TOTAL` bump, no diagnostic).  Returns the selected table
+/// index, the run-queue depth observed, and — when the force backstop chose the
+/// pick — the forced thread's `(index, raw_wait_age)` so the caller can apply
+/// the side effects exactly once.  Extracted verbatim from the legacy in-line
+/// picker (Perf P2 phase 2b): same `score = priority*4 + aff_bonus(0..2) +
+/// wait_age_bonus(age)`, same strict-max first-in-candidate-order tie-break,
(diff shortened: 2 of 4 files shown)


me1iissa:
Independent adversarial review complete — verdict **APPROVE-WITH-NITS**, all four required properties confirmed:
- (a) extraction byte-faithful (normalized diff vs the old inline loop: only the loop-header change + side-effect hoist; scoring/skip-order/force-override reproduced exactly),
- (b) SMP=1 pick-equivalence holds,
- (c) authoritative hot path allocation-free (lazy iterator; all allocating builders are feature-gated/test-only),
- (d) #634 preserved (STARVE_* + wait_age_bonus byte-identical; cross-check is side-effect-free).

**Nit (non-blocking, addressed in 466ee52f):** Test 648 models per-CPU membership as affinity-eligibility, while the live path routes by `target_cpu_for(affinity, last_cpu)`. These coincide on SMP=1 only because `last_cpu ≡ 0` (no AP stamps it, no nonzero-affinity/idle-class BSP-eligible thread exists) — unreachable on a live SMP=1 boot, and the live `sched-pick-xcheck` cross-check would catch any real occurrence. Added a coverage note so the boundary is explicit and revisited at SMP=2 promotion. The reviewer confirmed sub-probes for current_idx-exclusion (rotation sweep), idle-pool omission, and FIFO/duplicate-score order are all covered.

Note this PR is **stacked on #644 (phase 2a)** — its commits collapse once #644 merges.

PR #647, opened 2026-06-24
sched: load-aware wakeup-target selection at the wake edge (Perf P2 phase 3b)

me1iissa:
## Perf P2 #19 — Phase 3b (stacked on #646)

> **Base:** this PR targets the Phase 3a branch (#646), not master, so the diff shows only 3b. After #646 merges I will rebase onto master.

Adds the `select_task_rq`-equivalent wakeup-target choice and resolves the Phase 2b reviewer nit (Test 648 modelled per-CPU membership as affinity-eligibility, which only coincided with the real target because `last_cpu ≡ 0` on SMP=1).

### Target decision (`wake_target_with_loads`, three tiers)
1. **Hard affinity pin** — `cpu_affinity = Some(a)` is a correctness constraint → CPU `a`.
2. **Cache-warm `last_cpu`** — kept while that runqueue is not overloaded (`nr_running < RQ_OVERLOAD_THRESHOLD = 2`).
3. **Least-loaded runqueue** — otherwise spread; ties broken toward the warm CPU (then lowest index) so a tie never forces a gratuitous migration.

The pure decision is parameterised by a load-reader so one implementation serves three callers without duplication:
- `select_wake_target` — reads live `RQS` loads (for external wake-path callers holding no rq lock).
- `mirror_maintain`'s wake edge — reads the guards it **already holds** (re-locking the leaf spinlock would self-deadlock).
- `test_wake_target` — caller-supplied loads (Test 650).

### Driving it at the wake
Rather than instrument each of the ~18 scattered Blocked/New→Ready wake sites, the target is chosen the moment a thread ENTERS the mirrored-runnable set (`mirror_slot` was `None`, now Ready non-idle) inside `mirror_maintain` — the same centralisation `ready_since_tick` stamping uses: one auditable spot, race-free under THREAD_TABLE, impossible for a new wake site to forget. The chosen CPU is baked into `last_cpu` so the **deterministic** `desired_slot` placement follows it (it does NOT recompute a load-aware target each pass — that would thrash a thread between runqueues).

### Behaviour-preserving on SMP=1
Every tier is gated on `ncpus > 1`; on a uniprocessor the wake edge is skipped, `last_cpu` is untouched, and placement is bit-for-bit identical to Phase 3a.

### Evidence
- **Test 650** (new) exercises the >1-CPU routing: hard pin / warm-cpu-kept / least-loaded fallback / warm-on-tie / strict-min migrate / uniprocessor collapse / last_cpu clamp — closing the Phase 2b coverage nit.
- Tests 645/647/648/649 still GREEN; 3b boot: 227 pass, **0 non-allowlisted failures**, 0 bugchecks.

No new locks; lock order unchanged (THREAD_TABLE → RQS[cpu] leaf; the wake edge reads loads through guards already held in ascending order).

### DIFF
--- a/kernel/src/sched/percpu.rs
+++ b/kernel/src/sched/percpu.rs
@@ -425,15 +425,19 @@ pub fn rq_nr_running(cpu: usize) -> u32 {
     RQS[cpu].lock().nr_running()
 }
 
-/// Decide which CPU's runqueue a Ready thread is mirrored onto.
+/// Decide which CPU's runqueue a Ready thread is mirrored onto — the
+/// DETERMINISTIC placement read from the thread's stored fields.
 ///
-/// Phase-1 placement policy (mirror only): a thread is mirrored onto the CPU it
-/// would run on under the *current* picker's affinity preference — its
-/// `cpu_affinity` pin if set, else the CPU it last ran on (`last_cpu`).  This
-/// keeps the mirror's membership aligned with where the authoritative picker
-/// would place the thread, so the membership cross-check is meaningful.  The
-/// real wakeup-target selection (`select_task_rq`-equivalent, with
-/// least-loaded fallback) arrives in phase 3.
+/// A pinned thread (`cpu_affinity = Some(a)`) is always placed on its pinned
+/// CPU; an unpinned thread is placed on the CPU it last ran on (`last_cpu`).
+/// This is a pure function of stored state, so the per-CPU runqueue placement is
+/// stable across maintenance passes (it does NOT recompute a load-aware target
+/// each pass, which would thrash a thread between runqueues as load fluctuates).
+///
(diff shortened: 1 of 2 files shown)


me1iissa:
CI note: this PR is stacked on #646 (base = `sched/p2-phase3a-authoritative-pick`, not master), so the kernel-smoke workflow does not auto-trigger here. It will run once #646 merges and this PR is rebased/retargeted onto master. Local validation on the 3b kernel: all Perf P2 tests (645/647/648/649/650) PASS, 0 non-allowlisted failures, 0 bugchecks, SMP=1 behaviour-preserving (every wake-target tier gated on `ncpus > 1`). Reviewer (independent): APPROVE-WITH-NITS, no blocker; the Test 650 boundary nit is addressed in eed69b66.

PR #648, opened 2026-06-24
sched/arch: reschedule IPI (vector 0xF2) for cross-CPU wakeups (Perf P2 phase 3c)

me1iissa:
## Perf P2 #19 — Phase 3c (stacked on #647 → #646)

> **Base:** stacked on the Phase 3b branch (#647). After #646/#647 merge I will rebase onto master. CI does not auto-trigger on a non-master base; local validation summary below.

Closes the remote-wakeup latency hole: before this, a thread made runnable on a different CPU was invisible to that CPU until its next ~10 ms timer tick, because every reschedule request set only the LOCAL `NEED_RESCHEDULE` flag. This is the AstryxOS analogue of `resched_curr` / `smp_send_reschedule`.

### New arch wiring
- **`apic::send_ipi_noblock`** — fire-and-forget IPI (no post-send delivery-bit spin), for the wake path which may hold a lock or run with interrupts disabled. A bounded drain of any prior in-flight IPI precedes the write (so a send isn't silently lost), with a hard spin cap so a wedged LAPIC can't hang the caller.
- **IDT[0xF2] → `irq_resched_handler` → `resched_interrupt`.** Vector **0xF2**, distinct from the TLB shootdown (0xF0) and the diagnostic DR-sync IPI (0xF1, only assigned under `w215-diag` — its vector is avoided unconditionally so the two never collide in any build; verified by building with `w215-diag`). The handler does the minimum: bump a counter, set this CPU's `NEED_RESCHEDULE`, EOI. NO lock, NO scheduler logic in interrupt context.

### New sched primitives
`RESCHED_VECTOR = 0xF2`, `resched_ipi_count()`, `set_need_reschedule_local()`, and `resched_cpu(cpu)` — arm the target's flag, then, only if `cpu` is a DIFFERENT online CPU, send the 0xF2 IPI. A self-target (always the case on SMP=1) is just the local flag, no IPI → SMP=1 unchanged.

### Driving it
`mirror_maintain` already detects the wake edge (`mirror_slot` None → enqueued). When that enqueue lands on a REMOTE runqueue, the CPU is recorded in a bitmask; the IPIs are fired AFTER the per-CPU runqueue guards drop, so no IPI is sent from inside the locked window. On SMP=1 the mask is always empty (no-op).

### Interaction with the TLB-shootdown ACK spin (ab84aaa class)
**None.** The reschedule IPI handler takes no lock and EOIs unconditionally — no shared lock to contend, no ACK to spin on; it interleaves freely with a shootdown.

### Evidence
- **Test 651** (new) pins the deterministic invariants: vector distinct from 0xF0/0xF1; `resched_cpu(self)` sends NO IPI (serviced-IPI counter unchanged); out-of-range CPU is a safe no-op; `set_need_reschedule_local` arms the flag. (Note: the kernel test phase runs with `is_active()` false, so the full `resched_cpu` active-path arming is exercised via its ungated primitive — see the test comment.)
- Tests 645/647/648/649/650 still GREEN; boot: **228 pass, 0 non-allowlisted failures, 0 bugchecks**, no spurious 0xF2. Builds clean with and without `w215-diag`.

No new locks; lock order unchanged.

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -547,6 +547,54 @@ pub fn send_ipi(dest_apic_id: u8, vector: u8) {
     while lapic_read(LAPIC_ICR_LO) & (1 << 12) != 0 {}
 }
 
+/// Send an IPI without waiting for the delivery-status bit to clear
+/// (fire-and-forget).
+///
+/// Used by the reschedule IPI (Perf P2 phase 3c): the sender only needs to
+/// kick the target CPU into noticing a freshly-runnable thread, and the target
+/// has already had its `NEED_RESCHEDULE` flag set before this call — so even if
+/// the ICR's previous send is still draining, no correctness depends on this
+/// particular write landing synchronously, and spinning on the delivery bit
+/// from inside a wake path (sometimes with interrupts disabled) would only add
+/// latency.  Per Intel SDM Vol. 3A §10.6.1 the write to ICR_LOW issues the IPI;
+/// we set ICR_HIGH (destination) first, then ICR_LOW, and return immediately.
+///
+/// A short bounded wait for any *prior* in-flight IPI to drain precedes the
+/// write so this send is not silently dropped by overwriting an ICR that has
+/// not yet latched; the bound keeps it from becoming an unbounded spin if the
+/// LAPIC is wedged.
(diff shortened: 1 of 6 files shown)


me1iissa:
Superseded — this change landed via the re-apply in #709 (7a06fe25). Closing.

PR #650, opened 2026-06-24
arch/sched: self-heal the SMP dead-CPU LAPIC-timer wedge (de-facto single core)

me1iissa:
## Problem

On SMP=2 the TID-0 BSP poll reactor logged all heartbeats on `cpu=1` and zero on `cpu=0` — SMP=2 was de-facto single core, a residual after the earlier CPU0-LAPIC-death bringup fix.

## Root cause (byte-proven, dispositive differential)

Under **KVM**, the BSP's LAPIC **periodic timer stops delivering interrupts** post-bringup (host-side injection suppression). GDB autopsy of an SMP=2 boot: `TIMER_ISR_PER_CPU[0]` is frozen at 122 for the entire run while `[1]` advances at ~100 Hz.

The differential is dispositive: the **same kernel under TCG** (no KVM) delivers the BSP timer normally — `TIMER_ISR_PER_CPU` is balanced across both CPUs (4369 vs 4405), heartbeats land on **both** cpu0 and cpu1, and the self-heal added here **never fires**. So the guest kernel is correct; the wedge is a host (KVM) artifact on the BSP vCPU.

Why the existing self-heal missed it: `irq::idle_tick` checked the **global** `TICK_COUNT` vs the TSC. `TICK_COUNT` is TSC-derived and CAS-published by *whichever* online CPU's timer ISR runs, so a single healthy sibling keeps the global clock fresh and **masks** a locally-dead timer. The afflicted CPU saw `TICK_COUNT` tracking the TSC, concluded its timer was healthy, and `hlt`'d into a sleep its own dead timer could never end.

## Fix

Two complementary **per-CPU** mechanisms (`arch/x86_64/irq.rs`):

1. **Self-check** (`cpu_timer_live`): each CPU tracks its *own* `TIMER_ISR_PER_CPU` advancement vs the TSC. On a wedge it sets a sticky dead flag, re-arms the local LAPIC, and reports unhealthy so the scheduler wait path **spins** (self-clocked off the TSC) instead of halting. Immune to the sibling-masking that defeats the global check. The pure decision logic is factored into `timer_live_decision` (unit-tested, Test 651). The re-arm retry is bounded by `MAX_FUTILE_REARMS=32`: a host that permanently suppresses injection on a vCPU will never resume no matter how the guest re-programs its LAPIC, so past the cap the CPU stops churning LAPIC MMIO and relies purely on the TSC-clocked spin + cross-CPU poke that already keeps it scheduled.

2. **Cross-CPU poke** (`poke_stale_siblings` + timer-wake IPI, **vector 0xF3**): from a live CPU's timer ISR, poke any sibling whose ISR count has gone stale so it cannot sleep through a dead timer. Self-rate-limited (≤ one poke per sibling per ~200 ms); a uniprocessor finds no siblings and does nothing.

`idle_tick` now gates its `hlt` on `cpu_timer_live` even when the global clock looks fresh; the scheduler's three `sti;hlt;cli` wait sites and the AP idle loop now use a timer-fault-tolerant `sched_wait_quantum`.

`apic::rearm_timer` was also hardened: it previously rewrote only the LVT + initial-count, which was observed **not** to resume injection on a wedged timer. It now re-programs from scratch — stop, mask, re-write the **divide-configuration** register, fresh periodic unmasked LVT, then the initial count — matching the boot-time init path. Per Intel SDM Vol. 3A §11.5.1 the counter is clocked through the divide-configuration register, and §11.5.4 mandates the mode/divide be established before the initial-count write that (re)starts the count.

## Verification

- **KVM SMP=2** (×4 boots): zero bugchecks; cpu0 schedules real work for the whole run (per-CPU current-TID advancing, watchdog 0) via the spin + cross-CPU poke, with the self-heal telemetry firing (timer-wake IPIs sent/received, re-arm count reaching the cap).
- **TCG SMP=2**: heartbeats on **both** cpu0 and cpu1, balanced per-CPU ISR counts, self-heal never fires (zero false positives).
- **SMP=1**: with a healthy timer, byte-identical to baseline (the self-heal only diverges when a timer is actually wedged).
- **test-mode**: pass/fail counts identical to `master` (the two fails are the allow-listed `tcc`/`busybox` env-gap fixtures). New Test 651 covers the decision state machine.

## Diagnostics added

kdb `cpu-state` op (per-CPU `timer_isr`/`timer_dead`/`current_tid`/`watchdog`/`nr_running` + TID-0 reactor placement + self-heal telemetry), wired through the harness allowlist; `proc::current_tid_on_cpu`/`current_pid_on_cpu` and `irq::watchdog_counter` accessors.

## Notes for the coordinator (cross-cutting)

- The timer-wake IPI uses **vector 0xF3** specifically to avoid colliding with the reschedule IPI (0xF2) on the in-flight Perf P2 phase-3c/3d branches.
- **Separate pre-existing issue, not in scope here:** SMP=1 GUI-Firefox crashes deterministically with a `#DB` at `switch_context_asm`'s `popf` (a torn `RFLAGS` with `TF=1` restored off a recycled kernel stack). This **reproduces on a clean `origin/master` build with none of this PR's changes** (bisect-confirmed) — it is the kstack-recycle / stack-canary saga and needs its own owner.
- This PR does **not** flip the SMP=2 default — that stays a separate, coordinator-sequenced change.

### DIFF
--- a/kernel/src/arch/x86_64/apic.rs
+++ b/kernel/src/arch/x86_64/apic.rs
@@ -409,11 +409,22 @@ pub fn lapic_eoi() {
 /// a single core there is no sibling, so the BSP idle path calls this to keep
 /// the timer — and therefore the scheduler tick and `TICK_COUNT` — alive.
 ///
-/// Writing the initial-count register restarts the down-counter (Intel SDM
-/// Vol. 3A §11.5.4), so a wedged periodic timer begins firing again. The LVT
-/// rewrite is belt-and-suspenders in case the vector/mode bits were lost.
-/// Safe to call from any context: it touches only this CPU's LAPIC MMIO and
-/// the value written is the immutable boot-time calibration. No-op before
+/// Re-programs the timer with the FULL from-scratch sequence rather than just
+/// rewriting the initial count: the divide-configuration register, then a fresh
+/// (masked → unmasked) LVT, then the initial count. This mirrors the boot-time
+/// `init_timer` path exactly. A minimal re-arm that rewrites only the LVT and
+/// initial-count register was observed NOT to resume injection on a wedged KVM
+/// LAPIC (the timer-ISR counter stayed frozen): per Intel SDM Vol. 3A §11.5.1
+/// the down-counter is derived through the divide-configuration register, and a
+/// counter wedged with a stale/indeterminate divisor latch does not re-latch on
+/// a bare initial-count write. The order is mandated by §11.5.4 — the LVT mode
+/// and divide configuration must be established *before* the initial-count write
+/// that (re)starts the count. The mask→unmask of LVT bit 16 forces the hardware
+/// to re-evaluate the timer LVT from a known-masked state so the subsequent
+/// initial-count write re-latches cleanly.
+///
+/// Safe to call from any context: it touches only this CPU's LAPIC MMIO and the
+/// values written are the immutable boot-time calibration. No-op before
 /// calibration (period == 0) or if the APIC is disabled.
 pub fn rearm_timer() {
     if !is_enabled() {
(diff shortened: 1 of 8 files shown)


me1iissa:
### CI status — kernel-smoke failure is inherited from `master`, not introduced by this PR

**3/5 checks green:** bootloader+kernel builds ✓, cargo check (fmt+clippy) ✓, tooling smoke ✓.

**`kernel smoke` ✗ — inherited from the base commit, with zero unexpected failures from this change:**

- This PR's `kernel smoke` failed with `[WATCH] ✗ HARD TIMEOUT after 900.1s → Exit: TIMEOUT (code 3)`. The structured allow-fail report shows **`"unallowed_failures": []`** — the only `[FAIL]` lines are the four allow-listed env-gap fixtures (`dynamic_elf`/`pie_elf`/`TCC`/`busybox`, all `InterpNotFound`/`NotFound`). No test introduced or broken by this change failed; the job timed out because the suite did not *complete* within the 900s budget (it wedges mid-suite).
- The **base commit `f4376551` (#647) already fails `kernel smoke` on `master` itself** (master CI run `28095058619`): same `HARD TIMEOUT after 900.1s`, same `"unallowed_failures": []`. The immediately-preceding master commit `035d4868` (#646) **passed** `kernel smoke`, so the hard-timeout regression entered with #647 — present with none of this PR's code.
- `qemu-harness smoke (tier 1)` was **cancelled** as a side effect of `kernel smoke` failing earlier in the same workflow run (not a real failure of that check).

**Conclusion:** the `kernel smoke` red is a pre-existing `master` infrastructure regression (a full-suite hard-timeout that landed in #647), not attributable to this fix. This PR's own contribution is verified clean: it builds, passes clippy/fmt, and adds no unallowed test failure. The master `kernel smoke` timeout should be addressed separately by the owner of the #647 series.

PR #651, opened 2026-06-24
ci/test: gate Test 198's 25k-line PNG base64 serial dump behind an opt-in feature

me1iissa:
## What

Master `kernel smoke (QEMU boot + test suite)` intermittently fails with `[WATCH] ✗ HARD TIMEOUT after 900.1s` (exit 3) and **no failing test**. This gates the deterministic half of that timeout: Test 198's host-extraction-only PNG base64 serial dump.

**CI-config / test-infra only.** No kernel behaviour changes.

## Root cause (dispositive)

The 900s watchdog only fires on *continuous* serial output for the full budget (a bugcheck exits via isa-debug-exit → caught as a test failure; a 30s output gap → caught as idle-hung). The continuous-output case is a **serial-UART saturation**:

- Test 198 ("GDI PNG screenshot") renders a 1.44 MB boot-splash PNG and base64-encodes it over COM1 as **~25,275 `[SCREENSHOT-B64:N/M]` lines**. Each is a synchronous PIO THR write (one VM-exit/byte under KVM; NS16550A model, Intel SDM Vol. 3C §25), so the dump alone costs minutes.
- Under `-smp 2` it races a faulting process's own serial spew on the other CPU, pushing the suite past 900s.

The dump is **host-extraction only** (the harness `read-png` subcommand decodes it to a host PNG); **CI never invokes `read-png`**. Test 198's pass/fail is decided by the in-guest PNG-signature check (`[SCREENSHOT] … signature_ok=true`) emitted **before** the dump — so gating it off changes no test outcome.

## Change

- `kernel/Cargo.toml`: new leaf feature `screenshot-b64-dump`, **off by default**, deliberately **not** pulled into `test-mode` (mirrors the existing `test-mode-trace` split-out).
- `kernel/src/test_runner.rs`: Test 198's step-7 base64 dump is now `#[cfg(feature = "screenshot-b64-dump")]`; when off it emits a single `[SCREENSHOT-B64-DISABLED]` sentinel. Steps 1–6 (render, signature check, `signature_ok=true`, `test_pass!`) are unchanged.
- `scripts/qemu-harness.py`: `start --screenshot-dump` appends the feature (expansion printed to stderr, never silent — mirrors `--trace`); `read-png` now fast-fails on the sentinel with `{"ok":false,"error":"screenshot_dump_disabled"}` rc=1 instead of waiting out its timeout.
- `scripts/qemu-harness-smoke.py`: new host-only `run_read_png_disabled_check` (in the `--staleness-only` CI subset) pins the fast-fail contract.

## Verification

- Host-only smoke (`qemu-harness-smoke.py --staleness-only`) green, including the new 4/4 read-png check.
- Live TCG `-smp 2` boot of the gated `--features test-mode` kernel: Test 198 emits `signature_ok=true` (PASS) **and** the `[SCREENSHOT-B64-DISABLED]` sentinel; **0 `[SCREENSHOT-B64:…]` chunk lines** (dump fully eliminated).
- Independent review: APPROVE (correctness-preservation, no-collateral, feature-hygiene, harness-additivity all PASS). The Firefox `/tmp/out.png` path is a distinct `[FF-OUT-PNG:…]` emitter and is unaffected.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -38,6 +38,21 @@ heap-canary = []
 # so on the hot path the spew dominates wall-clock.
 test-mode-trace = []
 test-mode = ["heap-canary", "test-mode-trace"]
+# `screenshot-b64-dump` — opt-in for Test 198 (GDI PNG screenshot) to base64-
+# encode the rendered 1.44 MB boot-splash PNG over COM1 as the chunked
+# `[SCREENSHOT-B64:N/M]` stream (~25k lines) that the harness `read-png`
+# subcommand decodes back to a host PNG.  Off by default and DELIBERATELY NOT
+# pulled into `test-mode`: the dump is host-EXTRACTION only — Test 198's
+# pass/fail is decided by the in-guest PNG-signature check (`signature_ok=true`)
+# BEFORE the dump, so gating the dump off changes NO test outcome, only what is
+# transcribed to COM1.  CI never invokes `read-png`, so for CI the dump is pure
+# UART saturation: at ~25k synchronous PIO THR-write lines (Intel SDM Vol. 3C
+# §25 I/O-instruction VM-exits; NS16550A THR-write model) it dominates the
+# kernel-smoke wall-clock and, racing a faulting process's serial spew under
+# `-smp 2`, can push the suite past the 900 s watchdog.  Enable only when a
+# human/agent actually wants the boot-splash PNG extracted (the harness sets it
+# on the `read-png` build path).
+screenshot-b64-dump = []
 gui-test = []
 # `firefox-test-core` — the FUNCTIONAL behaviour the Firefox bring-up path
 # needs to RUN: heap-canary fencing, the FF workload selection + launch in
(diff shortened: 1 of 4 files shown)


me1iissa:
Added a second commit `afd5754`: **harness/ci-run zombie-aware pid check** (the small follow-up). `ci-run`'s suite-wait loop used `_pid_alive`, which reports a `<defunct>` QEMU (exited via isa-debug-exit on a mid-suite bugcheck, not yet reaped) as alive — so it spun out the full timeout instead of returning promptly. New `_pid_running` reads `/proc/<pid>/stat` state and rejects `Z`; host-only smoke forks a real zombie to pin the contract. Both commits are CI/harness-infra only; host-only smoke green (incl. 2 new checks).

me1iissa:
Superseded — its two commits (Test-198 dump gate + ci-run zombie fix) were cherry-picked into #652 and merged there together with the mirror_maintain overflow fix, to clear the chicken-and-egg where neither PR fixed both CI failure modes alone.

PR #671, opened 2026-06-29
drivers: virtio-input (modern transport) — VNC-drivable absolute pointer

me1iissa:
## Problem

VNC sends **absolute** pointer events, but the guest's only pointing device was the **relative-only PS/2 mouse**. There was no absolute pointing device, so `move-abs` had no handler and a VNC client's cursor could not move the guest cursor.

## Approach — virtio-input / virtio-tablet (modern transport)

A **virtio-tablet** reports absolute coordinates (`EV_ABS`/`ABS_X`/`ABS_Y`) and maps directly onto the VNC/QMP absolute-pointer model — far less work than reviving the dead USB-HID stack, and it rides the virtio infrastructure we already understand.

The catch: virtio-input is a **modern-only** device (PCI id `0x1052`, no legacy id), so the existing legacy port-I/O virtio drivers (blk/serial/net) can't be reused. This PR implements the slice of the **virtio 1.x modern PCI transport** an input device needs:

- Walk the PCI capability list for the vendor virtio caps (COMMON/NOTIFY/ISR/DEVICE cfg), each naming BAR+offset+length (virtio 1.2 §4.1.4).
- Map those MMIO windows uncached (same scheme as the LAPIC MMIO map), handling 64-bit BARs.
- Common-config bring-up: feature negotiation (`VIRTIO_F_VERSION_1`), `FEATURES_OK` handshake, per-queue desc/avail/used addresses (§4.1.4.3).
- One split virtqueue — the **event queue** — filled with device-writable 8-byte `virtio_input_event` buffers (§2.7, §5.8).
- Device-config queries (`EV_BITS` to confirm a pointer; `ABS_INFO` for the absolute axis range) for coordinate scaling.

Events (Linux evdev `EV_ABS`/`EV_KEY`/`EV_SYN`, little-endian) are **polled** from the GUI input pump once per compositor tick and committed to the mouse state on `SYN_REPORT`, scaling the device range onto the framebuffer. This reuses the existing cursor / X11 pointer-injection sink unchanged; **PS/2 stays as a relative fallback**. The device INTx is masked (`VRING_AVAIL_F_NO_INTERRUPT` + PCI Interrupt Disable) since the queue is polled.

## Harness

- `astryx_qemu`: attach `-device virtio-tablet-pci` on the GUI/VNC paths, **pinned to a fixed PCI slot** — QEMU auto-slot assignment is non-deterministic across device sets and intermittently placed the device where the guest enumerator missed it (same fix pattern as the boot disk's pinned addr).
- `qemu-harness`: new `input` subcommand (`input <sid> --frac FX FY [--click left]` / `--abs-x/--abs-y` / `--button --down/--up`) injecting QMP `input-send-event`; plus a VNC-aware guard so `ASTRYX_VNC` + `--ff-gui` don't emit a duplicate `-vga`.

## Verification (over the QMP absolute-pointer path that VNC drives)

Driver bring-up: `[PCI] 00:0a.0 1af4:1052` → `[VIRTIO-INPUT] claimed absolute (tablet) pointer: abs_max=(32767,32767) qs=64`.

Cursor tracks the absolute pointer — `kdb mouse-state` after `input-send-event` move-abs, all matching `abs*(screen-1)/0x7fff` on a 1920×1080 framebuffer:

| inject | expected | observed |
|---|---|---|
| baseline | center | (960, 540) |
| frac 0.15, 0.20 | (287, 215) | **(287, 215)** |
| frac 0.85, 0.85 | (1631, 917) | **(1631, 917)** |

**Screendump diff** between the two positions isolated exactly **122 differing pixels in two clusters** — a 12×12 cursor arrow at bbox `(287,215)–(295,226)` and `(1631,917)–(1639,928)`, i.e. only the cursor moved, to exactly the injected locations.

Button: `BTN_LEFT` press/release toggled the button state `0 → 1 → 0`.

## Scope / known gap

The full *click-navigates-example.com* over a windowed-Firefox session is gated by the pre-existing FF SMP=1 BSP-starvation issue (content-init wedge), which halts the GUI input pump after Firefox loads — independent of this driver (the driver and cursor pipeline are proven correct in the xeyes/GUI path above). Button-state propagation through the driver is proven; X11 `ButtonPress` delivery is the existing, PS/2-shared sink.

## References

OASIS virtio 1.2 (§2.7 split virtqueues, §4.1.4 modern PCI layout, §5.8 input device); Linux evdev event codes (`EV_*`/`ABS_*`/`BTN_*`); PCI Local Bus Spec 3.0 §6.2.2/§6.7.

### DIFF
--- a/kernel/src/drivers/mod.rs
+++ b/kernel/src/drivers/mod.rs
@@ -20,6 +20,7 @@ pub mod pty;
 pub mod tty;
 pub mod usb;
 pub mod virtio_blk;
+pub mod virtio_input;
 #[cfg(feature = "qga")]
 pub mod virtio_serial;
 pub mod vmware_svga;
@@ -52,6 +53,12 @@ pub fn init(boot_info: &BootInfo) {
     usb::init();
     if virtio_blk::init() {
         crate::serial_println!("[DRIVERS] Virtio-blk initialized");
+    }
+    // virtio-input (modern transport): claims a virtio-tablet/-mouse PCI device
+    // to drive the cursor from an *absolute* pointer (what VNC delivers).  Must
+    // run after pci::init() above.  Polled from gui::input::pump_input.
+    if virtio_input::init() {
+        crate::serial_println!("[DRIVERS] Virtio-input initialized (absolute pointer)");
     }
     #[cfg(feature = "qga")]
     {
--- a/kernel/src/drivers/mouse.rs
+++ b/kernel/src/drivers/mouse.rs
@@ -194,6 +194,34 @@ pub fn is_initialized() -> bool {
     MOUSE_INITIALIZED.load(Ordering::Relaxed)
 }
 
+/// Current screen bounds (width, height) used for cursor clamping.  Exposed so
+/// an absolute pointing device (virtio-tablet) can scale its device-space
+/// coordinates onto the framebuffer.
+pub fn screen_bounds() -> (i32, i32) {
+    (
+        SCREEN_WIDTH.load(Ordering::Relaxed),
+        SCREEN_HEIGHT.load(Ordering::Relaxed),
+    )
+}
+
+/// Set the full pointer state (absolute screen position + button bitmask) in a
+/// single update.  Used by absolute pointing devices (e.g. virtio-tablet) whose
+/// hardware already reports an absolute location, unlike the PS/2 mouse which
+/// integrates relative deltas in [`handle_irq`].  Coordinates are clamped to the
+/// screen.  `buttons` follows the same bit layout as [`buttons`] (bit 0=left,
+/// 1=right, 2=middle).
+///
+/// Marking the mouse initialized here lets a virtio-tablet drive the cursor even
+/// on a guest where the PS/2 controller never came up.
+pub fn set_state(x: i32, y: i32, buttons: u8) {
+    let max_x = SCREEN_WIDTH.load(Ordering::Relaxed) - 1;
+    let max_y = SCREEN_HEIGHT.load(Ordering::Relaxed) - 1;
+    MOUSE_X.store(x.clamp(0, max_x), Ordering::Relaxed);
+    MOUSE_Y.store(y.clamp(0, max_y), Ordering::Relaxed);
+    MOUSE_BUTTONS.store(buttons, Ordering::Relaxed);
+    MOUSE_INITIALIZED.store(true, Ordering::Relaxed);
+}
+
 /// Best-effort read of the live i8042 controller command byte (CCB), for the
 /// `mouse-state` kdb diagnostic.
 ///
--- a/kernel/src/drivers/virtio_input.rs
+++ b/kernel/src/drivers/virtio_input.rs
@@ -1,0 +1,765 @@
+//! Virtio-input PCI Driver (Modern / virtio-1.0 transport) — absolute pointer
+//!
+//! Brings up a **virtio-input** device (QEMU `-device virtio-tablet-pci`,
+//! `virtio-mouse-pci`, …) and feeds its pointer events into the kernel mouse
+//! state so the in-kernel X11 server / GUI compositor cursor tracks an
+//! absolute pointing device.  This is what a VNC client needs: VNC delivers
+//! *absolute* pointer coordinates, and a virtio-tablet exposes exactly that
+//! (`EV_ABS`/`ABS_X`/`ABS_Y`), unlike the relative-only PS/2 mouse.
+//!
+//! # Why a new transport
+//!
+//! virtio-input has no legacy/transitional PCI device id — it is a
+//! **modern-only** device (PCI device id 0x1052 = 0x1040 + VIRTIO_ID_INPUT(18)).
+//! The existing `virtio_blk` / `virtio_serial` / `virtio_net` drivers all use
+//! the legacy BAR0 port-I/O register window, which a modern-only device does
+//! not provide.  This driver therefore implements the small slice of the
+//! **virtio 1.x modern PCI transport** that an input device needs:
+//!
+//!   * Walk the PCI capability list for the vendor-specific virtio caps
+//!     (`VIRTIO_PCI_CAP_COMMON_CFG`/`NOTIFY_CFG`/`ISR_CFG`/`DEVICE_CFG`),
+//!     each of which names a BAR + offset + length (virtio 1.2 §4.1.4).
+//!   * Map those MMIO windows uncached (mirrors `apic::init`'s LAPIC map).
+//!   * Drive the common-config register block (device/driver feature select,
+//!     status, per-queue desc/avail/used addresses) (virtio 1.2 §4.1.4.3).
+//!   * Set up one split virtqueue — the **event queue** (queue 0) — filled
+//!     with device-writable 8-byte `virtio_input_event` buffers
+//!     (virtio 1.2 §5.8: Input Device).
+//!
+//! # Event flow (polled)
+//!
+//! The device writes Linux-evdev `input_event` records (type/code/value, all
+//! little-endian) into the event queue as the host pointer moves.  We drain
+//! the used ring from [`poll`], which the GUI input pump calls once per tick
+//! (`gui::input::pump_input`).  Polling at the compositor cadence (~50 Hz) is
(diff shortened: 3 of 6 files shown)


me1iissa:
VERDICT: APPROVE (rebase + CI re-run required before merge; no code-level blockers)

Independent adversarial review of the modern virtio-1.x transport, DMA/buffer lifecycle, the (absent) interrupt path, staleness vs current master, input-event semantics, and integration. I read the driver, its kernel integration points, and a mainstream virtio-pci-modern + virtio-input reference, then judged against the OASIS Virtio v1.2 spec and the published Linux evdev event-code interface. **Code-only review — no boot run** (justification at the end).

## Transport correctness — spec-clean

Cross-checked every common-config offset and handshake step against OASIS Virtio 1.2 §4.1.4.3 / §3.1.1:

- **Status handshake** (`bringup`, virtio_input.rs:487-621) is correctly ordered: reset→spin-until-0→ACKNOWLEDGE→DRIVER→(feature negotiate)→FEATURES_OK→(read-back confirm)→queue setup→DRIVER_OK, with STATUS_FAILED written on every abort. Matches §3.1.1 exactly.
- **Feature negotiation** (virtio_input.rs:499-511): requires VIRTIO_F_VERSION_1 (bit 32 → feature-select word 1, bit 0), refuses all device-specific bits, writes both driver-feature words. Correct for a device that needs no device-specific features.
- **Common-cfg register map** (virtio_input.rs:85-99): device_status@0x14, queue_select@0x16, queue_size@0x18, queue_enable@0x1C, queue_notify_off@0x1E, queue_desc@0x20, queue_avail@0x28, queue_used@0x30 — all match §4.1.4.3. The `w16` write to queue_size at virtio_input.rs:594 with its comment about not clobbering the adjacent queue_msix_vector (@0x1A) is exactly right.
- **Split-vring layout** (virtio_input.rs:198-286): descriptor 16-byte / avail 2-byte / used 4-byte alignment all satisfied (used/evbuf are over-aligned to 4096 — harmless under §2.7 modern rules, which unlike legacy impose no page alignment). Descriptor (addr/len/flags/next), avail (flags,idx,ring@+4), and used-elem (`id` le32 first) structures are all correct.
- **Notification** (virtio_input.rs:622-623, 719-722): writes to notify_base + queue_notify_off*notify_off_multiplier per §4.1.4.4, with the 16-bit virtqueue index (0) as the payload — correct, and correct even when notify_off_multiplier is 0.
- **Endianness**: little-endian throughout; `virtio_input_event` decode (type le16 / code le16 / value le32, virtio_input.rs:673-677) and the ABS_INFO `max` read at device_cfg+12 (union+4, virtio_input.rs:540/543) match §5.8.

The BAR decode (virtio_input.rs:384-393) correctly handles 64-bit memory BARs; I verified `PciDevice.bar[]` (pci.rs:82-85) stores **raw, unmasked** register values, so the `lo & 0x6 == 0x4` 64-bit-type test is valid.

## Memory / DMA safety — sound

- Queue memory is one physically-contiguous allocation (`pmm::alloc_pages`, virtio_input.rs:570) accessed through the higher-half direct map; x86 PCIe DMA is cache-coherent, and a SeqCst fence orders the avail-ring slot write before the `avail.idx` bump (virtio_input.rs:262-270).
- **No use-after-free surface**: the device is a permanent singleton (held in a static, never freed), so the virtio-blk #568/#601 "device DMAs into a recycled frame" class cannot occur — there is no free path. Buffers are correctly re-armed after consumption (fill_desc+publish, virtio_input.rs:714-715).
- Every fallible init step precedes the allocation, so no failure path leaks pages.

## Interrupt path — N/A by design, correctly quiesced

This is a **polled** driver: INTx is masked via PCI command-register Interrupt Disable bit 10 (virtio_input.rs:463-472) *and* VRING_AVAIL_F_NO_INTERRUPT (virtio_input.rs:607), and the ISR cap is intentionally unused. No unrouted GSI is ever asserted. The historical lock-across-IRQ deadlock class (there is no ISR here) does not apply.

## SMP safety — holds under real SMP=2

`poll()` takes a `spin::Mutex` only across atomic-only work: `commit()` → `mouse::set_state` / `screen_bounds` / `position` are all pure atomic loads/stores (mouse.rs:150-164, 165-167) — no schedule, no nested lock. It runs solely from GUI/main-thread loops (never IRQ context), and the lock is released before `pump_input` takes `INPUT_STATE` (gui/input.rs). The lock-held-across-dispatch deadlock lineage does not apply.

## Input-event semantics — correct

EV_ABS accumulate + commit-on-SYN_REPORT, EV_KEY→button bitmask (BTN_LEFT/RIGHT/MIDDLE), abs→screen scaling `abs*(screen-1)/abs_max` (virtio_input.rs:736-759). Event-queue overflow degrades gracefully (a slow 50 Hz poll drops intermediate cursor positions, never blocks) — acceptable for a pointer. SCREEN_WIDTH/HEIGHT default to 1024/768 (mouse.rs:27-28), so the `-1` clamp cannot underflow even pre-GUI-init.

## Integration — well-guarded, narrow blast radius

`drivers::init` calls `virtio_input::init()` unconditionally, but on any boot with no virtio-input device on the bus it scans PCI, matches nothing, and returns false (benign no-op; AVAILABLE stays false, poll() early-returns). The tablet is only attached by `astryx_qemu._input_args` on the ASTRYX_VNC / show_window / gui-test / firefox-test paths; the harness `start` entrypoint is hard-wired to `mode="test"`, so standard harness and headless test_runner boots get **no** tablet. This confines the change to the GUI/VNC/FF-GUI paths. The `-vga` duplicate guard (only skips injection under ASTRYX_VNC) and the QMP `input` subcommand are additive and non-interactive.

## Findings (all non-blocking)

1. **MMIO mapping leak + force-FAILED for non-pointer virtio-input devices** (virtio_input.rs:474-485 vs 528-532): `bringup` maps 3 MMIO windows and completes the reset/FEATURES_OK handshake *before* the EV_BITS pointer check, so a virtio-keyboard is left in STATUS_FAILED with 3 leaked higher-half mappings. Benign today (no virtio-keyboard is attached; PS/2 keyboard is independent), but if a guest ever presents keyboard+tablet the keyboard is force-failed. Prefer identifying the device from config space before the destructive handshake, or unmap on the non-pointer path.
2. **u8 capability-offset arithmetic** (virtio_input.rs:356-367): `off + 16` is u8 arithmetic and would overflow-panic in a debug build if a vendor cap sat above 0xEF. QEMU virtio caps live ~0x40-0xB0 so it never triggers, but widening to `usize` removes the latent panic.
3. **config_generation not re-checked** around the multi-field ABS_INFO reads (virtio_input.rs:538-544). Theoretical for a static tablet range; cosmetic per §4.1.4.3.1.
4. **Used-elem written-length ignored** (virtio_input.rs:673-677): always reads a full 8-byte event. Correct in practice (virtio-input always writes 8 bytes); a `len >= 8` guard would harden against a malformed device. Cosmetic.

## Rebase / merge gating

The branch is at base caddad9b; master is a099a254 (+#672-#719). I verified every consumed API is byte-identical on current master: `pci_config_read32/write32`, `devices()`, `enable_bus_master`, `PciDevice.bar[6]`, `vmm::map_page` + the PAGE_* flags, `pmm::alloc_pages`, and `KERNEL_VIRT_BASE`. It should rebase cleanly and compile. **The merge should be gated on a rebased, green CI run** — that is the real integration test (particularly any FF-GUI/VNC job, the only paths that attach the tablet); a manual VNC cursor sanity check is worthwhile if convenient.

**Why no boot:** it would not be dispositive here — the worktree isn't rebased onto #672-#719 (so a boot tests the old tree, not the integration in question), the transport question is fully answerable from code + spec, and two live sessions were already running an unrelated investigation at review time. CI-after-rebase is the correct integration gate.

Spec references: OASIS Virtio v1.2 §2.7 (split virtqueues), §3.1.1 (device init), §4.1.4 (modern PCI transport), §5.8 (input device); the published Linux evdev event-code interface (EV_*/ABS_*/BTN_*); PCI Local Bus Spec 3.0 §6.2.2 (command register).

me1iissa:
Superseded by the rebased successor PR (same code, byte-identical diff per range-diff; commit-metadata hygiene fixed; review verdict carries). Closing in favor of the successor.

PR #681, opened 2026-07-01
net/unix: re-land peer-sever at teardown — stop slot-recycle cross-channel bleed (was #549)

me1iissa:
## Re-land — NDE

This change was originally merged as **PR #549** ("Gate-1 Finding A") on the
pre-rewrite history. The 2026-06-10 history rewrite orphaned that lineage and
this fix was never carried onto the rewritten `master`. The bug is live again
at HEAD: `kernel/src/net/unix.rs` has no `rx_eof`, no `peer_closed`, and
dereferences `peer_id` raw after teardown. This PR re-implements the fix,
reconciled with three weeks of drift (SEQPACKET slot-ring, SCM changes,
NDE-24 shutdown/backpressure work, and the ring-after-lock-drop poll-bell
discipline — all preserved). The removed kdb `unix-diag` surface is *not*
re-added (it no longer exists in the tree).

## The bug — slot-recycle cross-channel bleed

The AF_UNIX socket table recycles slot indices the instant a slot's last open
file description is closed. Teardown left the **surviving** endpoint's
`peer_id` pointing at the freed index. Once that index was re-allocated to an
unrelated connection, three things went wrong through the dangling back-pointer:

1. **`write(2)` bled into a foreign channel.** The stale survivor's writes
   pushed bytes into the *new* connection's receive ring (silent cross-channel
   stream corruption), and reported success where POSIX `send(2)`/`write(2)`
   on a peer-gone stream socket require `-EPIPE`.
2. **`shutdown(SHUT_WR)`/`close(2)` flipped EOF on a live stream.** Half-close
   state landed on the unrelated connection, delivering a spurious orderly
   hang-up that `poll(2)`/`recv(2)` observe as an EOF that never happened.
3. **Queued bytes were discarded at peer close.** The peer-close path marked
   the survivor `shut_rd`, which EOFs immediately — but `recv(2)` requires
   already-queued data to be returned *before* the 0-byte orderly-shutdown
   indication.

Under the multi-process Firefox launch this cascaded: a child holds AF_UNIX
IPC ends across a window where earlier short-lived connections close and their
indices recycle; one stale write/half-close on a fresh channel kills the
child's IPC handshake, the parent respawns, and the dead child's slots recycle
again.

## The fix (all under the existing `TABLE` lock)

- **`close()`** — mutual-pairing gate (`peer.peer_id == id`) before touching
  the peer; mark the survivor `rx_eof` + `peer_closed` and **sever** its
  `peer_id` to `u64::MAX` so no later path can dereference the recyclable index.
- **`shutdown(SHUT_WR)`** — same mutual-pairing gate; sets the peer's `rx_eof`
  (drain-then-EOF) rather than `shut_rd`.
- **`write()`** — mutual-pairing gate → `-EPIPE` on a stale/foreign peer;
  `peer_closed` → `-EPIPE` fast path.
- **`read_msg()`** — empty ring + `rx_eof` → orderly EOF, but queued bytes are
  always delivered first (`recv(2)` drain-then-EOF). Local `SHUT_RD` keeps its
  immediate-EOF semantics (IEEE 1003.1 `shutdown`).
- **`read_shutdown()`/`fully_hung_up()`/`writable()`** — `rx_eof`/`peer_closed`
  drive the `POLLIN`/`POLLRDHUP`/`POLLHUP`/`POLLOUT` edges with timing
  identical to the pre-fix peer-`Free` probes that severing makes unreachable,
  so the `PollBellSource::UnixShutdown` wake edges the poll/epoll layer depends
  on are unchanged.

## Recycle-safety invariant

Every dereference of `peer_id` is now provably safe against recycling:
severed to `u64::MAX` at teardown (fails every bounds check) **and** gated on
mutual pairing (`peer.state == Connected && peer.peer_id == id`) at each write/
shutdown/close site.

## Test (test-mode)

`test_unix_peer_close_sever_and_recycle` proves: (1) drain-then-EOF (queued
bytes then 0), (2) `-EPIPE` on write after peer close, (3) the survivor's
`POLLIN`/`POLLRDHUP` + `POLLHUP` + writable edges, and (4) recycle isolation —
after the dead peer's slot index is re-allocated to a fresh pair, a stale
`write`/`shutdown`/`close` on the survivor injects no bytes and flips no EOF
state on the new pair, which stays fully functional.

## Build

Both feature sets compile clean via `scripts/qemu-harness.py build`:
`test-mode` (`ok:true`) and `firefox-test-core` (`ok:true`). No local QEMU
booted this change — the in-kernel test suite runs under CI.

## Spec

POSIX.1-2017: `write(2)`, `send(2)`, `recv(2)`, `shutdown(2)`, `poll(2)`;
`man 7 unix`; `man 7 epoll` (`EPOLLRDHUP` vs `EPOLLHUP`).

### DIFF
--- a/kernel/src/net/unix.rs
+++ b/kernel/src/net/unix.rs
@@ -99,6 +99,21 @@ struct UnixSocket {
     /// the in-memory pipe.
     shut_rd:     bool,
     shut_wr:     bool,
+    /// "No more bytes will ever arrive" — set on this socket when its peer
+    /// performs `shutdown(SHUT_WR)` or fully closes.  Distinct from the local
+    /// `shut_rd` (SHUT_RD discards queued data and EOFs immediately per
+    /// IEEE 1003.1 §shutdown): with `rx_eof` the reader first DRAINS any
+    /// bytes already queued in `recv_buf` and only then observes the orderly
+    /// EOF — the recv(2) contract for a peer that performed an orderly
+    /// shutdown ("return 0 once all queued data is consumed").  Feeds the
+    /// `POLLIN`/`POLLRDHUP` readiness edges via [`read_shutdown`].
+    rx_eof:      bool,
+    /// The peer endpoint has FULLY closed (last open file description gone).
+    /// Set together with severing our `peer_id` back-pointer in [`close`], so
+    /// no code path can ever dereference the freed — and possibly RECYCLED —
+    /// peer slot through us.  Drives `POLLHUP` ([`fully_hung_up`]), the
+    /// always-writable EPIPE fast-path ([`writable`]), and write(2) → -EPIPE.
+    peer_closed: bool,
     /// Count of open file-description references to this socket slot.
     ///
     /// `socket(2)` / `socketpair(2)` initialise this to 1.  Every call that
@@ -148,6 +163,8 @@ impl UnixSocket {
             backlog_len: 0,
             shut_rd:     false,
             shut_wr:     false,
+            rx_eof:      false,
+            peer_closed: false,
             ref_count:   0,
             creator_pid: 0,
             creator_uid: 0,
@@ -169,6 +186,8 @@ impl UnixSocket {
         self.backlog_len = 0;
         self.shut_rd     = false;
         self.shut_wr     = false;
+        self.rx_eof      = false;
+        self.peer_closed = false;
         self.ref_count   = 0;
         self.creator_pid = 0;
         self.creator_uid = 0;
@@ -511,9 +530,28 @@ pub fn write(id: u64, data: &[u8]) -> i64 {
         if s.state != UnixState::Connected { return -32; }
         // SHUT_WR locally → -EPIPE per IEEE 1003.1 §shutdown.
         if s.shut_wr { return -32; }
+        // Peer fully closed → -EPIPE per POSIX write(2)/send(2) on a
+        // stream socket whose peer is gone.  (`peer_id` is severed to
+        // u64::MAX at peer teardown, so the bounds check below also
+        // catches this; the explicit flag keeps the intent readable.)
+        if s.peer_closed { return -32; }
         (s.peer_id, s.kind)
     };
     if peer_id as usize >= MAX_UNIX_SOCKETS { return -32; }
+    // Mutual-pairing gate: only deliver if the resolved peer slot still
+    // points back at US.  The socket table recycles slot indices the moment
+    // a slot's last reference is closed; a stale `peer_id` held across that
+    // recycling would otherwise inject this stream's bytes into an UNRELATED
+    // connection's receive ring (cross-channel corruption).  A connected
+    // pair is mutual by construction (socketpair(2)/connect(2)), so a
+    // mismatch can only mean "freed and possibly recycled" → the write must
+    // observe -EPIPE, exactly as if the peer had closed (POSIX send(2)).
+    {
+        let peer = &t.0[peer_id as usize];
+        if peer.state != UnixState::Connected || peer.peer_id != id {
+            return -32;
+        }
+    }
 
     if kind == SockKind::SeqPacket {
         // SEQPACKET: an entire message must fit in the receiver's ring AND
@@ -601,7 +639,14 @@ pub fn read_msg(id: u64, buf: &mut [u8]) -> Result<(usize, usize), i64> {
     if s.kind == SockKind::SeqPacket {
         // SEQPACKET: dequeue exactly one message, truncating any tail that
         // does not fit per `man 7 unix` §SOCK_SEQPACKET.
-        if s.seq_head == s.seq_tail { return Err(-11); } // EAGAIN — no message
+        if s.seq_head == s.seq_tail {
+            // No queued message.  If the peer's write side is gone
+            // (orderly shutdown or full close), report EOF — but only
+            // AFTER the queue is empty: recv(2) requires queued data to
+            // be returned before the 0-byte orderly-shutdown indication.
+            if s.rx_eof { return Ok((0, 0)); }
+            return Err(-11); // EAGAIN — no message
+        }
         let msg_len = s.seq_lens[s.seq_head] as usize;
         s.seq_head = (s.seq_head + 1) % SEQ_QUEUE_CAP;
 
@@ -621,7 +666,13 @@ pub fn read_msg(id: u64, buf: &mut [u8]) -> Result<(usize, usize), i64> {
         result  = Ok((copied, discard));
     } else {
         // STREAM: byte-stream — copy as many bytes as fit, no boundaries.
-        if s.recv_available() == 0 { return Err(-11); }
+        if s.recv_available() == 0 {
+            // Drain-then-EOF (recv(2)): once the ring is empty AND the peer
+            // can never push more bytes (peer SHUT_WR or peer fully closed),
+            // the read observes the orderly EOF.  Queued bytes always win.
+            if s.rx_eof { return Ok((0, 0)); }
+            return Err(-11);
+        }
         let copied = s.pop(buf);
         drained = copied;
         result  = Ok((copied, 0));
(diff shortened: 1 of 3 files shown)


me1iissa (commented):
## Independent pre-merge review — NDE

**Verdict: APPROVE-WITH-NITS.** The core recycle-safety fix is correct, complete
for every *mutating* `peer_id` dereference, POSIX-faithful, and wake-edge
preserving. Two non-blocking follow-ups and one defense-in-depth note below.
I diffed the re-land against the orphaned original, enumerated every `peer_id`
consumer in `net/unix.rs` (plus the raw `get_peer` consumers in the SCM send
path and the X11 reaper), and walked the concurrent/SMP teardown ordering.

---

### 1. Recycle safety — is it total? Mostly yes (mutating paths: yes).

**Every mutating deref is both severed-guarded (`u64::MAX` fails bounds) and
mutual-pairing-gated. Verified safe:**

- `write()` — the gate (`net/unix.rs:549-554`, `peer.state==Connected &&
  peer.peer_id==id`) sits *before both* transport arms, so it covers the
  SEQPACKET push (`:561`) and the STREAM push (`:582`). `peer_closed` fast-path
  at `:537`. Good.
- `shutdown(SHUT_WR)` — gate at `:721` before the only peer write (`p.rx_eof`).
- `close()` — gate at `:804` before the sever/flag writes (`:814-816`).

**Read-only predicate derefs are NOT mutual-pairing-gated** (`fully_hung_up`
`:908` `t.0[peer].state`; `writable` `:943`; `peer_creds` `:1053`). They only
test `peer != u64::MAX`, then rely on the sever-invariant to guarantee the
index is either MAX or a genuine mutual peer. Because they are read-only they
cannot corrupt or bleed — worst case is a wrong readiness/creds answer *if* the
invariant were ever violated. Under this PR the invariant holds (severing is
mutual at every teardown), so this is a defense-in-depth note, not a defect.

**SCM-attach path — recycle-safe, but newly exposes a resource leak.** The SCM
send path resolves the receiver with the *raw* `get_peer()`
(`subsys/linux/syscall.rs:2263-2265`), not the mutual gate, then
`scm_queue(receiver_id=peer, …)` at `:2459`. This is still recycle-safe for
delivery: (a) after a mutual sever `get_peer()` returns `u64::MAX`, so
`scm_queue` is skipped (`:2426`); and (b) `close()` calls
`scm_drain_receiver(id)` (`net/unix.rs:830`) which drops any batch keyed to a
closing slot *before* the index can recycle — so no fd batch can be
misdelivered to a fresh connection at the reused index. No cross-channel bleed.

> **NIT-1 (follow-up, non-blocking, out of this PR's diff).** `FileDescriptor`
> has no `Drop` impl; the `inc_ref`/pipe-add/`pin_inode` taken while cloning
> `sender_fds` (`syscall.rs:2383/2390/2415`) are balanced *only* by
> `scm_drop_fds`, which on the send side is reached only through the
> `scm_queued` revoke arms (`:2529`, `:2556`). When `scm_bind_peer_id ==
> u64::MAX` (peer already fully closed), `scm_queue` is skipped, `scm_queued`
> stays `None`, and the cloned fds are dropped at scope end **without release**
> → a bounded ref/pin leak on the narrow `sendmsg(SCM_RIGHTS)`-to-just-closed-peer
> race. This is *newly exposed by severing* (pre-fix `get_peer` returned a live
> index so `scm_queue` ran and balanced), but the pre-fix alternative was worse
> (it queued/mis-delivered), so this is strictly an improvement. Suggested
> follow-up in the SCM send path: `scm_drop_fds(sender_fds)` on the
> `peer_id == u64::MAX` skip. Cite `unix(7)`/`recvmsg(2)` SCM_RIGHTS.

**connect racing close:** `connect()` is atomic under `TABLE` (`:372-457`) and
sets mutual pairing; a `close()` of the client before `accept()` severs the
backlogged accept-side slot cleanly (it comes up `peer_closed` → EOF/EPIPE,
which is the correct `accept(2)` outcome for a peer that closed first). Safe.

### 2. POSIX semantics vs CURRENT master — correct.

- Drain-then-EOF (`recv(2)`): `read_msg` drains `recv_buf`, then `rx_eof → 0`
  (`:669-674`, `:642-648`). Fixes the pre-fix immediate-EOF data-loss.
- `write` after peer close → `-EPIPE` (`:537`, `:552`): correct return value per
  `write(2)`/`send(2)`. **SIGPIPE injection is a documented pre-existing gap**
  (`syscall.rs:7071`) and is *unchanged*: the new `-EPIPE` path flows through the
  exact conversion the pre-existing local-`shut_wr` `-EPIPE` path uses, so
  whatever SIGPIPE behavior exists is neither improved nor regressed here.
- Local `SHUT_RD` immediate-EOF preserved (`:629`, unchanged).
- Half-close vs full-close distinction is correct: peer `SHUT_WR` sets `rx_eof`
  only → `POLLIN|POLLRDHUP`, and `fully_hung_up` stays false (peer still
  `Connected`) so **no** `POLLHUP`; peer full close sets `peer_closed` →
  `POLLHUP`. Matches `poll(2)`/`epoll(7)`/`man 7 unix`.

### 3. Wake-edge preservation — sound.

Every new EOF/HUP/writable flag is committed under `TABLE`, the lock is dropped,
*then* the bell rings: `close()` `:814-816` → `drop(t)` `:820` → ring `:834-836`;
`shutdown()` `:722` → `drop(t)` `:734` → ring `:735`. No ring-before-commit on
any branch. Timing is identical to the pre-fix peer-`Free` probes: the same
`PollBellSource::UnixShutdown` wake now carries the `read_shutdown`
(POLLIN/RDHUP) and `fully_hung_up` (POLLHUP) edges, with the pre-fix
`t.0[peer].state==Free` POLLHUP probe replaced by the `peer_closed` flag set at
the same instant. The new `writable`-on-`peer_closed` edge is covered by the
same broadcast wake (the poll bell wakes all waiters; source is attribution
only). No waiter is stranded to the resync floor.

### 4. Mutual-pairing gate failure mode / SMP=2 — fails safe.

- Asymmetric mid-teardown (A→B, B re-paired to C): every mutating arm no-ops on
  the foreign pair because the gate `peer.peer_id == id` fails; read predicates
  give a benign wrong-answer at worst (read-only).
- Concurrent close from both ends (`TABLE` spin-mutex serializes on SMP): the
  second closer reads its own `peer_id`, already severed to `u64::MAX` by the
  first, and skips the peer-notify branch (`:797`). No double-sever, no
  double-EOF, no write into a recycled index. Severing `peer_id` is orthogonal
  to `ref_count`, so there is no double-free interaction.

### 5. Test quality — good; one NIT.

`test_unix_peer_close_sever_and_recycle` pins drain-then-EOF, `write→-EPIPE`,
the survivor readiness edges, the **sever-invariant** (`get_peer(a)==u64::MAX`,
`test_runner.rs:15614`), no-bleed on stale write/shutdown/close, and recycled-pair
functionality. It **would** have caught the original Gate-1 Finding A bleed: the
`:15614` sever assertion fires on the pre-fix code (unsevered `peer_id`)
independent of any recycle alignment.

> **NIT-2 (test, non-blocking).** The *direct* cross-channel-bleed assertions
> (`:15621`, `:15631`) only exercise the recycle hazard when the fresh pair
> `(c,d)` happens to reuse `b`'s index; if a lower slot is free the test skips
> that leg (`:15606-15612`) yet still passes. And because `a.peer_id` is severed
> to MAX, the stale `write(a)` cannot reach *any* slot, so the bleed check is
> somewhat vacuous either way — the sever assertion is what actually guards the
> property. The mutual-pairing *gate* (as distinct from severing) is never
> directly exercised because the public API cannot construct an
> asymmetric-but-valid dangling `peer_id`. Consider forcing the recycle
> (close intervening slots so `b`'s index is guaranteed lowest-free) to make the
> isolation leg deterministic.

### 6. Drift reconciliation — honest.

`fb749192` touched `kdb.rs` (2 lines: additive `rx_eof`/`peer_closed` fields on
`op_unix_diag`), `net/unix.rs`, and `test_runner.rs`. The re-land drops **only**
the `kdb.rs` hunk and the corresponding `UnixDiag` struct fields. `grep` for
`unix-diag`/`UnixDiag`/`diag_for` at HEAD returns **zero** matches — that diag
surface was removed during the three-week drift, so dropping it is correct. No
functional hunk was silently dropped; the remaining differences from the
original are doc-comment wording only. Honest.

---

**Bottom line:** merge-ready. The recycle-safety invariant is complete for all
mutating paths and verified against concurrent/SMP teardown; POSIX drain-then-EOF,
`-EPIPE`, and the half-close/full-close `POLL*` distinction are faithful; wake
edges are preserved. NIT-1 (SCM-send-to-closed-peer fd leak) and NIT-2 (make the
test's isolation leg deterministic) are worth a follow-up but neither blocks.

Spec basis: POSIX.1-2017 `write(2)`, `send(2)`, `recv(2)`, `shutdown(2)`,
`poll(2)`; `man 7 unix`; `man 7 epoll` (`EPOLLRDHUP` vs `EPOLLHUP`).

PR #682, opened 2026-07-01
ipc/eventfd: per-open-file-description refcount across fork/dup/SCM (re-land Fix A of #545)

me1iissa:
## Summary

Re-lands **Fix A** of the original **PR #545**, which merged on the pre-rewrite history and was **not carried across the 2026-06-10 history rewrite**. Fix B of that PR (EPOLLET wakeup-driven re-fire) was separately re-landed as **#605 / #609** via the rise-generation mechanism and is intentionally **not** duplicated here.

## The bug

An eventfd slot was freed unconditionally on the **first** `close(2)`, while `fork(2)`/`vfork(2)`/`clone(2)`-without-`CLONE_FILES`, `dup(2)`/`dup2(2)` (and thus `fcntl` `F_DUPFD` / `dup3`), and `SCM_RIGHTS` all duplicate descriptors that refer to the **same open file description** (POSIX.1-2017 §2.14).

A child process's pre-`execve(2)` close-on-exec scrub therefore destroyed the **parent's** live counter:

- the parent's next `write(2)` failed `EBADF`, and
- a later `eventfd2(2)` recycled the slot out from under the stale id, silently routing writes to an unrelated eventfd (CWE-416 class via slot reuse).

gecko's per-content-process `CloseSuperfluousFds` hits this on every content-process spawn (the AudioIPC waker eventfd).

Per `close(2)`: *"If fildes is the last file descriptor referring to the open file description, the resources associated with the open file description shall be released"* — only the last close may free.

## The fix (entry-side refcount)

- `EventFdEntry` gains `refs`; `create()` = 1, `close()` now decrements and frees (+ wakes parked readers) **only at zero**.
- `inc_ref()` at each descriptor duplication, mirroring the existing unix-socket / pipe refcount discipline:
  - fork / vfork / clone fd-table duplication (`inc_eventfd_refs_for_fork` at all three sites),
  - `dup(2)` / `dup2(2)`,
  - `SCM_RIGHTS` enqueue.
- A drop where a descriptor reference dies **outside** the `close(2)` arm:
  - `execve(2)` `FD_CLOEXEC` purge,
  - `dup2(2)` displaced-slot close,
  - process-exit fd teardown,
  - `scm_drop_fds` (undelivered in-flight SCM batches).

The inc/dec surface is balanced: create/dup/dup2/fork×3/SCM-enqueue increment; close-arm / process-exit / exec-CLOEXEC / dup2-displaced / scm-drop / create-rollback decrement. SCM delivery adds no ref (the enqueue-time inc already accounts for the receiver's copy), exactly matching the socket/pipe model.

## Tests

**Test 297** (`test_runner.rs`, gated on `firefox-test-core` / `test-mode`) drives the **real inc/dec call sites end to end**, so a future regression that drops a dec (re-introducing the *leak* direction) fails in the test, not just by inspection:

1. **Live trigger** — an eventfd shared across a simulated fork inherit + close scrub: the parent's `write(2)`/`read(2)`/readiness must survive the child's close; free only on last close (refs 1→2→1→0); write-after-last-close = `EBADF`.
2. **Recycle** — last-close-frees-then-`create()`-reuses check.
3. **Part 2** — the real `eventfd2`/`dup`/`close`/`write`/`read` dispatch arms (close-arm + dup inc/dec).
4. **Part 2b** — the real `dup2(2)` dispatch arm displacing an eventfd (dup2 inc + displaced-slot dec).
5. **Part 3** — an SCM enqueue → `scm_drop_fds` round trip (the real `scm_drop_fds` eventfd dec).
6. **Part 4** — `fork_process` (real fork inc via `inc_eventfd_refs_for_fork`) → `exit_group_pid` process-exit teardown (real exit-teardown dec), asserting the shared count returns to 1 across the child's teardown and 0 on the parent's own close.

The `execve(2)` `FD_CLOEXEC` purge dec is left to inspection (driving a real `execve` in-kernel replaces the address space and needs a target program), but it invokes the same `eventfd::close` dec the test proves via the close/exit paths.

## Known limitations (out of scope, matching the original PR's follow-up notes)

- eventfd ids are bare slot indices with **no generation tag**, so a genuinely-recycled slot cannot *reject* a truly-stale id. The guarantee here is that a **premature free never happens**, so no live holder is ever left holding a reused id. A generation tag to reject stale ids is a possible follow-up.
- `timerfd` / `signalfd` / `inotify` share the same unconditional-free defect and are left for a follow-up.
- `close_range(2)` routes each fd only through `vfs::close`, which **does** drop *pipe* reader/writer counts but **not** socket / eventfd / timerfd / signalfd refcounts — a **pre-existing** gap **unchanged** by this change (an eventfd closed via `close_range` leaked its slot before this change too), best fixed by routing `close_range` through the full close teardown.

## Verification

`scripts/qemu-harness.py check` passes clean for `test-mode`, `firefox-test-core`, and the default profile (no boot — a live measurement session owns the box; CI runs the full build + QEMU test suite).

Refs: POSIX.1-2017 §2.14 (open file description), close(2), fork(2), dup(2), eventfd(2), unix(7) SCM_RIGHTS.

### DIFF
--- a/kernel/src/ipc/eventfd.rs
+++ b/kernel/src/ipc/eventfd.rs
@@ -56,6 +56,15 @@ struct EventFdEntry {
     counter: u64,
     flags:   u32,
     in_use:  bool,
+    /// Open-file-description reference count.  One reference per
+    /// `FileDescriptor` (in any process's fd table) that names this slot,
+    /// plus one per in-flight SCM_RIGHTS copy.  Per POSIX.1-2017 §2.14,
+    /// `fork(2)`, and `dup(2)`: duplicated descriptors refer to the SAME
+    /// open file description, and per `close(2)` the underlying object is
+    /// released only when the LAST descriptor referring to it is closed.
+    /// `create()` starts this at 1; `inc_ref()` bumps it on fork/dup/
+    /// SCM-enqueue; `close()` decrements and frees the slot only at zero.
+    refs:    u32,
     /// Monotonic readiness-rise generation.  Bumped on every counter
     /// `0 → non-zero` transition (a fresh not-ready→ready edge per
     /// `eventfd(2)` + `epoll(7)`), and never on a write that only grows an
@@ -72,7 +81,7 @@ struct EventFdEntry {
 
 impl EventFdEntry {
     const fn empty() -> Self {
-        Self { counter: 0, flags: 0, in_use: false, rise_seq: 0 }
+        Self { counter: 0, flags: 0, in_use: false, refs: 0, rise_seq: 0 }
     }
 }
 
@@ -104,10 +113,29 @@ pub fn create(initval: u64, flags: u32) -> u64 {
             slot.in_use  = true;
             slot.counter = initval;
             slot.flags   = flags;
+            slot.refs    = 1; // the creating fd's open-file-description ref
             return i as u64;
         }
     }
     u64::MAX // No free slot
+}
+
+/// Add one open-file-description reference to a live slot.
+///
+/// Call sites mirror the unix-socket / pipe refcount discipline:
+/// `fork(2)`/`clone(2)`-without-CLONE_FILES fd-table duplication,
+/// `dup(2)`/`dup2(2)`/`fcntl(F_DUPFD)`, and SCM_RIGHTS enqueue.  Per
+/// POSIX.1-2017 §2.14 the duplicate descriptor refers to the same open
+/// file description; without this bump the FIRST `close(2)` from either
+/// holder would free the shared counter slot, leaving the survivor with
+/// a dangling id that fails EBADF — or worse, lands on a recycled slot.
+pub fn inc_ref(id: u64) {
+    let mut table = TABLE.lock();
+    if let Some(slot) = table.get_mut(id as usize) {
+        if slot.in_use {
+            slot.refs = slot.refs.saturating_add(1);
+        }
+    }
 }
 
 /// Non-blocking read from eventfd.  Returns the current counter as a u64
@@ -218,17 +246,38 @@ pub fn write(id: u64, val: u64) -> Result<(), i64> {
     Ok(())
 }
 
-/// Free an eventfd slot.  Wakes every reader parked on the slot so they
+/// Drop one open-file-description reference; free the slot only when the
+/// LAST reference is released.  Per POSIX `close(2)`: "If fildes is the
+/// last file descriptor referring to the open file description, the
+/// resources associated with the open file description shall be
+/// released."  An eventfd inherited across `fork(2)` (or duplicated via
+/// `dup(2)`/SCM_RIGHTS) is one shared object with multiple references —
+/// a child's close (e.g. a pre-`execve(2)` close-on-exec scrub) must NOT
+/// destroy the counter the parent is still writing to.
+///
+/// On the final free, wakes every reader parked on the slot so they
 /// observe EBADF on the next try_read call rather than blocking
 /// indefinitely against a counter that nobody can ever post to.
 pub fn close(id: u64) {
-    {
+    let freed = {
         let mut table = TABLE.lock();
-        if let Some(slot) = table.get_mut(id as usize) {
-            *slot = EventFdEntry::empty();
+        match table.get_mut(id as usize) {
+            Some(slot) if slot.in_use => {
+                if slot.refs > 1 {
+                    slot.refs -= 1;
+                    false
+                } else {
+                    *slot = EventFdEntry::empty();
+                    true
+                }
+            }
+            // Not in use: double-close or stale id — nothing to drop.
+            _ => false,
         }
+    };
+    if freed {
+        wake_readers_all(id);
     }
-    wake_readers_all(id);
 }
 
 /// Check if counter > 0 (used by `poll` / `select`).
@@ -317,4 +366,13 @@ pub fn waiter_cleanup(efd_id: u64, tid: u64) {
 pub fn debug_reader_waiter_count(efd_id: u64) -> usize {
     let waiters = EVENTFD_READ_WAITERS.lock();
     waiters.get(&efd_id).map(|l| l.len()).unwrap_or(0)
+}
+
+/// Test-only diagnostic: returns the open-file-description reference
+/// count for `efd_id` (0 when the slot is free).
+pub fn debug_ref_count(efd_id: u64) -> u32 {
+    let table = TABLE.lock();
+    table.get(efd_id as usize)
+        .map(|s| if s.in_use { s.refs } else { 0 })
+        .unwrap_or(0)
 }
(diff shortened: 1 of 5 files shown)


me1iissa (commented):
**Independent pre-merge review verdict: APPROVE-WITH-NITS** (posted as a comment — GitHub blocks self-approval; merge/approve authority remains with the coordinator).

## Independent pre-merge review — APPROVE-WITH-NITS

Adversarial read of the full inc/dec surface against `HEAD` (I re-enumerated every
descriptor-duplication and descriptor-destruction path independently rather than
trusting the PR's own table). The refcount is balanced; the release-blocking
(premature-free) direction is fully covered; two non-blocking nits below.

Spec basis: POSIX.1-2017 §2.14 (open file description), `close(2)`, `fork(2)`,
`dup(2)`, `dup2(2)`, `eventfd(2)`, `unix(7)` SCM_RIGHTS.

### 1. Balance audit — PASS

**INC (duplication) — every path covered:**
- `create()` → `refs = 1` (`ipc/eventfd.rs:116`).
- `dup(2)` → `inc_ref` (`syscall/mod.rs:4114`). `fcntl(F_DUPFD)` and
  `F_DUPFD_CLOEXEC` route through `sys_dup` (`subsys/linux/syscall.rs:8292-8294`) — covered.
- `dup2(2)` → `inc_ref` (`syscall/mod.rs:4179`). `dup3(2)` routes through `sys_dup2`
  (`subsys/linux/syscall.rs:5339`) — covered. `old_fd == new_fd` short-circuits
  before any inc (`syscall/mod.rs:4143-4149`) — correct, no spurious bump.
- `fork(2)` / `vfork(2)` / `clone`-VM: `inc_eventfd_refs_for_fork` at all three
  fd-cloning constructors (`proc/mod.rs:3338`, `:3607`, `:4316`) — sits 1:1 with the
  established `inc_socket_refs_for_fork` / `inc_pipe_refs_for_fork` sites, no
  fork-family function missed.
- SCM_RIGHTS enqueue → `inc_ref` (`subsys/linux/syscall.rs:2431`).

**DEC (destruction) — every path covered, all gated at `refs == 0`:**
- `close(2)` arm (`subsys/linux/syscall.rs:1300`).
- process-exit teardown (`proc/mod.rs:2601`, snapshot-then-drop, PROCESS_TABLE released).
- `execve(2)` FD_CLOEXEC purge (`syscall/mod.rs:1741`) — runs *after* `vm_space.replace`
  (`syscall/mod.rs:1691`), i.e. past the point of no return, so a *failed* exec cannot
  double-purge. Correct.
- `dup2(2)` displaced-slot close (`syscall/mod.rs:4202`).
- `scm_drop_fds` (`syscall/mod.rs:695`) — reached on SCM revoke (`subsys/linux/syscall.rs:2545,2572`)
  and on receiver-socket teardown (`net/unix.rs:752-754`, `scm_drain_receiver` → `scm_drop_fds`).
- create-rollback on fd-install failure (`subsys/linux/syscall.rs:10431,10463`).

**Paths that must NOT count, verified they don't:**
- `clone` with a shared fd table (`CLONE_THREAD`) creates a thread in the same
  process and does not clone the fd table (`subsys/linux/syscall.rs:3374-3381`) → no
  inc. And the kernel *always copies* the fd table on process-clone regardless of
  `CLONE_FILES` (`proc/mod.rs:3568-3578`), so the inc is always balanced by the child's
  independent closes — `CLONE_FILES` cannot cause a double-count here.
- SCM delivery/install (`subsys/linux/syscall.rs:2809-2818`) adds no second ref — the
  enqueue-time inc *is* the receiver's reference, balanced by the receiver's later
  `close(2)`. Matches the socket/pipe model.
- `/proc/self/fd/<N>` reopen declines sentinel fds (eventfd) outright with `NotFound`
  (`vfs/mod.rs:2255-2263`) rather than cloning the slot, so it introduces no unbalanced
  reference. (Guarded by the existing Test 295b.)

**Pre-existing gap, correctly disclosed and not worsened:** `close_range(2)` routes
each fd only through `vfs::close` (`subsys/linux/syscall.rs:5734-5735`), which drops
*pipe* reader/writer counts (`vfs/mod.rs:2480-2489`) but not socket/eventfd/timerfd/
signalfd refs. This is unchanged by the PR (`close_range` is untouched); an eventfd
closed via `close_range` leaked its table slot before this change too, so the change
is neutral there. See nit 2 for a wording tweak.

### 2. Free-at-zero semantics — PASS

`close()` commits `*slot = EventFdEntry::empty()` under the `TABLE` lock, drops the
lock, *then* calls `wake_readers_all(id)` — ring-after-commit ordered
(`ipc/eventfd.rs:250-280`). A parked reader is woken and re-checks via `try_read`,
which returns `-9` EBADF on the now-free slot (`ipc/eventfd.rs:151-154`), so a blocked
reader on a fully-closed eventfd does not hang. The wake fires *only* on the actual
final free (`freed == true`) — correctly suppressed on a non-last close, where the
object is still live and the reader should keep waiting.

### 3. Concurrency — PASS (one narrow, pre-existing, non-regressive note)

`refs` is mutated only under `TABLE.lock()` (atomic). `inc_ref` guards on
`slot.in_use` (`ipc/eventfd.rs:135`), so it cannot resurrect a logically-free slot
into an unbalanced phantom ref. The eventfd `TABLE` is effectively a leaf lock (its
functions never acquire `PROCESS_TABLE`), so the `dup`/`fork` nesting order
(PROCESS_TABLE → eventfd TABLE) has no reverse-order holder → no deadlock. Close
arithmetic cannot underflow (the `refs > 1` guard means the `-= 1` branch only runs
at `refs ≥ 2`).

Narrow SMP window (matches the accepted socket/pipe discipline, not introduced here):
a `fork` clones an fd referencing slot X, then a sibling thread frees X before the
post-clone `inc_ref` runs — `inc_ref` sees `in_use == false` and no-ops, leaving the
child with a dangling id. This never underflows or double-frees; its worst case is
exactly the dangling-id-on-recycle the PR already documents as a known limitation
(no generation tag), and it is no worse than status quo. Not a blocker.

### 4. Fix A / Fix B separation — PASS

No `write_edge` / `take_write_edge` machinery anywhere in the tree. The `rise_seq`
mechanism and `write()` body are untouched — the only edit to the shared struct is
adding `refs: 0` beside the existing `rise_seq: 0` in `empty()`. The sole
`EventFdEntry` constructor is `empty()` (used by the `TABLE` initializer and the free
path); no struct literal bypasses the new field, so `cargo check` passing is
sufficient to prove the field is initialized everywhere.

### 5. Test 297 — GOOD (coverage nit)

Part 1 pins the live regression (child scrub `close` → parent `write` still `Ok`;
`1→2→1→0`; write-after-last-close = EBADF). Part 1b pins the recycle. Part 2
exercises the *real* `eventfd2`/`dup`/`close`/`write`/`read` dispatch arms end to
end. Solid for the close-arm + dup balance.

### Nits (non-blocking)

1. **Leak-direction coverage.** Test 297 models fork inheritance and the child scrub
   with *direct* `eventfd::inc_ref` / `eventfd::close` calls; it does not drive the
   real `exit_group` process-teardown dec, the `execve` FD_CLOEXEC purge dec, or an
   SCM enqueue→drop round trip. Those paths are correct by inspection, but a future
   regression that drops the exit-teardown dec (re-introducing the *leak* direction)
   would not fail Test 297. Consider a fork→child-`_exit` or an SCM round-trip subcase.

2. **Disclosure wording.** The `close_range(2)` note calls the gap "type-agnostic";
   `vfs::close` does drop *pipe* reader/writer counts, so pipe is handled and only
   socket/eventfd/timerfd/signalfd are not. Minor precision only.

CI must run the full build + QEMU suite (the author ran `check` only) — that gate
stays with the coordinator, as does merge authority.

PR #683, opened 2026-07-01
ipc/pipe: re-land Linux-parity capacity + byte-preserving launcher drain + blocking-write livelock/atomicity fixes (lost in the 2026-06-10 history rewrite)

me1iissa:
## Re-land: lost pipe subsystem fixes

These changes were originally merged as **PR #559** (Linux-parity pipe capacity + launcher-pipe drain) plus the **#551 follow-ups** (blocking-write livelock / lost-wake / concurrent-writer atomicity) on the pre-rewrite history. The **2026-06-10 git history rewrite** orphaned that lineage; the #577 regraft brought the *base* blocking-write path back but not the follow-up fixes, and PR #559's capacity work was never carried across. A fresh static audit independently re-found the atomic-write livelock, confirming the regression. This PR re-lands the still-missing pieces onto current `master`, adapted to ~3 weeks of drift (`pipe.rs` was partially regrafted, the `syscall.rs` pipe arms and `gui/terminal.rs` evolved through the June GUI/reactor overhaul).

The value here is **capacity parity + data-integrity + a real correctness fix**, not "fixes multi-second parks": HEAD already suppresses the writer park for the console pipe via an inline drain — but that inline drain *discards* the child's bytes, and the atomic-write livelock is a genuine non-yielding 100% CPU spin.

### What's fixed

**1. Linux-parity pipe capacity (`pipe(7)`, `fcntl(2)`).**
The ring was `PIPE_BUF`-sized (4 KiB, the bare POSIX minimum) while `pipe(7)` documents a 65,536-byte default. The ring is now a per-pipe heap buffer, default 64 KiB; `PIPE_BUF` (4096) is preserved as the **atomicity** threshold (`limits.h`), a distinct concept from ring capacity. New `fcntl` `F_SETPIPE_SZ` (1031) / `F_GETPIPE_SZ` (1032) round the request up to a power of two (floor `PIPE_BUF`), return `EBUSY` when the buffered bytes would not fit, `EPERM` above the 1 MiB `pipe-max-size` ceiling, and `EBADF` on a non-pipe fd; a successful grow wakes writers parked on the previously-full ring.

**2. Byte-preserving launcher-pipe drain thread (data integrity).**
The kernel-side capture of a child's fd 1/2 is provisioned at the 1 MiB ceiling and drained by a dedicated kernel thread woken directly by the child's `write(2)`. Previously the only backstop against a starved reader was the inline console-drain in the blocking-write path, which frees the ring by **discarding** the child's stdout/stderr (data loss, plus a lock+memcpy pass per fill at gecko's >1 MiB stderr volume). The drain thread frees ring space **and** wakes the parked writer (`pipe_read_and_wake`) **without** discarding the bytes, mirroring them to serial / the terminal grid.

Single drain ownership (`DRAIN_THREAD_STARTED`): once the thread owns the pipe, `poll_output()` skips its own dequeue (keeping only the throttled render), and the inline console-drain **degrades to a last-resort backstop** that fires only when no drain thread owns the pipe (e.g. a build/context where it was never started). The no-park / no-undrained-window guarantee (#551/#577) is preserved, while the steady state prefers a byte-preserving park-with-wake over silent discard.

**3. Atomic-write livelock (`pipe_write_blocking`, `count <= PIPE_BUF`).**
The old writer-wait predicate returned "ready" on *any* free space, while the atomic arm only deposits when the *whole* record fits — so a blocking atomic write into `0 < space < count` spun `check -> ready -> retry` at 100% CPU (a non-yielding loop; re-found by static audit). `wait_writable` is now needs-aware: it parks unless `space() >= need` (`need == count` for an atomic write, `1` for a large write) and re-checks under the wait-list lock, so a drain freeing less than `need` re-parks cleanly instead of wake-spinning.

**4. `<=PIPE_BUF` atomicity under concurrent writers.**
The space-check (`pipe_space`) and the deposit (`pipe_write`) were separate `PIPE_TABLE` acquisitions, so two concurrent atomic writers could both observe `space >= count`, each deposit a self-capped short record, and publish two interleaved partials. New `pipe_write_atomic` does the check-and-deposit under one `PIPE_TABLE` critical section — the whole record iff `space() >= len`, else `NoSpace` without writing. A partial `<= PIPE_BUF` record is never published (`pipe(7)` `PIPE_BUF` atomicity).

### Deliberately not re-landed
- **The lost-wake fix** from the same follow-up (a reader draining without waking a parked writer) is already present at HEAD — every drain consumer reads through `pipe_read_and_wake`.
- **The kdb / per-syscall-arg diagnostics** from the original capacity commit: the kdb `uread` op already supersedes the proposed `user-mem` op, so re-adding it would be redundant.

### Interaction review (#668 / #678 / #680)
The launcher drain thread emits only the existing `[CHILD-OUT]` child-output mirror (gated as before) — it does **not** reintroduce per-read serial spam on `firefox-test-core`. It defers the repaint to the desktop reactor via `RENDER_DIRTY`, so drain throughput is not coupled to framebuffer blit speed.

### Tests
`test_pipe_blocking_write_semantics`:
- new capacity sub-test — default `65536`, round-up `5000 -> 8192`, shrink-below-buffered `EBUSY`, `>1 MiB` `EPERM`, non-pipe `EBADF`;
- exact-fill sub-tests pinned back to 4 KiB via `F_SETPIPE_SZ` so their arithmetic is capacity-independent;
- blocking-atomic-insufficient-space sub-test — a watchdog proves the writer **parks** on `PIPE_WRITE_WAITERS` (never observed under the old livelock) before any drain, then completes with the full record;
- concurrent two-writer sub-test — asserts the drained stream is exactly two contiguous same-byte runs (no split / interleave).

`test_pipe_console_drain_on_park` pinned to 4 KiB for its space assertions.

Both `test-mode` and `firefox-test-core` compile locally; CI runs the full suite.

Refs: `pipe(7)`, `write(2)`, `fcntl(2)` (`F_SETPIPE_SZ`/`F_GETPIPE_SZ`), POSIX.1-2017 §write.

### DIFF
--- a/kernel/src/gui/terminal.rs
+++ b/kernel/src/gui/terminal.rs
@@ -465,10 +465,10 @@ pub fn handle_key(msg: u32, wparam: u64, _lparam: u64) {
                     // Async path: spawn child with stdout → pipe.
                     match spawn_async(trimmed) {
                         Ok((pid, pipe_id)) => {
+                            // `spawn_async` already published EXEC_STDOUT_PIPE/
+                            // EXEC_PID/EXEC_RUNNING before unblocking the child;
+                            // set only the cosmetic `running_exec` here.
                             state.running_exec = Some((pid, pipe_id));
-                            EXEC_STDOUT_PIPE.store(pipe_id, Ordering::Release);
-                            EXEC_PID.store(pid, Ordering::Release);
-                            EXEC_RUNNING.store(true, Ordering::Release);
                             // Don't draw prompt yet — poll_output() will do it on exit.
                         }
                         Err(msg) => {
@@ -1321,6 +1321,14 @@ fn spawn_async(cmd: &str) -> Result<(u64, u64), alloc::string::String> {
     // backstop both defer to it (see `DRAIN_THREAD_STARTED`).
     ensure_output_drain_thread();
 
+    // Publish the authoritative drain handle BEFORE unblocking the child, so
+    // there is no window in which the child can run (and fill its console pipe)
+    // while the drain thread / poll_output still read `EXEC_STDOUT_PIPE == 0`
+    // and park.  Callers only need to set the cosmetic `TERMINAL.running_exec`.
+    EXEC_STDOUT_PIPE.store(pipe_id, Ordering::Release);
+    EXEC_PID.store(pid, Ordering::Release);
+    EXEC_RUNNING.store(true, Ordering::Release);
+
     // Now allow the scheduler to run the child.
     crate::proc::unblock_process(pid);
 
@@ -1346,13 +1354,10 @@ pub fn launch_process(path: &str) {
 
     match spawn_async(path) {
         Ok((pid, pipe_id)) => {
-            // Publish the drain handle FIRST, unconditionally — `poll_output()`
-            // drains `EXEC_STDOUT_PIPE` regardless of whether `TERMINAL` is
-            // currently `Some` or gets re-`init()`d later (which would reset
-            // `running_exec`).  This is the authoritative console-pipe handle.
-            EXEC_STDOUT_PIPE.store(pipe_id, Ordering::Release);
-            EXEC_PID.store(pid, Ordering::Release);
-            EXEC_RUNNING.store(true, core::sync::atomic::Ordering::Release);
+            // `spawn_async` already published the authoritative drain handle
+            // (`EXEC_STDOUT_PIPE`/`EXEC_PID`/`EXEC_RUNNING`) before unblocking
+            // the child, so `poll_output()` drains it regardless of `TERMINAL`
+            // re-`init()`.  Here we only set the cosmetic `running_exec` handle.
             let mut guard = TERMINAL.lock();
             if let Some(ref mut state) = *guard {
                 state.running_exec = Some((pid, pipe_id));
--- a/kernel/src/gui/terminal.rs
+++ b/kernel/src/gui/terminal.rs
@@ -72,6 +72,133 @@ const NO_EXEC_EXIT: i64 = i64::MIN;
 static LAST_EXEC_EXIT_CODE: core::sync::atomic::AtomicI64 =
     core::sync::atomic::AtomicI64::new(NO_EXEC_EXIT);
 
+/// Render-throttle state for the launcher-pipe drain: tick of the last
+/// framebuffer repaint, and whether drained-but-unrendered text is pending.
+/// Draining the launcher pipe (which frees space for POSIX-blocked writers)
+/// must not be rate-limited by framebuffer blit speed, so the dedicated
+/// drain thread only flags `RENDER_DIRTY`; `poll_output()` on the desktop
+/// reactor performs the throttled repaint.
+static LAST_RENDER_TICK: AtomicU64 = AtomicU64::new(0);
+static RENDER_DIRTY: AtomicBool = AtomicBool::new(false);
+
+/// One-shot guard and single source of truth for drain OWNERSHIP of the
+/// launcher pipe.
+///
+/// Exactly one consumer may dequeue bytes from a pipe at a time.  Once this
+/// flag is `true` the dedicated [`output_drain_thread`] owns byte consumption
+/// of whatever `EXEC_STDOUT_PIPE` currently names, and both `poll_output()`
+/// and the inline console-drain backstop (`pipe_write_blocking`) must defer to
+/// it: a second consumer appending to the terminal grid under a separate
+/// TERMINAL-lock acquisition would interleave the child's output on screen.
+///
+/// There is never a window in which NO drainer is responsible for the pipe
+/// (the deadlock class #551 closed): the drain thread never exits — it parks
+/// when no exec is attached and re-snapshots `EXEC_STDOUT_PIPE`/`EXEC_RUNNING`
+/// on every wake — so "flag is true" ⟺ "a live drainer owns the pipe".  Before
+/// the thread is started (pre-first-launch) `poll_output()` remains the drainer
+/// and the inline backstop stays armed.
+static DRAIN_THREAD_STARTED: AtomicBool = AtomicBool::new(false);
+
+/// True once the dedicated drain thread owns byte consumption of the launcher
+/// pipe (so `poll_output()` skips its own dequeue and the inline console-drain
+/// backstop defers to a REAL, byte-preserving drain).  See
+/// [`DRAIN_THREAD_STARTED`].
+#[inline]
+pub fn drain_thread_owns_pipe() -> bool {
+    DRAIN_THREAD_STARTED.load(Ordering::Acquire)
+}
+
+/// Dedicated kernel-thread drain loop for the launcher stdout/stderr pipe.
+///
+/// `poll_output()` runs only on the shared desktop reactor, whose iteration
+/// latency is bounded but non-zero; POSIX write(2) blocks a writer on a full
+/// pipe, so a child whose diagnostic output outruns the reactor drain would
+/// park its writers.  Two pre-existing mechanisms already suppress the park —
+/// the 64 KiB (launcher: 1 MiB) ring absorbs bursts, and an inline
+/// console-drain backstop in `pipe_write_blocking` frees the ring from the
+/// writer's own context — but the backstop *discards* the drained bytes (data
+/// loss for anyone reading the child's stdout/stderr).
+///
+/// This thread provides a REAL, byte-preserving drain decoupled from the
+/// reactor: it parks on the pipe's reader wait-list (`wait_readable`), is woken
+/// DIRECTLY by the child's `write(2)` (`wake_readers_all`), and drains via
+/// `pipe_read_and_wake` — which both frees ring space and wakes any parked
+/// writer (so a blocked `writev(fd=2)` is never wedged) WITHOUT throwing the
+/// bytes away.  Drained text is mirrored to serial (for the harness) and, when
+/// a terminal window exists, appended to the grid; the throttled repaint stays
+/// with `poll_output()` via `RENDER_DIRTY`.
+fn output_drain_thread() {
+    crate::hal::enable_interrupts();
+    let mut raw = [0u8; 4096];
+    loop {
+        // Snapshot the live drain target from the authoritative atomics (NOT
+        // `TERMINAL.running_exec`, which a GUI re-init can clear out from under
+        // a live child — see `EXEC_STDOUT_PIPE`).
+        let running = EXEC_RUNNING.load(Ordering::Acquire);
+        let pipe_id = EXEC_STDOUT_PIPE.load(Ordering::Acquire);
+        if !running || pipe_id == 0 {
+            // No child attached — park on the global poll bell with a bounded
+            // resync so an attach we missed is re-checked within ~100 ms.
+            let now = crate::arch::x86_64::irq::get_ticks();
+            crate::ipc::waitlist::wait_poll_event(now + 10, || {
+                EXEC_RUNNING.load(Ordering::Acquire)
+                    && EXEC_STDOUT_PIPE.load(Ordering::Acquire) != 0
+            });
+            continue;
+        }
+
+        // REAL drain: frees ring space AND wakes any writer parked on a full
+        // pipe (POSIX write(2); man 7 pipe) — WITHOUT discarding the bytes.
+        let n = crate::ipc::pipe::pipe_read_and_wake(pipe_id, &mut raw).unwrap_or(0);
+        if n > 0 {
+            // Mirror the child's stdout/stderr to serial for the harness — the
+            // same child-output stream `poll_output()` emits, NOT a per-read
+            // diagnostic marker (kept off the fast-path spam budget).
+            #[cfg(any(feature = "xeyes-test", feature = "firefox-test-core"))]
+            {
+                let text = core::str::from_utf8(&raw[..n]).unwrap_or("\u{FFFD}");
+                crate::serial_println!("[CHILD-OUT] {}", text.trim_end_matches('\n'));
+            }
+            // Append to the terminal grid for display when a window exists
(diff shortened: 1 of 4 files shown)


me1iissa (commented):
## Independent pre-merge review — REQUEST-CHANGES

Adversarial read of the full diff (`ipc/pipe.rs`, `subsys/linux/syscall.rs`, `gui/terminal.rs`, `test_runner.rs`) against base `23fd66ae`. One **hard blocker** (CI-red, a regression the PR itself introduced); the four core fixes are otherwise correct. Verdict: **REQUEST-CHANGES** — the blocker is a one-line test fix, but green CI is a merge gate.

---

### 🔴 BLOCKER — CI is red on an unallowlisted regression this PR introduced

`kernel smoke (QEMU boot + test suite)` **fails**: `431/438 passed`, and the allowlist drift check flags exactly one *unallowlisted* failure:

```
[FAIL] pipe_console_drain — non-console full-pipe O_NONBLOCK write returned 256 (expected -11 EAGAIN — no inline drain)
```

Root cause is in this PR's own test edits. `test_pipe_console_drain_on_park` pins its **primary** pipe to 4 KiB (`test_runner.rs:39670`) so its exact-fill arithmetic survives the new 64 KiB default — but the **negative-control** pipe created at `test_runner.rs:39726-39739` was **not** pinned:

```
let (ctl_rfd, ctl_wfd) = ...pipe2(O_NONBLOCK)...      // 39735 — DEFAULT 65536 capacity
write(ctl_wfd, filler /*4096*/, 4096);                // 39737 — fills 4096 of 65536
let ctrl = write(ctl_wfd, chunk /*256*/, 256);        // 39738 — 61440 free ⇒ returns 256
if ctrl != -11 { FAIL }                               // 39742 — expected EAGAIN
```

With the default now 64 KiB (`pipe.rs:47`), the control pipe holds `65536 - 4096 = 61440` free bytes, so the 256-byte `O_NONBLOCK` write legitimately succeeds (`write(2)`/`pipe(7)`: a non-full pipe accepts the write) instead of `EAGAIN`. The **product code is correct**; the test's "full to the brim" premise is stale. Fix is test-only: pin `ctl_wfd` to 4096 mirroring `:39670` (or fill it to `65536`). Until CI is green this cannot merge.

Note: the other 6 failures in the run (`dynamic_elf`, `pie_elf`, `TCC`, `busybox`, `firefox_oracle`, `test_522_tcp_rx_poll_bell`) are pre-existing allowlisted/environmental — not attributable to this PR.

---

### Dim 1 — No-park regression invariant (highest stakes): PASS, one low-severity ordering fragility

The chain holds. After first launch the inline discard backstop is gated off for the launcher pipe forever (`DRAIN_THREAD_STARTED` latches true; `syscall.rs:7207`), so the **sole** waker of a parked launcher-pipe writer is `output_drain_thread`. It is provably live: the loop never returns (`terminal.rs:127-166`), only bounded parks; and `create_kernel_process` **panics** (not silently returns) on kstack-alloc failure (`proc/mod.rs:1588-1589`), so `DRAIN_THREAD_STARTED==true ⟺ a live drainer` — no silent-failure deadlock. The writer's own `wake_readers_all` on the pipe-filling deposit wakes the drainer (`syscall.rs:7127`/`:7148`), the drainer's `pipe_read_and_wake` frees space + `wake_writers_all` before it ever touches `TERMINAL.lock()` (load-bearing work first, `terminal.rs:139-150`), and both TOCTOU windows are closed by the re-check-under-waitlist-lock (`pipe.rs:598-604`, `:553-563`). The drain thread never nests PIPE_* under TERMINAL, so it cannot be a side of a TERMINAL/PIPE deadlock.

**Fragility (low severity, not a deadlock):** `spawn_async` latches `DRAIN_THREAD_STARTED` and `unblock_process(pid)` (`terminal.rs:1322,1325`) *before* the caller publishes `EXEC_STDOUT_PIPE`/`EXEC_RUNNING` (`launch_process` `:1353-1355`; terminal handler `:469-471`). In that gap the drainer reads `pipe_id==0`/`running==false` and parks on the poll bell while the inline backstop is *already* gated off. Bounded by the `now+10` resync (`wait_poll_event` returns on timeout → loop re-snapshots, `waitlist.rs:347,372-378`) plus the write's own poll-bell ring, so worst case is a ≤~100 ms first-write stall in a race that requires the child to fill 1 MiB before the caller's next few instructions — practically unreachable on SMP, impossible on a single core. **Recommend** publishing `EXEC_STDOUT_PIPE`/`EXEC_RUNNING` *before* `unblock_process` to remove the window structurally.

### Dim 2 — Livelock fix correctness: PASS (CI-verified)

`wait_writable(need)` returns `Ready` iff `space() >= need || readers==0`, else `Enqueued`, with the re-check under `PIPE_WRITE_WAITERS` briefly holding `PIPE_TABLE` (`pipe.rs:594-614`). `need==count` (atomic) / `1` (large) at `syscall.rs:7233`. A drain freeing `<need` leaves `Enqueued` → genuine park (no spurious-wake livelock); `>=need` wakes and re-evaluates. `pipe_write_atomic` does check-and-deposit under one `PIPE_TABLE` section (`pipe.rs:302-317`) — no TOCTOU between check and copy, and no pipe lock is held across `schedule()` (the park lives entirely inside `wait_writable`). `test_pipe_blocking_write_semantics` (capacity + atomic-park watchdog + concurrent-two-writer sub-tests) **passed** in the same CI run, so the atomicity/park behavior is empirically confirmed and the multi-process park test does not hang.

### Dim 3 — Capacity semantics: PASS, two nits

`set_capacity` (`pipe.rs:95-116`) rounds up to power-of-two with floor `PIPE_BUF`, `EBUSY` on shrink-below-fill, `EPERM` above 1 MiB; boundary `want==1<<20` is accepted, `want>1<<20` → EPERM (verified by trace). `O_NONBLOCK` partial writes `>PIPE_BUF` are handled (`syscall.rs:7136-7170`), atomic `≤PIPE_BUF` never partial. Nits:
- `F_SETPIPE_SZ(arg==0)` → `EINVAL` (`syscall.rs:8373`); `fcntl(2)` on Linux silently rounds a sub-page request up to the minimum rather than erroring. Minor divergence, harmless.
- **Eager 64 KiB zeroing under the global lock.** `create_pipe` evaluates `Pipe::new(id)` (a `vec![0; 65536]`, `pipe.rs:79`) as the argument to `PIPE_TABLE.lock().push(...)` (`pipe.rs:222`), so the 64 KiB memset runs *while the guard is held*; `pipe_set_capacity` likewise zeroes up to 1 MiB under `PIPE_TABLE` (`pipe.rs:107`). Both serialize all pipe traffic system-wide across a memset. Recommend hoisting `Pipe::new` out of the lock (`let p = Pipe::new(id); PIPE_TABLE.lock().push(p);`) and considering lazy/paged growth — dozens of FF pipes × eager 64 KiB is a few MiB (fine vs budget) but wasteful for 1-byte self-wake pipes.

### Dim 4 — Behavioral compatibility: PASS except the blocker

The 4 KiB pins on the exact-fill sub-tests are legitimate decoupling, not masking — sub-test 0 separately asserts the 64 KiB default (`pipe(7)`). No production code assumes `capacity==4096` (`grep PIPE_BUF_SIZE` outside `pipe.rs`/tests is clean; the public `PIPE_BUF=4096` const is retained as the atomicity threshold). The omitted `arg1`/`record_syscall` diagnostics are cleanly absent (no dangling refs in the syscall diff). The one gap is the un-pinned negative-control pipe = the CI blocker above.

### Dim 5 — FF hot-path perf: PASS, one suggestion

The self-wake pipes are now heap-`Vec`-backed; per-write cost is unchanged (bounds math is identical, no reallocation on the write path). The drain thread reads 4096 B/iter and rings the **global** poll bell via `pipe_read_and_wake → wake_writers_all → ring_poll_bell_for(Pipe)` on every non-empty read (`pipe.rs:658`) — ~256 herd-wakes per 1 MiB stderr burst. Per-byte frequency matches the old `poll_output` loop (not a new regression), but the dedicated thread's tight `continue`-while-flowing loop can drain faster → more rings/sec under the storm regime. Suggest a larger heap drain buffer (e.g. 64 KiB) to amortize the wake+bell cost ~16×.

### Dim 6 — Drift honesty: VERIFIED

`pipe_read_and_wake` (the a9aa0855 lost-wake fix) is confirmed already at base `23fd66ae` (`pipe.rs:187` there), and both drain consumers — `poll_output` and `output_drain_thread` — route through it. The `uread`-supersedes-`user-mem` claim holds (no `user-mem` op reintroduced). No silent drops from the three source commits observed. Documented lock orders (`PIPE_WRITE_WAITERS→THREAD_TABLE`, `PIPE_READ_WAITERS→PIPE_TABLE`) are preserved, including the new `wake_writers_all`-on-grow in `pipe_set_capacity`.

---

**Summary:** the capacity/atomicity/livelock/drain design is sound and CI-proven for the new behavior. Blocking on the single red CI test (unpinned negative-control pipe, `test_runner.rs:39726`). Recommended before merge: (1) **fix the control-pipe pin** [required], (2) publish `EXEC_STDOUT_PIPE`/`EXEC_RUNNING` before `unblock_process` [robustness], (3) hoist the eager alloc out of `PIPE_TABLE` [perf]. Refs: `pipe(7)`, `write(2)`, `fcntl(2)`.

me1iissa (commented):
## Re-review of delta `e6dd138a..bad0ca5b` (4 files, +42/-16) — blocker cleared

Re-read only the delta. All four review items are correctly addressed and CI is now green at `bad0ca5b` (`kernel smoke (QEMU boot + test suite)` passes). Item-by-item:

**1. Blocker fix — negative-control pipe now genuinely full (`test_runner.rs:39736-39746`).** The control pipe is pinned to 4 KiB via `F_SETPIPE_SZ` *before* the 4096-byte fill, so it is now truly full (4096/4096) when the 256-byte `O_NONBLOCK` write is issued → `EAGAIN (-11)`. The negative control now tests exactly what it claims: that the inline console-drain is *not* applied to an ordinary (non-console) pipe even when full (`pipe(7)`, `write(2)`). The pin-failure branch closes all four fds (control + primary) — no leak. **Confirmed fixed** (this test now passes in the green run).

**2. Rec 1 — spawn_async reorder: window closed, no new one opened.** `EXEC_STDOUT_PIPE`/`EXEC_PID`/`EXEC_RUNNING` are published at `terminal.rs:1328-1330`, *before* `unblock_process` at `:1333`, so the child cannot run (and fill its console pipe) while the drain thread / `poll_output` still read `EXEC_STDOUT_PIPE == 0`. Verified no new window:
   - *Latch-before-publish holds:* `ensure_output_drain_thread()` (`:1322`, the `DRAIN_THREAD_STARTED` CAS) precedes the `EXEC_RUNNING` Release store (`:1330`) in program order. Any thread that observes `EXEC_RUNNING == true` via the Acquire load in `console_drain_pipe_id()` therefore also observes `drain_thread_owns_pipe() == true` (release/acquire on `EXEC_RUNNING` orders the prior latch). So the inline backstop — which only recognises the pipe once `EXEC_RUNNING` is true — always sees the drain thread owning it and correctly defers; there is no inverted "pipe recognised but backstop thinks no owner" window.
   - *Reap/teardown race unchanged:* the reap clear (`poll_output` `:1700-1702`) fires only on child exit and is gated against a new launch by the caller (a new spawn happens only when no child runs), exactly as before. The reorder moved the spawn-side stores a few instructions earlier (into `spawn_async`) but did not change reap-vs-spawn concurrency.
   - *Cosmetic decoupling correct:* the two callers (`:466`, `:1355`) now set only `TERMINAL.running_exec`; the authoritative drain path keys off the atomics, so a `TERMINAL == None` / re-`init()` no longer loses the handle. Confirmed exactly one launch-path publish site (`:1328-1330`) and no duplicate/missing stores.

**3. Rec 2 — create_pipe hoist: no id race.** `id` comes from `NEXT_PIPE_ID.fetch_add` (`pipe.rs:221`), atomic, so two concurrent `create_pipe` callers get distinct ids regardless of the lock; each allocates its own ring via `Pipe::new(id)` before the lock and pushes under it (`:226-229`). The id was already computed before the lock pre-delta — only the (self-contained, id-keyed) allocation moved out — so no collision and no observability gap (a pipe is not in `PIPE_TABLE`, hence not lookup-visible, until `push`). **Confirmed safe.**

**4. Rec 3 — F_SETPIPE_SZ(0) rounds up.** Removing the `arg == 0 → EINVAL` guard (`syscall.rs:8371`) and letting `pipe_set_capacity(pipe_id, 0)` floor to `PIPE_BUF` (4096) matches `fcntl(2)`: "Attempts to set the pipe capacity smaller than the page size are silently rounded up to the page size." `set_capacity(0)` leaves `cap` at the `PIPE_BUF` floor (`pipe.rs:96-97`) and returns 4096 (or `EBUSY` if buffered > 4096); a shrink from the 64 KiB default is the documented "at least arg bytes, rounded up" behaviour, and `wake_writers_all` is correctly *not* fired on a shrink. The previous `EINVAL` actually contradicted the man page, so this is a correctness improvement. **Confirmed aligned.**

### Residual (optional, non-blocking — carried from the first pass, not requested)
- Drain thread reads 4096 B/iter and rings the *global* poll bell per non-empty read (`pipe.rs:658`) — a larger heap drain buffer would amortise the wake+bell cost under the stderr-storm regime. Per-byte frequency matches the old `poll_output` loop, so not a regression.
- `create_pipe` still eagerly zeroes the full 64 KiB default (now out of the lock, per Rec 2); lazy/paged growth would be more memory-frugal for tiny self-wake pipes. Spec-compatible as-is.

Both are optional follow-ups; neither blocks merge.

### FINAL VERDICT: APPROVE-WITH-NITS
The blocker is resolved, all four requested fixes are correct, the reorder closes the park window without opening a new one, and CI is fully green. The two residual items are optional perf/memory follow-ups. Refs: `pipe(7)`, `write(2)`, `fcntl(2)`.

Issue #655, opened 2026-06-25
#582 torn-saved-RFLAGS 0x7f: third kstack-recycle path (post case-A/B hardening)


me1iissa:
## Summary

After the case-A + case-B kernel-stack-recycle hardening (PR linked below), the
SMP=2 windowed-Firefox `0x7f UNEXPECTED_KERNEL_MODE_TRAP` (a kernel-mode `#DB`
from `popfq` of a torn saved-RFLAGS slot, TF=1) is **reduced but NOT fully
closed**: ~2 of 4 SMP=2 KVM boots still bugcheck. A **third kstack-recycle
path** remains that trips neither hardening gate.

## What is already closed (the hardening PR)

- **Case A** — the reaper returned emergency-tier (16K/8K/4K) kernel stacks
  straight to the page allocator with no quiescence gate (unlike the standard
  256 KiB dead-stack cache). Closed by a `CPU_SWITCH_GEN` + tick quiescence
  quarantine on the emergency-tier free path.
- **Case B** — `alloc_kernel_stack` could return a base whose span overlaps a
  *live* thread's kernel-stack range. Closed by an alloc-side range-overlap
  guard (`kstack_candidate_aliases_live`) that rejects + quarantines + retries.

Both are regression-tested (Test 655 free-side survey, Test 656 alloc-side
guard — both PASS).

## The remaining third path (this issue)

Verification with the `582-diag` DR0 write-watch instrument (SMP=2 KVM, 4
windowed-FF boots): the boots that reached the crash window still bugchecked
`0x7f` (faulting RIP `current_pid+0x37` on the resume path, and `memset+0x39`
on a kstack zero-fill — the same torn-RFLAGS family), with:

- **zero** alloc-side alias-guard rejections (the guard never fired), and
- **zero** foreign DR0 catches (`writer_tid != owner_tid`).

So the residual is **not** case B (no live-range overlap at alloc time) and is
not caught by the free-side survey. The recycled frame's owner is **not present
in `THREAD_TABLE` with a matching `kernel_stack_base`** at the moment of the bad
allocation — so the alloc range-guard sees no overlap — and the frame did **not**
pass through `reap_dead_threads_sched`'s free-side survey. The 1-in-32 DR0
sampling also missed the actual torn victim, so no catch named it.

Mechanism family (per Intel SDM Vol. 3A §6.14 interrupt/TSS-RSP stack switch;
[...]

PR #685, opened 2026-07-01
sched: candidate-gate starvation hypothesis REFUTED (0/43M gate-skips) + instrumentation + latent self-resume wait-stamp fix

me1iissa:
## Summary

Investigation of the reported **scheduler candidate-gate starvation** of freshly-spawned content-process threads (dispatched as the prime suspect for a ~50x page-load slowdown). The hypothesis is **REFUTED by live measurement**; this PR lands the instrumentation that proves it, plus a real-but-latent anti-starvation bookkeeping fix surfaced along the way.

**Do NOT merge without independent review** (self-authored kernel change).

## Hypothesis under test

A freshly-Ready thread was thought to sit `Ready` with `ctx_rsp_valid == false` for seconds — invisible to the entire fairness ladder (`select_next_core`'s `ctx_rsp_valid`/affinity candidate gates and the force backstop) — producing occasional very-high wait-age `[SCHED/STARVE] force-select` events (e.g. 706 ticks).

## Result: refuted by live evidence (SMP=1, KVM, live Wikipedia load)

Over **43.1M picker passes** the per-gate skip population was:

```
ctx_rsp_invalid = 0    affinity = 0    mirror = 0    self_resume_ready = 0
```

- Not once was a Ready thread skipped by the `ctx_rsp_valid` gate; every Ready thread in the live table read `ctx_rsp_valid == 1` (5 samples x ~21 threads, 0 with `crv==0`).
- `maintain_passes == ctx_switches` (43.1M each) — schedule() runs continuously; it is not frozen.
- The `[SCHED/STARVE] force-select` lines are ~99% the **by-design TID-0 reactor cadence** ("waited 1 tick", per its tight `STARVE_FORCE_TICKS_BSP` deadline), plus occasional legitimate rescues of a low-priority worker out-scored in a busy run-queue that reached its ~1 s force deadline — all with `ctx_rsp_valid == true`. The high wait-age is real run-queue wait under momentary contention, not a gate-invisible thread.

A force-select log's own semantics require the fired thread to be a *candidate* (it passed every gate), so it cannot have been `ctx_rsp_valid == false` at fire time. **The candidate-gate starvation hypothesis does not hold; the page-load slowdown is not a scheduler gate.**

## Changes

1. **Diagnostic instrumentation** (feature `sched-gate-stats`, compiled OUT by default — the pick hot path and `firefox-test-core` stay byte-identical): per-gate skip counters (`ctx_rsp_invalid`/`affinity`/`mirror`) + a `self_resume_ready` counter, surfaced via kdb `cpu-state` alongside a new `sched` object (`maintain_passes`, `ctx_switches`, `force_total`). Per-thread `cpu-state` rows gain `rst` (`ready_since_tick`) so a reader can compute each Ready thread's true wait age. New `percpu::maintain_passes()` accessor.

2. **Latent fix** — the picker's in-place self-resume path (no other thread Ready; the in-picker due-wake drain flipped the current thread `Blocked -> Ready` as its own sleep/timeout deadline elapsed) returned to the thread leaving it `Ready` with its original `ready_since_tick`, violating the documented invariant ("`ready_since_tick` cleared the moment the thread is selected to Run"). It now applies the normal pick's post-selection bookkeeping (`state -> Running`, `ready_since_tick -> 0`, `last_cpu -> this CPU`). `ctx_rsp_valid` is deliberately untouched (the saved RSP stays stale until a real switch-out re-validates it — the invariant the `ctx_rsp_valid` gate and the #672/#673 on-CPU interlock depend on); promoting to `Running` only strengthens candidate exclusion, so SMP correctness is preserved. The live counter shows this path never fires under the busy load run-queue (`self_resume_ready = 0`) — this is **hardening for quieter regimes, NOT the cause of the slowdown**.

3. **Test 681** (`test_681_self_resume_clears_wait_stamp`): deterministic proof that a self-resuming `Ready` current is promoted to `Running` with its wait stamp cleared and `last_cpu` set, while `Running`/`Blocked`/`Sleeping`/`Dead` currents are untouched and `ctx_rsp_valid` is never mutated.

## Verification

- **Test-mode**: Test 681 PASS; scheduler cluster (648/649/681/650/652) all green; 208 PASS / 3 FAIL where all 3 FAILs are pre-existing environmental (missing `/disk/bin/tcc`, `/disk/bin/busybox`, `pie_elf` process test) — unrelated to this change.
- **A/B (live Wikipedia load, SMP=1)**: baseline (instrumentation only) vs fixed build are byte-identical in scheduler behaviour — both `gate_skips` all 0, both `maintain_passes == ctx_switches`, both idle-dominant, same force-select wait-age distribution, 0 bugchecks/faults. Gap profile and time-to-content are unchanged by construction (scheduler exonerated).

Harness-only testing (`scripts/qemu-harness.py`). Cite: POSIX sched(7) SCHED_OTHER.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -16,6 +16,12 @@ default = []
 # deterministic Test 648 is the authoritative merge gate; this is supplementary
 # live evidence.
 sched-pick-xcheck = []
+# `sched-gate-stats` — count the population of Ready threads skipped by each
+# scheduler candidate gate (ctx_rsp_valid / affinity / mirror-slot) and how
+# often the picker returns to a self-resuming Ready current.  Diagnostic-only:
+# the increments compile out entirely by default so the pick hot path (and the
+# `firefox-test-core` profile) stays byte-identical.  Read via kdb `cpu-state`.
+sched-gate-stats = []
 # `heap-canary` — fence every live kernel-heap allocation with front/rear
 # magic red-zone words (validated on free) so an out-of-bounds write that
 # clobbers free-list metadata is caught at free-time, naming the victim
--- a/kernel/src/kdb.rs
+++ b/kernel/src/kdb.rs
@@ -1898,9 +1898,9 @@ fn op_cpu_state(out: &mut String) {
                 let crv = t.ctx_rsp_valid.load(core::sync::atomic::Ordering::Relaxed);
                 let _ = write!(
                     out,
-                    r#"{{"tid":{},"pid":{},"st":"{:?}","lcpu":{},"aff":{},"mcpu":{},"mprio":{},"prio":{},"wt":{},"crv":{}}}"#,
+                    r#"{{"tid":{},"pid":{},"st":"{:?}","lcpu":{},"aff":{},"mcpu":{},"mprio":{},"prio":{},"wt":{},"crv":{},"rst":{}}}"#,
                     t.tid, t.pid, t.state, t.last_cpu, aff, ms_cpu, ms_prio,
-                    t.priority, t.wake_tick, crv as u8);
+                    t.priority, t.wake_tick, crv as u8, t.ready_since_tick);
             }
             out.push(']');
         }
@@ -1908,6 +1908,31 @@ fn op_cpu_state(out: &mut String) {
             out.push_str(r#","threads":"THREAD_TABLE held""#);
         }
     }
+
+    // ── Scheduler-liveness + candidate-gate skip population ──────────────────
+    // `maintain_passes` and `ctx_switches` advance once per picker pass / real
+    // context switch; a reader sampling them twice across a STARVE window can
+    // tell whether `schedule()` is running (H3: gate/stamp artifact) or frozen
+    // (H2: a monopolising in-kernel path).  `force_total` is the cumulative
+    // anti-starvation force-select count.  `gate_skips` (feature `sched-gate-
+    // stats`) is the population of Ready threads skipped by each candidate gate.
+    let _ = write!(
+        out,
+        r#","sched":{{"maintain_passes":{},"ctx_switches":{},"force_total":{}"#,
+        crate::sched::percpu::maintain_passes(),
+        crate::perf::context_switches(),
+        crate::sched::starve_force_count(),
+    );
+    #[cfg(feature = "sched-gate-stats")]
+    {
+        let (ctx_rsp, aff, mirror, self_resume) = crate::sched::gate_stats::snapshot();
+        let _ = write!(
+            out,
+            r#","gate_skips":{{"ctx_rsp_invalid":{},"affinity":{},"mirror":{},"self_resume_ready":{}}}"#,
+            ctx_rsp, aff, mirror, self_resume);
+    }
+    out.push('}');
+
     out.push('}');
 }
 
--- a/kernel/src/sched/mod.rs
+++ b/kernel/src/sched/mod.rs
@@ -452,6 +452,48 @@ pub fn starve_force_count() -> u64 {
     SCHED_STARVE_FORCE_TOTAL.load(Ordering::Relaxed)
 }
 
+// ── Candidate-gate skip population counters (diagnostic, feature-gated) ──────
+//
+// The force-select backstop only logs a thread whose stale wait-age survives
+// past its deadline; a Ready thread skipped by a candidate gate for a sub-log
+// interval leaves no trace.  These counters record the POPULATION of gate
+// skips so an A/B can show how often each gate hides a Ready thread from the
+// fairness ladder, and how often the picker returns to a self-resuming Ready
+// current (the path that historically left an uncleared `ready_since_tick`).
+//
+// Gated behind `sched-gate-stats` so the shipped hot path (default and
+// `firefox-test-core`) is byte-identical — the increments compile out entirely.
+#[cfg(feature = "sched-gate-stats")]
+pub mod gate_stats {
+    use core::sync::atomic::{AtomicU64, Ordering};
+
+    /// A Ready candidate was skipped because `ctx_rsp_valid == false` (its saved
+    /// kernel RSP is not yet valid — mid switch-out or self-resume window).
+    pub static SKIP_CTX_RSP_INVALID: AtomicU64 = AtomicU64::new(0);
+    /// A Ready candidate was skipped because its hard `cpu_affinity` pin does
+    /// not match the picking CPU.
+    pub static SKIP_AFFINITY: AtomicU64 = AtomicU64::new(0);
+    /// A Ready non-idle thread was filtered out of the per-CPU candidate set
+    /// because its `mirror_slot` targets a different CPU (or is still `None`).
+    pub static SKIP_MIRROR: AtomicU64 = AtomicU64::new(0);
+    /// The picker found no other Ready peer and returned to the current thread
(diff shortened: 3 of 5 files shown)


me1iissa (commented):
## Independent pre-merge review — PR #685 (a809184c, base 23fd66ae)

**Verdict: APPROVE-WITH-NITS.** Reviewed adversarially, read-only. All three nits are non-blocking (wording/comment/coverage); no correctness or SMP-safety defect found. Self-approve is blocked (I am the PR author on this token), so this is filed as a comment — treat as APPROVE.

### 1. Default-build neutrality — PASS
- `gate_stats` module, all four `SKIP_*`/`SELF_RESUME_READY` increments, and the two `SELF_RESUME_READY` bumps in the self-resume arms are each behind `#[cfg(feature = "sched-gate-stats")]` (`sched/mod.rs:462`, `:1909`, `:1915`, `:2270`, `:2424`, `:2616`). With the feature off they compile out entirely — the pick hot path carries zero added loads.
- The mirror-filter closure was restructured from a single-expression `.filter(|&idx| matches!(...))` to a block `{ let on_this_cpu = matches!(...); #[cfg…] if …; on_this_cpu }` (`sched/mod.rs:2421-2432`). With the feature off this is `let on_this_cpu = matches!(…); on_this_cpu`, which is semantically identical to the original and folds to identical codegen under `-O` (DCE of the dead `let`).
- kdb `cpu-state` (`kdb.rs:1911-1925`) gains the `sched` object + per-thread `rst` on **all** builds. This is an on-demand diagnostic command, not a hot path; the added cost is 3 relaxed atomic loads (`maintain_passes`/`context_switches`/`starve_force_count`) + a few output bytes per invocation. Acceptable. `MAINTAIN_PASSES` already existed and is already incremented in `mirror_maintain`; the PR only adds the `maintain_passes()` getter (`percpu.rs:811`).

### 2. Latent fix correctness under the SMP invariants — PASS
- The promoted thread is `current_tid = proc::current_tid()` (`sched/mod.rs:2144`), i.e. the thread executing `schedule()` on **this** CPU, so it is genuinely on-CPU in both arms (len<=1 `:2259`, `None` `:2602`). Marking it `Running` is the correct incoming-role transition; it is simultaneously the outgoing thread but it does **not** switch out, so leaving `ctx_rsp_valid` untouched is correct (the saved `context.rsp` is stale precisely because the thread stays live on its own stack).
- Cross-CPU: a sibling CPU can only pick a **Ready** thread (`select_next_core` gates `if t.state != Ready continue` at `:1902`), and even in the stale-`ctx_rsp_valid`=true / Ready window the #655 on-CPU interlock `defer_if_live_on_other_cpu` (`:131` → `is_tid_live_on_other_cpu`) defers any candidate that is current/on-stack on another CPU. `current_tid` is current on this CPU, so a sibling cannot dispatch it. Promoting Ready→Running only strengthens exclusion — this is a strict improvement, never a weakening, of the #672/#673 interlock. The pre-fix code already left the thread `Ready`-while-running; the fix makes the table *more* consistent, not less.
- No race with the lazy stamping pass or `mirror_maintain`: both run earlier in the same locked pass (`:2357`, `:2381`) and the promotion writes (`state`/`ready_since_tick`/`last_cpu`) happen under the still-held `THREAD_TABLE` guard before `drop(threads)`. All readers of these plain fields go through `THREAD_TABLE`, so no torn read.

### 3. Semantics of the promoted state — PASS
- The normal `Some(idx)` pick sets exactly `{state=Running, last_cpu=cpu, ready_since_tick=0}` for the selected thread (`:2503-2508`); `resume_current_if_ready` (`:2101-2115`) applies the same three writes. There is no per-thread quantum/deadline counter in the pick path, so nothing is skipped. `NEED_RESCHEDULE` is already cleared at schedule() entry (`:2131`), so the in-place return path needs no extra clear.
- No tick-preemption escape: the self-resumed thread keeps running and is preempted normally on the next timer tick. No STARVE force-accounting escape — on the contrary, clearing the stale `ready_since_tick` **prevents** a spurious future force-select. Traced the next-preemption cycle (stamping pass sees `Running`→stamp stays 0; `Some(idx)` outgoing `Running` branch flips to `Ready` with stamp 0; next pass re-stamps `now`) — wait age is correctly measured fresh. The pre-fix path kept the ancient stamp, which is the exact regression the fix targets.
- No reap/off-stack interaction: `reap_dead_threads_sched` only touches `Dead` threads; a `Running` current is never a reclaim candidate.

### 4. Counter correctness — PASS
- `SKIP_CTX_RSP_INVALID` / `SKIP_AFFINITY` are placed after the `state != Ready` gate (`:1902`) and immediately before their respective `continue`s (`:1908-1911`, `:1913-1918`), so they count only Ready threads, exactly as documented.
- `SKIP_MIRROR` (`:2424-2430`) counts Ready non-idle (`tid < 0x1000`) threads whose `mirror_slot` != this CPU — matches its doc comment; `mirror_maintain` runs earlier in the same pass so on SMP=1 this is ~0 (consistent with the reported 0).
- `SELF_RESUME_READY` (`:2270-2273`, `:2616-2619`) increments iff the pre-promotion `observed` state was `Ready` — exactly one increment per self-resume. `resume_current_if_ready` returns the pre-promotion state so all downstream match arms are unchanged.
- Relaxed atomics are fine for diagnostic counters. **NIT (non-blocking):** the gates are counted first-gate-wins (mirror → interlock[uncounted] → ctx_rsp → affinity), so the four counts are mutually exclusive by gate order, not independent per-gate populations, and the #655 interlock deferral is deliberately uncounted. A one-line note near the counters would save a future investigator from reading them as independent populations.

### 5. Test 681 quality — PASS
- Deterministic: drives `test_resume_current_if_ready` directly over a synthetic thread slice, no live scheduler. CI "kernel smoke (QEMU boot + test suite)" is green, so it compiles and passes.
- Case A pins the full invariant (Ready→Running, stamp→0, last_cpu 0xFE→0, ctx_rsp_valid untouched, observed==Ready); Cases B/C prove Running/Blocked/Sleeping/Dead currents are left byte-for-byte untouched incl. the 0xFE `last_cpu` sentinel and the 777 stale stamp. This meaningfully pins the stamp-clearing behaviour.
- **NIT (non-blocking):** the test exercises the helper in isolation, not full `schedule()` integration (correctly — the live pass can't be unit-tested deterministically), and doesn't assert `SELF_RESUME_READY` (a feature-gated diagnostic). Acceptable coverage.

### Nits summary (all non-blocking)
1. `sched/mod.rs:2421` — "byte-identical" is accurate only for the *instrumentation* (which compiles out); `firefox-test-core`/default do carry the latent fix vs base master. The PR body discloses this, but tightening the phrase to "instrumentation compiles out; the only behaviour delta from master is the latent self-resume fix" would prevent misreading.
2. `sched/mod.rs:~1908` — add a one-line note that the gate-skip counts are first-gate-wins / mutually-exclusive and the #655 interlock deferral is uncounted.
3. `test_runner.rs:test_681` — optional: a comment that this is a helper-level unit test (full self-resume integration is exercised live by the A/B), not a defect.

Cross-cutting checks: 5/5 CI green; MERGEABLE/CLEAN; the two converted self-resume sites are the only in-place return-to-current arms (the other `find(t.tid==current_tid)` sites at `:2171`/`:2479`/`:2738`/`:2872`/`:2922` are the canary read and the post-`break 'pick` switch bookkeeping — correctly untouched). Public-spec citation (POSIX sched(7)) is appropriate; no private-corpus leakage in the committed prose.

PR #686, opened 2026-07-01
ci: stop perf/timing flakes from flipping the kernel-smoke gate

me1iissa:
## Problem

The `kernel smoke (QEMU boot + test suite)` gate was flipped **twice today, on two different unrelated PRs**, by tests that are load-sensitive on shared GitHub runners (no `/dev/kvm`, TCG-only) and were **not** in `ci/allow-fail.json` — while their sibling env-gap failures (`dynamic_elf`, `pie_elf`, `TCC compile`, `busybox basic`, `firefox_oracle`, `test_522_tcp_rx_poll_bell`) already are. Both incidents had the textbook flake signature: **red on the first run, green on a same-head rerun.**

| Incident | Run | Result |
|---|---|---|
| PR #681 | `28548793834` | red → green on same-head rerun |
| PR #683 | `28549994543` | red → green on same-head rerun |

Three tests are involved:

1. **`log_ring write cost`** — `cycles/call` vs an 8× relative RAM-copy ceiling + a 50000-cyc backstop. Observed **6284 vs 6016 ceiling** (4.4% overshoot) on #681 and **126501 cyc/call** on #683 — a pure runner-load artifact (~168× the healthy ~750-cyc baseline).
2. **`log_ring A/B`** — same family (single-shot ring-vs-COM1 ratio ≥ 10×).
3. **`execve_leak`** — `leaked 38 pages > 24`, an async-GC timing flake (38 was previously observed flaking on a comment-only diff, green on rerun).

## Fix (gating hygiene, not a perf-test redesign)

Decided per test — **widen where a meaningful wider bound exists, allowlist where the test is inherently unmeasurable on shared runners**:

### `execve_leak` — **widen** (option a)
Not a timing test; it counts leaked PMM pages. The flake is asynchronous background reclamation (dead-thread kstack recycling + deferred GC) that has not fully drained at the sample point even after `wait_dead_stacks_quiesced()`. Widen `TOLERANCE_PER_ITER` **8 → 16** (total **24 → 48**).

Still meaningful: the regressions this test guards against are far larger — a `free_vm_space()` reverting to a no-op leaks the whole old VmSpace every iteration (≥ 16 eager user-stack frames + code/data + page tables ≈ 54–60 total across 3 iters) and a non-recycled kernel stack is `KERNEL_STACK_PAGES = 64`. Both exceed 48, so a genuine per-exec leak still **fails**; only the sub-64-page async-GC jitter is tolerated. Ordering: healthy `<24 < 38 (observed flake) < 48 (new bound) < 54 (smallest real regression)`.

### `log_ring write cost` + `log_ring A/B` — **allowlist** (option b)
Both are `rdtsc` cycle-budget assertions whose measurement window includes host wall-clock time whenever the runner preempts the KVM-less guest mid-batch. The VM-exit-per-byte regression they guard against (Intel SDM Vol. 3C §25.1.3) **only manifests under KVM, which GitHub runners lack**. On TCG the noise dynamic range (~130×) buries any regression a widened CI bound could still catch — a ceiling loose enough to pass 126501 cyc/call would be decorative — so there is no meaningful CI bound. Allow-list them with a reason + today's evidence.

- Ring framing / record-count correctness stays gated by the deterministic **`log_ring record/drain round-trip`** test (untouched).
- `ci/allow-fail.json` is **CI-only**: it is rendered into `watch-test.py --allow-fail` in `build.yml` and applied only there. Local and KVM runs — where the VM-exit-cost property is real and the box is quiet — **still gate** both tests.

## Verification
- `ci/allow-fail.json` parses; `qemu-harness.py allowlist regex` renders both new names and matches the exact `[FAIL] log_ring …` lines the tests emit, while `execve_leak` and unrelated `[FAIL]`s still GATE.
- Both build profiles compile clean: `test-mode` (the CI test profile, `ok: true`) and `firefox-test-core` (`ok: true`), via `scripts/qemu-harness.py build`.

BUILD-ONLY; no local QEMU sessions (another agent owns the box). GitHub CI verifies the boot.


me1iissa (commented):
Coordinator review: APPROVE-WITH-NITS.

- allow-fail entries: reasons are evidence-cited (both same-head-rerun incidents referenced); the CI-only gating mechanism claim verified against build.yml's use of the rendered allowlist regex — local/KVM runs still gate both tests where the VM-exit-cost premise holds. Correct call vs widening: a bound loose enough to pass 126501 cyc/call would be decorative.
- execve_leak 8→16/iter: ordering argument is sound (healthy <24 < observed flake 38 < bound 48 < smallest genuine regression class 54–64). Nit: a partial leak of exactly the eager user-stack frames (16/iter = 48 total) sits at the bound — same character as the old bound, acceptable for a hygiene widening.
- Nit: tracking: null on the two new entries — worth opening tracking issues if these tests are ever expected to gate in CI again (e.g. if CI gains KVM runners).

PR #687, opened 2026-07-01
tooling: build + boot a glibc-Firefox data image without clobbering the shared musl image

me1iissa:
## What

Make the glibc Firefox build a first-class, **isolated** artefact so the "does the ~50x wiki-load slowdown reproduce on the glibc build?" discriminator can run without disturbing the shared musl demo image.

## Why

`build/data.img` carries only the musl firefox-esr. `--firefox-variant glibc` previously either no-op'd (the glibc binary isn't on the image) or forced a destructive in-place `create-data-disk.sh --force` that overwrites the shared musl image every session. The glibc `libxul.so` (BuildID `492cd905…`) already has a fully-named `.symtab` companion, so the right move is to stage that binary onto its **own** image and boot it directly.

## Changes (all additive; default behaviour unchanged)

**`create-data-disk.sh` + `install-*.sh`**
- `--output=PATH` / `$ASTRYXOS_DATA_IMG` — write the image somewhere other than `build/data.img` (e.g. `build/data-glibc.img`).
- `--build-dir=PATH` / `$ASTRYXOS_BUILD_DIR` — stage into `PATH/disk` instead of `build/disk`. Exported to the `install-glibc/firefox/stubs/fonts/dbus` + `inject-libxul-symbols` sub-scripts (same override pattern already used by `patch-gui-caches.sh` / `patch-gui-gl.sh`) so the whole glibc pipeline stays in one isolated staging root and never overwrites the shared `build/disk` musl tree. Precedence: flag > env > `${ROOT_DIR}/build`.
- Upstream binaries are staged **as-shipped**; the wrappers are unchanged in what they install.

**`qemu-harness.py` (new JSON fields only — no renames)**
- `--firefox-variant glibc` auto-selects a **separate** prebuilt image (`build/data-glibc.img`, or `$ASTRYX_GLIBC_DATA_IMG`) when present and boots it **non-destructively** via a new `_launch_qemu_harness(data_img_override=…)` parameter — the shared `build/data.img` file is never unlinked, re-symlinked, or regenerated. On-disk staleness / variant-regen is skipped for a prebuilt image (authoritative, like `--data-img`). An explicit `--data-img` still wins.
- New pure, host-unit-testable helper `_variant_data_img_candidate`; `firefox_variant_info.prebuilt_data_img` records the booted image.

**`qemu-harness-smoke.py`** — new host-only (no-QEMU) check `run_variant_data_img_check` locking down the selection precedence, hermetic against a real image on the host.

## Failure mode this prevents

The pre-existing `--data-img` override unlinks `build/data.img` and replaces it with a symlink — fine in a throwaway worktree, destructive in the canonical checkout, where a later musl re-stage would write **through** the symlink and clobber the glibc image too. The glibc path deliberately avoids that mechanism, so a musl session and a glibc session can run back-to-back on the same tree with neither image corrupting the other.

## Verification (no QEMU boot — another agent owns the box)

Built `build/data-glibc.img` with the patched scripts (isolated `ASTRYXOS_BUILD_DIR`):
- `e2fsck -fn`: clean (239 files, 0 errors).
- `readelf -n` on the image's `libxul.so`: **BuildID `492cd905271cba58d3544dcbc1d3de79e9319d2b`** — matches the fully-named `.symtab` companion (398,675 FUNC symbols injected; `.symtab`/`.strtab` present).
- ld.so closure: `firefox-bin` interpreter `/lib64/ld-linux-x86-64.so.2` present; all 49 `libxul.so` `DT_NEEDED` sonames resolve within the image.
- Kernel variant probe: only `/opt/firefox` present, `/usr/lib/firefox*` absent → kernel selects glibc.
- Shared musl `build/data.img` byte-identical (mtime unchanged); `build/disk/` untouched.
- Host-only smoke: `qemu-harness-smoke.py --staleness-only` → exit 0 (120 checks, incl. 6 new selection cases).

**A boot still verifies:** that the glibc Firefox actually runs to the wiki-load stage on AstryxOS and whether the ~50x slowdown reproduces. Image is a build artefact (not committed).

### DIFF
--- a/scripts/create-data-disk.sh
+++ b/scripts/create-data-disk.sh
@@ -16,10 +16,33 @@
 #   ./scripts/create-data-disk.sh 128       # Create 128 MiB image
 #   ./scripts/create-data-disk.sh --force   # Recreate even if it exists
 #
+# Isolated / alternate-image builds (additive — default behaviour unchanged):
+#   --output=PATH        Write the produced image to PATH instead of
+#                        build/data.img (env: ASTRYXOS_DATA_IMG).  Lets a
+#                        variant image (e.g. build/data-glibc.img) be built
+#                        WITHOUT clobbering the shared build/data.img.
+#   --build-dir=PATH     Stage into PATH/disk instead of build/disk (env:
+#                        ASTRYXOS_BUILD_DIR).  Exported to the install-*.sh
+#                        sub-scripts so a variant build never overwrites the
+#                        shared build/disk staging tree.  When --output is
+#                        omitted the image defaults to PATH/data.img.
+# Example — build a glibc Firefox image in an isolated tree:
+#   ASTRYXOS_FIREFOX_VARIANT=glibc ASTRYXOS_BUILD_DIR="${PWD}/build-glibc" \
+#     ./scripts/create-data-disk.sh --output="${PWD}/build/data-glibc.img" --force
+#
 set -euo pipefail
 
 ROOT_DIR="$(cd "$(dirname "$0")/.." && pwd)"
-BUILD_DIR="${ROOT_DIR}/build"
+# BUILD_DIR (staging root) and the output image are overridable via env or the
+# --build-dir / --output flags so a variant image can be produced in isolation.
+# The sub-scripts (install-glibc.sh, install-firefox.sh, ...) honour the same
+# ASTRYXOS_BUILD_DIR env, so exporting it below keeps the whole pipeline pointed
+# at one staging root.  Precedence: flag > env > ${ROOT_DIR}/build default.
+BUILD_DIR="${ASTRYXOS_BUILD_DIR:-${ROOT_DIR}/build}"
+# Output-image override captured here from env; the --output flag replaces it in
+# the arg loop.  DATA_IMG is finalised (from this override or ${BUILD_DIR}/data.img)
+# after BUILD_DIR is known — see the normalisation block below the arg loop.
+DATA_IMG_OVERRIDE="${ASTRYXOS_DATA_IMG:-}"
 DATA_IMG="${BUILD_DIR}/data.img"
 SIZE_MB=2048
 FORCE=false
@@ -178,6 +201,8 @@ for arg in "$@"; do
         --firefox-debug=*) FIREFOX_DEBUG="${arg#--firefox-debug=}" ;;
         --firefox-debug)   FIREFOX_DEBUG=1 ;;
         --firefox-package=*) FIREFOX_PACKAGE="${arg#--firefox-package=}" ;;
+        --output=*)    DATA_IMG_OVERRIDE="${arg#--output=}" ;;
+        --build-dir=*) BUILD_DIR="${arg#--build-dir=}" ;;
         --xeyes) XEYES=1; FORCE=true ;;
         --busybox) BUSYBOX_CLI=1; FORCE=true ;;
         --sshd) SSHD=1; FORCE=true ;;
@@ -198,6 +223,22 @@ case "${FIREFOX_VARIANT}" in
         ;;
 esac
 echo "[DATA-DISK] Firefox variant: ${FIREFOX_VARIANT}"
+
+# ── Finalise the staging root + output image (after --build-dir / --output) ───
+# Export ASTRYXOS_BUILD_DIR so every install-*.sh sub-script stages into the
+# SAME (possibly isolated) build root rather than its own ${ROOT_DIR}/build.
+# This is what keeps a variant build (e.g. glibc) from overwriting the shared
+# build/disk musl staging tree.  Output-image precedence: --output flag /
+# ASTRYXOS_DATA_IMG env > ${BUILD_DIR}/data.img.
+export ASTRYXOS_BUILD_DIR="${BUILD_DIR}"
+if [ -n "${DATA_IMG_OVERRIDE}" ]; then
+    DATA_IMG="${DATA_IMG_OVERRIDE}"
+else
+    DATA_IMG="${BUILD_DIR}/data.img"
+fi
+mkdir -p "$(dirname "${DATA_IMG}")" 2>/dev/null || true
+echo "[DATA-DISK] Staging root: ${BUILD_DIR}"
+echo "[DATA-DISK] Output image: ${DATA_IMG}"
 
 case "${FIREFOX_PACKAGE}" in
     firefox-esr|firefox) ;;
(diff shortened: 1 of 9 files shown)


me1iissa (commented):
## Pre-merge review — PR #687 (`tools/glibc-data-image`, cca2292d)

**Verdict: APPROVE-WITH-NITS.** Additive, default-neutral, and the core no-clobber claim holds — in fact it is *stronger* than the PR states. No blocking findings.

### What I verified

**1. No-clobber under adversarial sequencing — holds.**
I enumerated every writer of `build/data.img` / `build/data-glibc.img`:
- `create-data-disk.sh` — `dd` + `mke2fs "${DATA_IMG}"` (create-data-disk.sh:948, :977) where `DATA_IMG` = `--output`/`$ASTRYXOS_DATA_IMG` else `${BUILD_DIR}/data.img`. One deterministic target per invocation.
- Harness `--data-img` block — `unlink()`+`symlink_to()` (qemu-harness.py:3668-3671). This is the pre-existing destructive symlink landmine. **The glibc prebuilt path does not reach it**: `_variant_data_img_candidate` returns `None` whenever `explicit_override` is set (qemu-harness.py:1895-1896), and the block is gated on `if _data_img_override:` (3653), so glibc-prebuilt and `--data-img` are mutually exclusive.
- Harness variant-regen — calls `create-data-disk.sh --force` in place; gated off for prebuilt by `and _prebuilt_data_img is None` (qemu-harness.py:3880) plus `no_regen_data_img=True` (3634).
- **QEMU boots do not write the image at all** — the data disk attaches `snapshot=on` (astryx_qemu.py:337) or a qcow2 overlay backed read-only (astryx_qemu.py:330-334). So *two concurrent glibc starts on the same `data-glibc.img` cannot corrupt it*, and no boot ever mutates the shared `build/data.img`.

Result: for the glibc-prebuilt path the image files never cross-clobber under any enumerated ordering (glibc-build-while-musl-staged, concurrent glibc starts, prior `--data-img` session). With `--firefox-variant glibc` + a present prebuilt image, `build/data.img` is provably never opened for write, unlinked, symlinked, or regenerated.

**2. Default-path neutrality — confirmed byte-path-identical.**
- `--firefox-variant` default `"musl"` (qemu-harness.py:12910) → `_variant_data_img_candidate` returns `None` (variant≠glibc) → every new gate (`and _prebuilt_data_img is None`) evaluates to legacy behaviour.
- `create-data-disk.sh` with no env/flags: `BUILD_DIR=${ROOT_DIR}/build`, `DATA_IMG=${BUILD_DIR}/data.img` — same paths as before. The new `export ASTRYXOS_BUILD_DIR="${BUILD_DIR}"` resolves to the same `${ROOT_DIR}/build` the sub-scripts already default to (`${ASTRYXOS_BUILD_DIR:-${ROOT_DIR}/build}`), so it is a no-op in the default case.
- CI (`build.yml:292`, `:449`) invokes `create-data-disk.sh --force` with no env/flags → unaffected.

**3. Selection correctness — correct, with one documented sharp edge (nit #1).**
- glibc + prebuilt absent → `_prebuilt_data_img=None`, falls to legacy variant-regen which rebuilds the shared image *as glibc* and boots glibc (or, under `--no-regen-data-img`, warns loudly and boots the mismatch, banner qemu-harness.py:3931). **Not a silent musl boot.**
- `--data-img` + `--firefox-variant glibc` → `--data-img` wins (candidate returns `None` on `explicit_override`); documented in the docstring and help. Sane.
- JSON additive: `firefox_variant_info.prebuilt_data_img` is a new key on the existing dict (qemu-harness.py:3775); `data_img_override` is a new kwarg. No renames.

**4. inject-libxul-symbols.sh — invariant-clean.**
Targets only `${BUILD_DIR}/disk/opt/firefox/libxul.so` (the *staged copy inside the built image*, inject-libxul-symbols.sh:87/:105), never a shipped/cached original. Pre-existing script (first added in #295 `1366b9eb`), not new here — the `.symtab`-into-image-copy pattern is established repo practice, text section byte-identical. No policy flag needed.

**5. Smoke check — solid and hermetic.**
`run_variant_data_img_check` (qemu-harness-smoke.py:607) monkeypatches `_CANONICAL_BUILD_DIR` to a sandbox so a real host `data-glibc.img` cannot leak, and pins musl→None, glibc-absent→None, glibc-sibling→selected, explicit-override→None, env-present→selected, env-missing→sibling-fallback. Ran it host-only: **7/7 PASS, `failures []`**. `py_compile` + `bash -n` clean on all 9 files.

### Nits (none blocking)

1. **(Top) Conditional non-destructiveness.** `--firefox-variant glibc` with `data-glibc.img` absent and no `--no-regen` silently falls back to the *destructive* in-place regen of the shared `build/data.img` (qemu-harness.py:3877-3885). Documented in help and it matches prior behaviour (no regression), but an agent expecting the non-destructive path could clobber the shared musl image by forgetting to build `data-glibc.img` first. Suggest a distinct stderr line when glibc is requested + no prebuilt image + a shared-image regen is about to run, so "the non-destructive path was NOT taken" is loud.

2. **Half-set-env staging pollution.** In `create-data-disk.sh`, setting `ASTRYXOS_DATA_IMG`/`--output` *without* `ASTRYXOS_BUILD_DIR`/`--build-dir` writes the image to the override path but stages into the *shared* `build/disk` (create-data-disk.sh:217-227, `STAGING_TREE="${BUILD_DIR}/disk"`). No image-file clobber (`build/data.img` the file is not rewritten), but the shared staging tree inherits variant content that a later shared regen could pick up. Consider defaulting `BUILD_DIR` to `dirname(--output)` when only `--output` is given, or warning on the asymmetric combination. The documented recipe sets both, so this is a foot-gun, not a live bug.

3. **Hardcoded `_CANONICAL_BUILD_DIR = /home/ubuntu/AstryxOS/build`** (qemu-harness.py:1866). `exists()`-guarded so it degrades gracefully on other hosts/CI, and consistent with the existing hardcode at qemu-harness.py:3683 — mentioning for awareness only.

4. **Smoke coverage of candidate #3.** In `_run_variant_data_img_cases`, `di.parent == _CANONICAL_BUILD_DIR == tdp`, so the "sibling" image and the canonical fallback resolve to the same path — candidate #3 (canonical tree) is never exercised independently. Consider placing `di` in a distinct subdir to pin the canonical-fallback branch (qemu-harness-smoke.py:645-651).

5. **(Cosmetic)** the glibc prebuilt banner box has slightly ragged right borders due to em-dash width (qemu-harness.py:3637-3644).

PR #688, opened 2026-07-02
NDE-31: AF_UNIX socket-table diagnostic + abstract-namespace regression tests (glibc display-gate finding)

me1iissa:
## Summary

Investigation of the **glibc-Firefox `Error: cannot open display: :0`** gate (which blocks the glibc load-discriminator experiment), plus the reusable tooling and regression guard that came out of it.

**Key finding (the display gate is NOT a socket-layer bug):**
- A per-syscall trace of the glibc firefox launch shows it makes **zero `socket(2)`/`connect(2)` calls** before failing — it never reaches the AF_UNIX layer. `gdk_display_open(":0")` returns before any socket.
- A side-by-side **musl-firefox** trace on the same kernel performs the exact libxcb connection sequence correctly: connect to the **abstract** name (`sun_path[0]==NUL`) is refused with `ECONNREFUSED` per unix(7); the **pathname** fallback connect to `/tmp/.X11-unix/X0` then succeeds and the X server accepts the client (`[X11] client fd=2`).
- Therefore our AF_UNIX abstract+pathname handling is **already correct**, and the dispatch's abstract-fallback hypothesis is **refuted**. The glibc gate is a **pre-socket divergence in the glibc-linked X-client stack** (system libxcb / libX11 / GTK / glibc), out of the socket layer's scope — flagged for the libc/loader/toolkit owners.

This PR **does not change socket behaviour** and **does not (cannot) fix the glibc display gate**. It lands:

- **kdb `unix-table`**: live occupancy of the global AF_UNIX socket TABLE (`occupied` / `capacity` / `high_water` / `full` + per-slot `state`/`kind`/`peer`/`name`, abstract names shown `@name`). A full table makes `socket(AF_UNIX)`/`socketpair(2)` return `EMFILE` (surfaces as "cannot open display"); this op attributes an exhaustion to its consumers. Used here to **refute a table-exhaustion hypothesis** — measured peak occupancy was **1**, not 64.
- **net::unix high-water mark**: since-boot peak of simultaneously-live slots, folded in at each allocation site under the already-held TABLE lock (diagnostic only).
- **Regression test** (`test_runner` Test 54.5): abstract bind/listen/connect/accept roundtrip; connect to an unbound abstract name → `ECONNREFUSED` (never `ENOENT`, per unix(7)); pathname fallback connect after abstract refusal (the libxcb sequence); same-fd reconnect after `ECONNREFUSED` (POSIX.1-2017 `connect()`).
- **qemu-harness**: expose the `unix-table` kdb op.

## Verification
- `--features test-mode`: new Test 54.5 **passes** (all four phases).
- Live **musl smoke** with this kernel: X connection still established (abstract refused → pathname accepted → `[X11] client fd=2`), **zero regression**.
- Live **glibc**: still `cannot open display` (expected — the pre-socket userspace divergence is not addressed here), `unix-table high_water == 1`.

## Spec references
unix(7) (abstract vs pathname sun_path, ECONNREFUSED vs ENOENT), POSIX.1-2017 `connect()` (reconnect after failed connect).

## Do not merge yet
Self-authored; requesting independent review first.

### DIFF
--- a/kernel/src/kdb.rs
+++ b/kernel/src/kdb.rs
@@ -369,6 +369,10 @@ pub fn dispatch(req: &str, out: &mut String) {
         "proc-tree"      => op_proc_tree(req, out),
         "fd-table"       => op_fd_table(req, out),
         "fd-map"         => op_fd_map(req, out),
+        // unix-table: live occupancy of the global AF_UNIX socket TABLE.
+        // A full table makes socket(AF_UNIX)/socketpair(2) return EMFILE,
+        // which surfaces to X clients (libxcb) as "cannot open display".
+        "unix-table"     => op_unix_table(out),
         "epoll-watch"    => op_epoll_watch(req, out),
         "syscall-trend"  => op_syscall_trend(req, out),
         "vfs-mounts"     => op_vfs_mounts(out),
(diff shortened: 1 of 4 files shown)


me1iissa (commented):
## Independent pre-merge review — VERDICT: APPROVE-WITH-NITS

Reviewed adversarially against the "diagnostic-only, no socket-behaviour change" claim. All five review dimensions pass; the nits below are non-blocking (doc precision + test hygiene). CI 5/5 green, `mergeStateStatus: CLEAN`. File list matches the described scope exactly (4 files; no silent bundling).

### 1. "Diagnostic-only" claim — HOLDS
`note_occupancy()` (`kernel/src/net/unix.rs:34`) is a pure read — `t.0.iter().filter(|s| s.state != Free).count()` + an `AtomicUsize::fetch_max(Relaxed)` — executed under the **already-held** `TABLE` lock at all three call sites (`create` :320, `connect` :461, `socketpair` :526). No new lock is taken, no lock-order change, and no return value / errno / wake-ordering is altered:
- `create()` (:304-321): the `return i` → `allocated = i; break;` refactor is behaviour-preserving — same slot index on success, `u64::MAX` on exhaustion, and `note_occupancy` is correctly skipped on the exhaustion path (no phantom high-water bump).
- `connect()` (:461): the note is inserted **before** the existing `drop(t)` + `ring_poll_bell_for(...)`, so the drop-before-ring poll-bell ordering the surrounding comment documents is untouched.
- The added scan is O(64) state-field reads — the same stride these paths already pay to find a `Free` slot — so cost is negligible.

High-water accuracy verified: the only `Free → live` transitions are create (:310), connect-peer (:426), socketpair a/b (:499/:508) — all three noted. `bind`/`listen`/`accept` mutate already-live slots and are correctly **not** allocation events, so the counter never undercounts a peak.

### 2. kdb op safety — HOLDS
`op_unix_table` (`kernel/src/kdb.rs:2459`) takes the `TABLE` lock **only** transitively via `snapshot_all()`, which releases it before any formatting; `capacity()` is a const and `high_water()` is an atomic load — neither locks. All JSON string-building (the allocating part) happens with **no lock held** → no serial-print-under-lock, no format-under-lock. The per-slot loop is bounded to ≤ `MAX_UNIX_SOCKETS` (64). JSON is well-formed (verified `j_kv`/`j_kv_str` trailing-comma + the inline trim before `}`; empty table → `"slots":[]`).
- Note (pre-existing, not introduced here): `snapshot_all()` (`net/unix.rs:1026`) does allocate a bounded `Vec` under the `TABLE` lock — identical to the established `fd-map` pattern that already reuses it. Not a regression; no allocator→AF_UNIX lock cycle exists.

### 3. Test 54.5 correctness — HOLDS (contracts are right AND independently satisfied)
- **Phase 2** — unbound *abstract* name → `-111 ECONNREFUSED`, never `ENOENT`: correct per unix(7) (an abstract name has no filesystem object). `connect()` :392-394 returns `-111` for `abstract_ns && no-listener`, and `-2 ENOENT` only for a missing *pathname* (:401) — the right two-mode split. The test asserts the correct contract, and the impl satisfies it independently (the `-111` branch is an early return that does not mutate the client slot).
- **Phase 4** — same-fd reconnect after `ECONNREFUSED` succeeds: genuinely satisfied *because* the phase-2 `-111` path returns before `t.0[id].state` is ever written (:450), so `c1` stays `Unbound` and is freely re-connectable. Not fitted-to-code.
- `sun_path`/addrlen handling is spec-correct end-to-end: bind glue (`subsys/linux/syscall.rs:3011-3020`) forwards `sun_path[..addrlen-2]`, and `unix_name()` (:287) keeps the full addrlen-derived length + embedded NULs for abstract names — so the exact-addrlen construction the test documents is the property actually under test.

### 4. Additivity + gating — HOLDS
- Harness: `unix-table` added to the zero-arg op list (`qemu-harness.py:7181`) and the argparse `choices` (:13149) — additive, correct placement (the kdb op parses no request fields). No renamed fields.
- Gating is the right split: the **reader** `op_unix_table` lives in `kdb.rs` (`#![cfg(feature = "kdb")]`, :5) so it only compiles into kdb builds; the **counter** (`note_occupancy`/`UNIX_HIGH_WATER`) is unconditional — a bounded always-on O(64) counter, which the CI "bootloader + kernel builds" (non-kdb) job confirms compiles clean (no dead-code failure).

### 5. CI + scope — HOLDS
All 5 checks pass; diff is exactly `kdb.rs (+70)`, `net/unix.rs (+32/-2)`, `test_runner.rs (+119)`, `qemu-harness.py (+7)`. Nothing bundled beyond the stated scope.

---

### Nits (non-blocking — land as-is or fold in a follow-up)
1. `net/unix.rs:24-26` — the `UNIX_HIGH_WATER` doc says "create/**accept**/connect-peer allocation", but `accept()` does not allocate (the accept-side slot is created in `connect()` and merely popped from the backlog at accept). The counter is correct; the word "accept" is imprecise. Suggest: "create / connect-peer / socketpair".
2. `test_runner.rs` phase 4 — the comment cites "POSIX.1-2017 connect()" for reconnect-after-failure, but POSIX leaves the socket state *unspecified* after a failed `connect()` (and advises portable apps to close+recreate). The reconnect-succeeds guarantee is Linux-specific. The assertion is correct for a Linux-ABI target; just cite it as Linux behaviour rather than POSIX.
3. `test_runner.rs` phases 3-4 — the fds returned by the `accept()` calls are not closed (2 leaked table slots + fds). Harmless (64-slot table, one-shot), but tidy for hygiene.
4. `kdb.rs:2521` — the inline `if out.ends_with(',') { out.pop(); }` duplicates the existing `j_trim_comma(out)` helper (`kdb.rs:344`); minor consistency nit.

Spec references: unix(7) (abstract vs pathname `sun_path`; `ECONNREFUSED` vs `ENOENT`), POSIX.1-2017 `connect()`.

(Posted as a comment rather than an Approve because the review account matches the PR author; the substantive verdict is **APPROVE-WITH-NITS**.)

PR #689, opened 2026-07-02
subsys/linux: opt-in file-op/write/return traces; root-cause glibc FF display-open gate (stub X libs, not a kernel bug)

me1iissa:
## Summary

Investigation into why the **glibc** Firefox variant dies in `gdk_display_open(\":0\")` with `Error: cannot open display: :0` before it ever calls `socket(2)` — the gate blocking the glibc startup discriminator experiment.

**Verdict: this is NOT a kernel syscall-fidelity bug and NOT an environment issue.** The `data-glibc.img` ships **no-op stub** X11/GTK client libraries (`libX11.so.6`, `libxcb.so.1`, `libgdk-3.so.0`, `libgtk-3.so.0`, …) generated by `scripts/install-firefox-stubs.sh` to satisfy the dynamic linker's NEEDED/versioned-symbol resolution. That script’s `gdk_display_open`/`gdk_display_get_default` are **deliberately** classified to return NULL (see its own comment: *“the canonical 'no display, run without one' signal”*). So the glibc build can never open a display — by design of the image, not any kernel defect.

This PR ships the additive diagnostic tooling that made the root cause findable in a single boot. It contains **no kernel behaviour change**.

## Evidence (live, glibc `data-glibc.img`, reproduced ×3 across relaunch attempts pid 1/2/3)

- `gdk_display_open(\":0\")` returns NULL making **zero syscalls**: in the return-annotated trace it is bracketed by `readlink(/proc/self/exe)=29` and the error `write(fd=2, …)=31` with nothing between. A real `gdk → XOpenDisplay → xcb_connect` path would issue `socket(2)`. Confirmed **0× `socket(41)` / `connect(42)`** in the whole ~1223-syscall run.
- Every file-op return in the pre-error burst is POSIX-contract-correct: `access` → ENOENT on absent paths, `mkdir` → EEXIST on existing dirs / 0 on new, `readlink /proc/self/exe` → 29, correct fds. Nothing diverges.
- No `Xlib: connection refused` / auth diagnostic on stderr → a silent, pre-socket, pure-userspace NULL.
- The `nspr_use_zone_allocator` `symbol lookup error … (fatal)` line in the loader log is a **non-fatal weak-symbol** miss — the process demonstrably continues (full `.mozilla` profile bootstrap runs afterward). Red herring, ruled out.

**Side-by-side control (musl, same kernel):** the `[SYSR]` trace shows musl FF doing `socket=7`, `connect=-111` (ECONNREFUSED, abstract `\0/tmp/.X11-unix/X0`), `socket=7`, `connect=0` (pathname) → `[X11] client fd=2` accepted. glibc makes **none** of these because its stub `gdk_display_open` returns NULL first. This is the whole difference.

## What this PR adds

Behind the existing opt-in `firefox-trace-verbose` feature (absent from the default and fast profiles), three self-contained syscall-exit diagnostics in `kernel/src/subsys/linux/syscall.rs`:

- `[FILEOP]` — decoded pathname + flags/mode + return for path-bearing/dir syscalls (syscall numbers per the published `syscall(2)`/`syscalls(2)` tables).
- `[WR]` — bounded, sanitised payload of `write`/`writev` to fd 1/2/3 (surfaces a launcher’s `LD_DEBUG_OUTPUT` narrative per `ld.so(8)` and any stderr diagnostic).
- `[SYSR]` — compact per-syscall return line paired with `[LINUX-SYS]`.

All are `#[cfg]`-gated off unless `firefox-trace-verbose` is selected; zero cost on the fast path.

## Verification

- Builds clean; **musl no-regression** confirmed: windowed FF still connects to X (`[FF/connect] \0/tmp/.X11-unix/X0` → `/tmp/.X11-unix/X0` → `[X11] client fd=2`).
- Traces verified live on both variants.

## Not fixed here (out of scope / by design)

Unblocking a real glibc render requires provisioning **real** glibc X11/GTK client libraries on `data-glibc.img` (a toolchain/image-build task) instead of the deliberate stubs. If only startup/DSO/event-loop timing is wanted, the glibc discriminator can run **headless** (drop `--ff-gui`; `MOZ_HEADLESS` skips `gdk_display_open`). No kernel change makes the stub libraries open a display.

### DIFF
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -1156,6 +1156,32 @@ pub fn dispatch(num: u64, arg1: u64, arg2: u64, arg3: u64, arg4: u64, arg5: u64,
     // [SC-RET] trace) rather than bypassing them.
     let ret: i64 = dispatch_body(num, arg1, arg2, arg3, arg4, arg5, arg6);
 
+    // ── Verbose file-op / write-content / return trace (opt-in) ──────────────
+    // Behind `firefox-trace-verbose` (never in the default or fast profiles):
+    // completes the existing `[LINUX-SYS]` entry line with the decoded path +
+    // flags/mode + return for file ops ([FILEOP]), the payload of writes to the
+    // small diagnostic channels ([WR]), and a compact return-value line paired
+    // with `[LINUX-SYS]` ([SYSR]).  Together these let a gate that leaves no
+    // failing syscall — e.g. a userspace library returning NULL with zero
+    // syscalls — be localised precisely.
+    #[cfg(all(feature = "firefox-test-core", feature = "firefox-trace-verbose"))]
+    {
+        if crate::proc::current_pid_lockless() >= 1 {
+            trace_fileop(num, arg1, arg2, arg3, arg4, ret);
+            trace_write_content(num, arg1, arg2, arg3, ret);
+            static RETN: core::sync::atomic::AtomicU64 =
+                core::sync::atomic::AtomicU64::new(0);
+            let n = RETN.fetch_add(1, core::sync::atomic::Ordering::Relaxed);
+            if n < 10000 {
+                crate::serial_println!(
+                    "[SYSR] #{} pid={} tid={} nr={} a1={:#x} a2={:#x} a3={:#x} ret={}",
+                    n, crate::proc::current_pid_lockless(),
+                    crate::proc::current_tid(), num, arg1, arg2, arg3, ret,
+                );
+            }
+        }
+    }
+
     // ── Close out the ring entry with the syscall's return value ─────────────
     #[cfg(feature = "syscall-trace")]
     {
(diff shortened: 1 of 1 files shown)


me1iissa (commented):
**Independent pre-merge review — APPROVE-WITH-NITS**

Additive, opt-in diagnostics in `kernel/src/subsys/linux/syscall.rs` (+137/-0, single file). Reviewed adversarially against the 5 dimensions; the change is structurally sound and CI-inert on every shipping/test profile. Merge-worthy; nits below are non-blocking (a couple worth a follow-up).

### 1. Gating totality — PASS (stronger than the PR body claims)
All three added emit blocks are gated `#[cfg(all(feature = "firefox-test-core", feature = "firefox-trace-verbose"))]` — matching the existing precedent at `syscall.rs:996` / `:1112` / `syscall/mod.rs:3445`.
- Block: `syscall.rs:1167`; `trace_fileop` def `:12262`; `trace_write_content` def `:12315`. All 4 `serial_println!` sites (`:1176`, `:12294`, `:12300`, `:12361`) are inside these gates — grep confirms zero stray/unconditional emit.
- The gate requires **both** features. `firefox-trace-verbose = []` (Cargo.toml:393) is standalone and is **not** pulled by `default`, `test-mode`, `firefox-test-core`, `firefox-test-trace`, or even the full `firefox-test` bundle. So even a full FF debug build stays silent unless `firefox-trace-verbose` is *explicitly* added. Nothing new in `default`/fast/test-mode — exactly as required.
- Helper fns are defined only under the same gate and referenced only from under it → no dead-code / unused-fn warning when the feature is off.

### 2. No behavior change — PASS
The block runs *after* `let ret = dispatch_body(...)` (`:1157`) and is purely observational: it reads memory + prints, and `ret` flows unmodified to `ring::end()` (`:1189`) and the return. It cannot alter return semantics, error paths, or partial-write accounting.
- `[WR]` double-fetches the user write buffer, which is fine: it's a post-dispatch read-only snapshot via `user_slice_snapshot` (`syscall/mod.rs:136`) — validates the pointer, SMAP-brackets, copies to an owned Vec, returns `None` on bad ptr (no fault). `[FILEOP]` paths go through `read_cstring_from_user` (`syscall.rs:476`) — bounded per-CPU arena slot, SMAP-bracketed, returns `&'static [u8]` (never lends a live user page). Both safe.
- Bounds: `[WR]` CAP=240B, writev capped at 8 iovs; `[FILEOP]` path bounded by the arena slot; `[SYSR]` counter-capped at 10000 total lines.

### 3. Hot-path cost when ON — acceptable (trace-verbose is the ~10x firehose profile anyway); one nit
Nothing pathological: allocations are bounded, `[SYSR]` is capped, `[FILEOP]` uses no heap. `[WR]` volume is comparable to the existing `[FF/write]`/`[FF/stderr]` mirror (the PR notes that was ~78% of a 45MB log), so it fits existing norms.

### 4. Data leakage — PASS
`[WR]` mirrors fd 0-3 content to serial — consistent with the existing `firefox-test-trace` stdout/stderr mirror. Serial only; no new persistent storage.

### 5. CI + scope — scope PASS; CI has a real coverage caveat
Diff is exactly `kernel/src/subsys/linux/syscall.rs` (+137/-0), matching the description. 3 checks green, 2 pending (kernel-smoke, qemu-harness-smoke) — merge should wait for those.

---
### Nits (non-blocking)

- **[Important caveat, not a blocker] CI does not compile the new code path.** Every CI job builds `default` (`cargo check`, `build.sh release`) or `test-mode` (kernel-smoke, qemu-harness-smoke). None enables `firefox-test-core`+`firefox-trace-verbose`, so the +137 lines are `#[cfg]`-stripped in all of them. **CI green here proves the change is inert on shipping/test profiles (the important guarantee) but does NOT prove it compiles.** That rests on the author's local build claim. I verified compile-correctness by inspection (identifiers resolve, format-arg arity matches, slice indices in-bounds since `user_slice_snapshot` returns exactly `len` bytes) and it looks clean — but a reviewer should not read the green checks as covering this code.

- **`trace_write_content` heap-allocs before the num guard (`syscall.rs:12320`).** `Vec::with_capacity(240)` runs for *any* syscall with `a1 ∈ [0,3]` and `ret > 0`, including non-writes (e.g. `mmap` with addr=0 → fd=0, ret=address; `read`/`lseek`/`dup`/`fcntl` on fd 0-3) that then hit the `else { return; }` at `:12357` and drop the Vec unused. Cheap fix: bail on `num != 1 && num != 20` (or move `with_capacity`) before allocating. Debug-profile-only, so minor.

- **fd range vs comment (`syscall.rs:12317`).** Guard is `fd < 0 || fd > 3` i.e. fd 0-3 *including stdin*, but the comment/PR say "fd 1/2/3". Harmless; code/doc mismatch.

- **`[WR]` snapshots `min(a3, CAP)` not `min(ret, CAP)` (`:12333`).** On a partial write it logs the requested-count region rather than the actually-written prefix, so logged content can exceed `n=`. Cosmetic (bounded + validated, no safety issue).

- **Asymmetric caps:** `[SYSR]` is counter-capped at 10000 but `[FILEOP]`/`[WR]` are uncapped. Fine — capping the highest-volume per-every-syscall line is the right choice; noting only for awareness.

Verdict: **APPROVE-WITH-NITS.** Correctly and conservatively gated, purely observational, zero risk to default/test-mode/CI. Recommend (a) let the 2 pending QEMU jobs go green and (b) optionally address the pre-num-guard alloc nit before or after merge.

PR #691, opened 2026-07-02
tooling: real glibc GTK3 + X11 closure for the glibc-Firefox image (real display-open)

me1iissa:
## What

New `scripts/install-gtk-real-glibc.sh` + a 21-line hook in `create-data-disk.sh` that overlay the **real glibc GTK3 + X11 client-library closure** (53 shared objects, ~23 MiB) onto the glibc-Firefox data image, replacing the headless no-op stubs for the display-critical libraries.

**Why:** the glibc variant currently satisfies its GTK3/X11/GLib/Cairo/Pango `DT_NEEDED` with stubs whose `gdk_display_open()` returns NULL by design → it can only run `--headless`. The "does the ~50× wiki-load slowdown reproduce on glibc Firefox?" discriminator needs a **real display-open** (the musl variant already has one via its real Alpine GTK/X11 closure).

## How

Same real-lib-over-stub paradigm as `install-fonts-real.sh` (real libfontconfig/libfreetype) and `install-dbus-real.sh` (real libdbus-1/libsystemd): stubs are built first, then real upstream binaries overlay them for the paths Firefox genuinely exercises. Kept as stubs on purpose: `libasound` (audio) and `libdbus-glib-1` (deprecated wrapper) — neither needed for display-open.

**Source:** upstream **Ubuntu 24.04 LTS ("noble")** binaries via `apt-get download` into an *isolated apt sandbox* (no system-wide install, no `/etc/apt` changes) + `dpkg-deb -x` into a reusable cache. noble is chosen because:
- its glibc (2.39) is **≤** the glibc the image ships (image carries host glibc via `install-glibc.sh`) → a GTK3 stack linked against 2.39 runs unchanged, **no version wall**;
- it is the reference glibc environment this Firefox ESR 115 build is known to render against;
- its `gdk-pixbuf` does not drag in a newer out-of-process image-loader sandbox → keeps the closure to plain in-process libraries.

Binaries ship **as-is — never patched**.

**Two closures are staged** (revised after review — head `7a5f128e`):
- **A. the `.so` DT_NEEDED closure** (53 shared objects, ~23 MiB) — a build-time `readelf` BFS from the display-critical roots ERRORS if any edge is unresolved. This check is **scoped to ELF DT_NEEDED edges only** and is **blind to dlopen'd modules (gdk-pixbuf loaders, GIO modules) and runtime data files (gschemas.compiled, mime.cache, …)**.
- **B. the GTK runtime-DATA closure** — the glibc equivalents of what the sibling musl variant (`install-firefox-musl.sh` c1/c3) ships, staged by an explicit list (not the BFS) at the paths the `terminal.rs` launcher env names: **gschemas.compiled** (`org.gtk.Settings.*` — load-bearing, the build `die`s if it can't compile it), gdk-pixbuf **loaders.cache** + modules, and **mime.cache**. Waived (documented, decorative on the glibc X11 display-open path): xkb, Xcursor, icon/GTK themes.

The **fonts gate** (real fontconfig config + a real font, needed for glyph rendering / FCP) is already owned by `install-fonts-real.sh` (`/etc/fonts/fonts.conf`) + `create-data-disk.sh` (`DejaVuSans.ttf`); this script only sanity-warns if absent.

## Closure size note (transparency)

The real GTK3+X11 `.so` closure is **53 sonames / 49 noble packages / ~23 MiB** — above the ~40-package rule-of-thumb, but this is the *intrinsic, expected* size of a real glibc GTK3 stack (the musl image already ships an equivalent closure; the X11 family alone is 18 packages because X ships every extension as its own tiny `.so`). It is **not** a scope explosion, **not** a version wall (2.39 ≤ image glibc), and **not** a musl clobber (isolated build to a new image). Flagging per the stop-and-report discipline; the image/PR are disposable if a different source is preferred.

## Verified (read-only, no boot)

- `e2fsck -fn` clean (RC=0).
- Full `firefox-bin` + `libxul` `DT_NEEDED` closure (**83 sonames**) resolves entirely inside the image, **zero unresolved**.
- Real-vs-stub confirmed by `readelf`/`nm`: libX11 1267K/1235 syms, libxcb 158K/676, libgtk-3 7965K/4574, libgdk-3 998K/761 — all **real** (stubs were 14–80K/few syms). libasound + libdbus-glib-1 confirmed **still stubs** (byte-identical to the generated stub).
- `/etc/fonts/fonts.conf` + `DejaVuSans.ttf` present; libxul `.symtab` present.
- Shared `build/data.img` and `build/data-glibc.img` left **byte-for-byte untouched**; `/tmp/.X11-unix` expectations unchanged (script touches only `lib64` + `lib/x86_64-linux-gnu`).

## Only a boot can verify (not done here, per dispatch)

display-open actually succeeding, GTK init, and fonts rendering on the live in-kernel X server.

## Merge order

**Stacks on #687** (`tools/glibc-data-image`: `--output`/`--build-dir` isolation + `ASTRYXOS_BUILD_DIR`-aware sub-scripts). Base is set to that branch; **merge after #687**. No harness behaviour change — boot the new image with `ASTRYX_GLIBC_DATA_IMG=build/data-glibc-real.img`. The image `build/data-glibc-real.img` is a build artefact, not committed.

Build command (isolated):
```
ASTRYXOS_FIREFOX_VARIANT=glibc ASTRYXOS_BUILD_DIR="$PWD/build-glibc-real" \
  scripts/create-data-disk.sh --output="$PWD/build/data-glibc-real.img" --force
```

References: [GTK3](https://docs.gtk.org/gtk3/), [fontconfig](https://www.freedesktop.org/wiki/Software/fontconfig/), [Ubuntu noble packages](https://packages.ubuntu.com/noble/)


me1iissa (commented):
## Independent pre-merge review — PR #691 (`install-gtk-real-glibc.sh` + create-data-disk hook)

Verdict: **REQUEST-CHANGES** (narrowly scoped — see below). The script is non-fatal, non-clobbering, and the `.so` half is correct; the block is on completeness + one overstated claim, not on safety.

Read-only review at head `21121b8c`, base `tools/glibc-data-image` (#687 head `cca2292d`). No boot.

---

### Dimension-by-dimension

**1. Isolation honesty — CLEAN.**
The apt sandbox is thorough. Every apt state/cache/etc path is redirected into `${APTROOT}` (`install-gtk-real-glibc.sh:214-223`): `Dir::Etc::sourcelist`, `Dir::Etc::sourceparts`, `Dir::Etc::preferencesparts`, `Dir::State::lists`, `Dir::State::status`, `Dir::Cache`, `Dir::Cache::archives`. The empty dpkg status (`:207`) stops apt from reading host install state. Keyring is *read* only (`[signed-by=${UBUNTU_KEYRING}]` in the generated sources.list, `:209-211`) — because trust is pinned per-source, apt never consults or writes the host trusted keyring store, so `Dir::Etc::trusted{,parts}` not being overridden is harmless. `set -euo pipefail` + `pipefail` means a failed `apt-get update`/`download` propagates (not masked by the `| sed`). Temp trees (`APTROOT`/`DEBS_DIR`/`EXTRACT_DIR`) live under `${CACHE_DIR}` (default `~/.cache/astryxos-gtk-glibc`), each `rm -rf`'d at the top of `acquire_noble` — no junk in shared build locations on failure. No host mutation, no sudo. Nit: run as non-root, `apt-get download` may emit a cosmetic "couldn't drop privileges for '_apt'" warning; harmless (download still happens as the invoking user).

**2. No-clobber — CLEAN.**
The script writes only to `DISK_LIB64`/`DISK_GNU` (`:113-114`) and the manifest at `${BUILD_DIR}/disk/.gtk-real-manifest` (`:343`). It never touches an image file, so `data.img`/`data-glibc.img` are structurally out of reach. `BUILD_DIR="${ASTRYXOS_BUILD_DIR:-…/build}"` (`:112`) is the same pattern as the sibling scripts, and `create-data-disk.sh` `export ASTRYXOS_BUILD_DIR` (`create-data-disk.sh:233`) runs *before* the new hook (`:404`), so an isolated glibc build stages into its own tree. The #687 half-set-env hazard does not recur here: this script deals with no image path at all, so there is no split-target to get half-set. Standalone invocation without `ASTRYXOS_BUILD_DIR` defaults to the shared `build/disk` — identical to `install-fonts-real.sh`/`install-dbus-real.sh`, so no new footgun.

**3. Closure correctness — the substance. See REQUEST-CHANGES #2 below.**
The `.so` DT_NEEDED closure is correctly built. Package names are right for noble's time_t64 transition (`libgtk-3-0t64`, `libglib2.0-0t64`, `libatk1.0-0t64`, `libpng16-16t64` at `:145-161`), the BFS (`:292-316`) does a real transitive `readelf -d` walk and hard-errors (`exit 2`) on any unresolved edge. Spot-checked roots (libx11, libxcb, libgtk-3, libgdk-3, gdk-pixbuf, pango+pangoft2, cairo, harfbuzz, glib) are all present in `NOBLE_PKGS`/`DIRECT_SONAMES`.

But: the closure is **`.so`-only**. The GTK *runtime-data* closure is entirely absent from the glibc image — `grep` confirms `install-glibc.sh` and `install-firefox.sh` (the glibc-branch scripts) stage none of `gschemas.compiled` / glib schema source / `gdk-pixbuf .../loaders.cache` / loader modules / `mime.cache` / `/usr/share/mime` / `/usr/share/X11/xkb`, and the new script stages none either. That entire set exists **only** in the musl branch (`install-firefox-musl.sh` sections c1 `loaders.cache` and c3 `gschemas.compiled` + `mime.cache`) — i.e. in exactly the variant this PR claims parity with, where it was added specifically because Alpine's apk triggers never fired (the AstryxOS windowed-GTK bring-up needed c3 to stop a `gtk_init` crash).

Load-bearing judgement for `gdk_display_open → gtk_init → first paint`:
- **`gschemas.compiled` + `/usr/share/glib-2.0/schemas`** — most likely load-bearing. GTK3 `GtkSettings` reads GSettings-backed properties; any code on the init path that calls `g_settings_new()` on a schema absent from the compiled cache is a hard `g_error`/abort (GLib GSettings contract, https://docs.gtk.org/gio/class.Settings.html). GTK3's own `org.gtk.Settings.*` bindings are `schema_source_lookup`-guarded and only *warn*, so this is not certain to abort — but it is the exact class the musl variant hit. Confidence it blocks first-paint: ~0.5.
- **gdk-pixbuf `loaders.cache` + loader modules** — *not* load-bearing for first paint. Modern gdk-pixbuf compiles PNG/JPEG as built-in loaders (https://docs.gtk.org/gdk-pixbuf/), so window/theme icons decode without an external module or cache; the Adwaita theme is a compiled-in GResource, not a disk file. Decorative.
- **`mime.cache` / shared-mime-info** — GtkFileChooser / `g_content_type_guess` only, not display-open/first-paint. Decorative.
- **xkb data (`/usr/share/X11/xkb`)** — server-side; the GDK-X11 client receives the keymap over the wire (XKB extension). Affects keyboard input, not paint. Decorative.
- **Xcursor themes** — graceful fallback to the built-in cursor. Decorative.

And the PR prose ("verified transitively complete at build time") + the header comment (`:47-49`) overstate coverage: the BFS validates `.so` DT_NEEDED edges and is **blind** to dlopen'd loaders, gio modules, and every runtime data file above. That's a factual-accuracy issue independent of the boot outcome.

**4. Soname eviction — CLEAN.**
`install-firefox-stubs.sh` writes its GTK/X11 stubs to `DISK_LIB64` + `DISK_MULTIARCH` only (`install-firefox-stubs.sh:89-90`); it does **not** drop GTK/X11 stubs into `/opt/firefox` (it only *reads* libxul/libmozgtk from there, `:107-108`). The new script's `copy_lib` (`install-gtk-real-glibc.sh:318-333`) overwrites the real file in both dirs and `os.remove`+`os.symlink`-evicts the stub at the soname path in both. The glibc LD_LIBRARY_PATH (`terminal.rs:910`) is `/lib/x86_64-linux-gnu:/usr/lib:…:/opt/firefox:…` — the real-lib dir is **first**, and `/opt/firefox` holds no GTK stub. So the "stub libX11 still resolves first → NULL display with real libs present" failure cannot occur. `gdk_display_open` is confirmed the NULL-returning stub being replaced (`install-firefox-stubs.sh:223-227`).

**5. Failure loudness — CLEAN (+ one confirmed out-of-scope follow-up).**
The DT_NEEDED BFS error path is loud and actionable (`exit 2` + named missing sonames + "Add the owning package(s) to NOBLE_PKGS", `:310-316`). Separate/out-of-scope: `inject-libxul-symbols.sh` (not in this diff) has success exit paths at `:76` and `:139-141` and stderr count-floor warnings at `:344`/`:348` — its "failed" warning surfaced through `create-data-disk.sh:333-334`'s `||` guard can appear spurious while the `.symtab` is in fact present. Confirmed pre-existing; do not fix here — track as its own PR.

**6. Stacking hygiene — CLEAN.**
`git merge-base tools/glibc-data-image HEAD` = `cca2292d` (= #687 head). Diff against that base is exactly the 3 files (`.gitignore` +1, `create-data-disk.sh` +21, `install-gtk-real-glibc.sh` +369; 0 deletions), and only one commit (`21121b8c`) is unique to #691 — #687's commit is not re-included. Base is set correctly; "merge after #687" is accurate. `.gitignore` add `build-glibc*/` is additive.

---

### REQUEST-CHANGES (2 asks; neither can regress/clobber — this is a completeness + accuracy block)

1. **(cheap) Scope the "transitively complete" claim.** In the PR body and the `install-gtk-real-glibc.sh:47-49` header, state that the build-time check verifies the **`.so` DT_NEEDED closure only** and is **blind to dlopen'd gdk-pixbuf loaders, GIO modules, and GTK runtime data files (gschemas.compiled / mime.cache / …)**. As written it implies a completeness the BFS does not deliver.

2. **(substance) Close, or explicitly justify away, the GTK runtime-data gap.** The `.so` overlay gives a real GTK stack, but the glibc image ships **none** of the runtime data the sibling musl variant found necessary (`install-firefox-musl.sh` c1/c3). Either:
   - stage the glibc equivalents — minimally `gschemas.compiled` + `/usr/share/glib-2.0/schemas` (from `libglib2.0-0`/`libgtk-3-common`; compile with `glib-compile-schemas`), plus `mime.cache` + `/usr/share/mime` and any external gdk-pixbuf loader modules + `loaders.cache` if the noble `libgdk-pixbuf-2.0-0` ships them out-of-line; **or**
   - document per-item why each is unnecessary on the glibc display-open path (my read: loaders/mime/xkb/cursor are decorative; **gschemas is the one I would not wave off without a boot**), so the next agent booting `data-glibc-real.img` isn't surprised by a `gtk_init`/GSettings abort the musl variant already solved.

Merging as-is cannot break or clobber anything (worst case the image is still headless-only, same as today with larger libs), so if the coordinator prefers to land with a tracked follow-up for #2, that's a defensible call — but the PR's stated *goal* (windowed display-open for the ~50× discriminator) is not demonstrably reached until #2 is resolved and a boot confirms.

### Nits (non-blocking)
- **Reproducibility**: `NOBLE_PKGS` is suite-pinned, not version/SHA-pinned; `create-data-disk.sh` passes `--force` for the glibc-real build so every build re-`apt-get download`s the current noble candidate (~23 MiB). The `.so`-edge BFS catches a *removed* dependency but not ABI/behaviour drift within a soname. PR acknowledges this; consider a version-lock or checksum manifest for a stable discriminator.
- `stage_closure` `resolve()` (`:272-282`) ranks candidates by **full-path** length, not basename; benign on noble (one file per soname) but would mis-pick if a soname existed in two dirs.

me1iissa:
## Revision addressing the review (head `7a5f128e`)

Thanks — the block was right. The `.so` BFS is structurally blind to the dlopen'd module + runtime-data closure, and the glibc image shipped none of what the sibling musl variant (`install-firefox-musl.sh` c1/c3) already ships. Fixed in `7a5f128e`.

**REQUEST-CHANGES #2 (substance) — GTK runtime-DATA gap.** New `stage_runtime_data()` stages a second, explicit closure (B) after the `.so` overlay, at the exact paths the `terminal.rs` launcher env points GTK/GLib at:

- **gschemas.compiled** (the non-waivable item) + `org.gtk.Settings.*` (libgtk-3-common) + `org.gnome.desktop.*` (gsettings-desktop-schemas) XML at `/usr/share/glib-2.0/schemas/` — the GIO default `XDG_DATA_DIRS` path (launcher sets no `GSETTINGS_SCHEMA_DIR`). Compiled with the host `glib-compile-schemas`; **the build now `die`s if `gschemas.compiled` is not produced**. Verified in-image: 39436 B, 33 schemas incl `org.gtk.Settings.FileChooser`.
- **gdk-pixbuf loaders.cache + modules** at `/usr/lib/gdk-pixbuf-2.0/2.10.0/` (the non-multiarch path `GDK_PIXBUF_MODULE_FILE` names, mirroring musl). You were right that PNG/JPEG are **built-in** on noble 2.42 — so this is decorative for first-paint — but since `terminal.rs` treats an absent cache as an empty loader set that can NULL-deref, I generate a valid non-empty cache by running the noble `gdk-pixbuf-query-loaders` **natively** (noble x86_64 runs on the glibc host; no qemu needed). libtiff loader skipped (libtiff not in the `.so` closure). Non-fatal.
- **mime.cache + /usr/share/mime** (shared-mime-info), host `update-mime-database`. Decorative (content-type/FileChooser only), shipped for parity. Non-fatal.

**Documented waivers** (not staged, decorative on the glibc X11 display-open path, in the script header): **xkb** (GDK-X11 gets the keymap from the server over the XKB wire protocol; libxkbcommon reads the rules tree only on RMLVO keymap-compile, which is the Wayland path), **Xcursor** (built-in fallback), **icon/GTK themes** (Adwaita is a compiled-in GResource in libgtk-3). None block first-paint. This matches your read exactly.

**REQUEST-CHANGES #1 (cheap) — scope the claim.** Done. The header (`install-gtk-real-glibc.sh`, the closure block) and the PR body now state the build-time check covers the **`.so` DT_NEEDED closure only** (closure A) and is **blind to dlopen'd loaders / GIO modules / runtime data files**, which closure B covers via the explicit staging list.

**Nits (non-blocking) — acknowledged:**
- **Suite-pinning vs version/SHA-pinning**: `NOBLE_PKGS` is suite-pinned; a stable discriminator would want a version-lock or checksum manifest. Left as-is for now (the `.so`-edge BFS catches a removed dep; ABI drift within a soname is not caught) — worth a follow-up if the image becomes a fixed reference artefact.
- **`resolve()` ranks by full-path length**, not basename — benign on noble (one file per soname), would mis-pick only if a soname existed in two dirs. Left as-is.

**Verification (read-only, no boot)** on the rebuilt `data-glibc-real.img`: e2fsck clean; full `firefox-bin`+`libxul` DT_NEEDED closure (83 sonames) resolves in-image, zero unresolved; the four data files present at the launcher-expected paths (loaders.cache confirmed referencing the *image* runtime paths, not the build staging path); `build/data.img` + `build/data-glibc.img` byte-for-byte untouched.

**CI**: `build.yml` is gated to `pull_request: branches: [master]`, so it does not run on this PR (base = the #687 branch). It'll run once #687 lands and this re-targets master; the change is bash-only (no Rust), so the cargo/kernel-smoke jobs are unaffected.

Still no boot — the P8 experiment owns the box; display-open / gtk_init / render validation is the next phase.

me1iissa (commented):
## Re-review of the delta `21121b8c..7a5f128e` — Final verdict: APPROVE-WITH-NITS

Both prior REQUEST-CHANGES asks are addressed. Delta is single-file (`install-gtk-real-glibc.sh` +164/-20): new `stage_runtime_data()` (closure B) + prose rescope + sentinel tightening + 4 data packages. The previously-verified `.so` half and the no-clobber guarantees are untouched. Verified read-only, no boot.

### 1. gschemas (B1) — VERIFIED, load-bearing item closed
- **Path correctness confirmed against `kernel/src/gui/terminal.rs`**: the FF env sets **no `GSETTINGS_SCHEMA_DIR`** and **no `XDG_DATA_DIRS`** (grep: zero hits for both). With `XDG_DATA_DIRS` unset, the noble `libgio` doing the lookup falls back to its Debian-compiled default `/usr/local/share:/usr/share`, so `/usr/share/glib-2.0/schemas` — exactly where B1 stages `gschemas.compiled` — is searched. Claim holds.
- Schema set includes `org.gtk.Settings.*` (incl. `org.gtk.Settings.FileChooser`, the schema the musl variant crashed on) via `libgtk-3-common`, plus `org.gnome.desktop.*` via `gsettings-desktop-schemas` (NOBLE_PKGS `:209-214`).
- Compiled with host `glib-compile-schemas` → version-stable GVDB; a 2.43-host writer emits GVDB a 2.39 `libgio` reads (format stable). **Hard `die` if `gschemas.compiled` isn't produced** (verified — the else-branch `die "... would abort on the GUI path (LOAD-BEARING)"`). Good — the one non-waivable item is hard-gated.

### 2. loaders.cache (B2) — VERIFIED sound, correctly decorative
- **Internal paths reference IMAGE paths, not staging paths**: the cache is generated with `GDK_PIXBUF_MODULEDIR="${ld_dst}"` (a staging path) then `sed "s|${ld_dst}|/usr/lib/gdk-pixbuf-2.0/2.10.0/loaders|g"` rewrites those to the image path, gated by `grep -q '^"/usr/lib/gdk-pixbuf-2.0'`. This is the identical, already-proven pattern from `install-firefox-musl.sh` c1. Runtime path matches `terminal.rs:1058` `GDK_PIXBUF_MODULE_FILE=/usr/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache` (non-multiarch). And `GDK_PIXBUF_MODULEDIR` is not set in the FF env, so the baked image paths are authoritative.
- **Host-native execution of the noble query tool is sound (glibc compat direction)**: a noble binary (glibc 2.39) runs on the glibc-2.43 host — forward-compatible per glibc symbol-versioning; deps resolved via `LD_LIBRARY_PATH=${gnu_ext}` against the extracted noble libs. Same direction the shipped image relies on (2.39 ≤ image 2.43). Correctly framed as decorative for first-paint (PNG/JPEG built-in on noble) and non-fatal.

### 3. Waivers — all sound for the FIRST-PAINT bar
- **xkb** — GDK-X11 receives the keymap from the server over the XKB wire protocol; libxkbcommon reads the `/usr/share/X11/xkb` rules tree only when compiling from RMLVO names (the Wayland path), which the X11 backend doesn't do. Affects keyboard input, not paint. Sound.
- **Xcursor** — missing theme falls back to the core/server cursor; doesn't block paint. Sound.
- **Adwaita/themes** — GTK3's default theme is a compiled-in GResource inside libgtk-3, not a disk file; a missing icon theme degrades to the built-in broken-image icon. Sound (matches the established behaviour).

### 4. No regression to the `.so` half or no-clobber — CONFIRMED
`stage_runtime_data()` writes only under `${BUILD_DIR}/disk/usr/share/...` and `${BUILD_DIR}/disk/usr/lib/gdk-pixbuf-2.0/...` (isolated staging tree) — it does not touch `stage_closure`/`copy_lib`, `lib64`/`lib/x86_64-linux-gnu`, or any image file. The 4 added packages only enlarge the extract-tree index; the BFS still ships only sonames reachable from `DIRECT_SONAMES`, so `.so` reachability is unperturbed. Sentinel now also requires `org.gtk.Settings.FileChooser.gschema.xml`, correctly refreshing a stale pre-data extract. No host mutation (the query child runs with a scoped `LD_LIBRARY_PATH`).

### CI situation (confirmed — for the merge plan)
`.github/workflows/build.yml` triggers on `push`/`pull_request` **`branches: [master]` only**. Because #691 is based on the `tools/glibc-data-image` (#687) branch, the PR event did **not** fire build.yml — accurate. This is a **bash-only** change (no Rust/kernel surface), so no compile signal is lost regardless. **Merge plan**: after #687 merges, re-target #691's base to `master`; the re-target fires build.yml at that point. Note there is no shell-lint/shellcheck job, so this script is never CI-exercised either way — human review remains the gate.

### Nits (non-blocking)
- The B1 hard `die` is demoted to a warning by `create-data-disk.sh:404-409`'s `|| echo WARNING`, so a build host lacking `glib-compile-schemas` (pkg `libglib2.0-bin`) would proceed to `mke2fs` an image with real `.so`s but no `gschemas.compiled` → runtime GTK abort. Ensure the build/CI host has `libglib2.0-bin` (the `die` message already says so). Cheap belt-and-braces: have `create-data-disk.sh` preflight-check the tool for the glibc-real path.
- loaders.cache `sed` rewrites only `${ld_dst}`-prefixed paths; the `grep` gate proves image paths are present but not that no staging path leaked. Decorative/non-fatal, and matches the musl pattern — acceptable.
- (carried from r1) `NOBLE_PKGS` remains suite-pinned, not SHA/version-pinned; `--force` re-downloads the current noble candidate each build. Fine for a disposable discriminator image; consider a checksum manifest if the image needs to be stable across time.

**Final verdict: APPROVE-WITH-NITS.** The load-bearing gschemas gap is closed with a hard gate at the correct XDG path; loaders/mime are staged and correctly labelled decorative; the three waivers are sound for first-paint; the `.so` half and no-clobber guarantees are intact. Boot-verification of live display-open + gtk_init remains the only thing this static review can't certify (as the PR states).

PR #692, opened 2026-07-02
sched/vfs: dead-LAPIC-timer immunity — route time-dependent decisions to the TSC-derived tick

me1iissa:
## Summary

The raw `TICK_COUNT` counter is advanced **only** by the periodic LAPIC timer ISR. Under a dead timer — e.g. a KVM vCPU whose LAPIC injection is suppressed (Intel SDM Vol. 3A §11.5.4) — it **freezes** while wall time keeps moving. `get_ticks()` already masks this for its callers by returning `max(TICK_COUNT, tsc_floor)` off the invariant TSC (Intel SDM Vol. 3B §17.17, constant-rate TSC), but a few paths still read the raw counter directly and so make **time-dependent decisions against a frozen clock**.

This PR routes every such raw reader through `get_ticks()`. **Robustness hardening — this is NOT a Firefox-latency fix** (see "Scope / what this does NOT fix").

## Audit — every raw `TICK_COUNT` decision-path reader

| Site | Role | Behaviour under a frozen counter | Severity | Change |
|---|---|---|---|---|
| `sched::current_tick_for_quiesce` | kstack dead-stack quiescence time source (push stamp + all compares) | delta pegs at 0 | LOW | → `get_ticks()` |
| `sched::entry_is_quiesced` tick gate | kstack cache re-issue wall-clock backstop (Gate 2; per-CPU switch-gen is the primary Gate 1) | tick gate + escape valve never elapse → cache never re-issues via tick path | LOW–MED | → shared helper |
| kstack reaper drain throttle (`kstack_drain_due` slot) | once-per-tick quarantine survey gate | admits one drain then wedges → reaper stalls, quarantine fills, falls back to immediate free | LOW–MED | → `get_ticks()` |
| `vfs::ramfs::now_secs` | file mtime/ctime/atime | every timestamp pins at the death instant | LOW (cosmetic) | → `get_ticks()` |
| d582 push-provenance stamp; test-mode diag/wait | diagnostics / test helpers | stale stamps / infinite wait | NONE (diag/test) | → shared helper |

Diagnostic-only raw readers (`mm/w215_diag`, `mm/stack_prov`, `arch/db582`, `mm/w215_crc`, `f3_code_dr_watch`) intentionally keep the raw ISR-published value.

**Explicitly NOT changed — `mm/tlb.rs` TLB-shootdown quarantine `enqueue_tick`.** It stamps from `TICK_COUNT` but its drain gate compares per-CPU `CPU_LAST_TICK` — a per-CPU quiescence *correctness* invariant (has this CPU reloaded CR3 since enqueue?), not a wall-clock backstop. Switching it to `get_ticks()` would break the units. A dead per-CPU timer there is a separate per-CPU-liveness problem (parked #650/#648 territory), out of scope for this global-clock fix.

## Design

Non-ISR contexts only, so the `rdtsc` + divide in `get_ticks()` is affordable; the timer-ISR increment path is unchanged. The kstack push stamp and its comparisons all read through one helper (`current_tick_for_quiesce`) so they can never drift. A guardrail doc on `TICK_COUNT` directs future readers to `get_ticks()` for any timeout / deadline / quiescence window / throttle / rate limiter. This design mirrors the standard separation of an interrupt-driven tick counter from a high-resolution clocksource that reconstructs elapsed ticks by dividing the monotonic delta when interrupts are lost.

## Scope / what this does NOT fix

This does **not** address the ~50× Firefox wiki-load slowdown. Every FF-relevant timed wait already reads the TSC-derived `get_ticks()` and is immune to a frozen `TICK_COUNT`:

- futex timeout — `relative_ns_to_wake_tick` (TSC) + `due_wake_scan` (`get_ticks`)
- poll / epoll / select / nanosleep / socket blocking waits — `get_ticks`
- timerfd, waitlist, TCP/ARP/DHCP/DNS timers — `get_ticks`
- scheduler anti-starvation force accounting + per-CPU `min_deadline` — `get_ticks`-relative

**Live confirmation (SMP=1, KVM, wiki repro, this commit):** `TICK_COUNT` frozen at **120 (~1.2s)** across **50s** of wall time, while the `get_ticks()`-driven reactor cadence advanced to `total=6593` and the wiki-load network-gap population was **unchanged** vs baseline (22–31s `c2s`-ended gaps, 65–96% of span) — i.e. the fix does not collapse the gaps. The 50× residual is the parked upstream event-loop wake-handoff (W101 family), not a native-kernel timed wait.

## Test

Test 688 (`test_688_quiesce_immune_to_frozen_tick`): with `TICK_COUNT` forced to 1, the quiescence source returns the TSC floor (observed 6441, not 1) and the tick gate judges an ~8-tick-old entry eligible. Deterministic on SMP (asserts against an independently-computed TSC floor). Full test-mode suite: 333 PASS, Test 659 (existing sched cluster) still green, no regressions from this change. FF still renders (screendump 4441 distinct colours).

### DIFF
--- a/kernel/src/arch/x86_64/irq.rs
+++ b/kernel/src/arch/x86_64/irq.rs
@@ -74,6 +74,19 @@ const PIC_EOI: u8 = 0x20;
 /// advance N× faster than wall time, which silently breaks every consumer
 /// that schedules deadlines from CLOCK_MONOTONIC (the userspace vDSO fast
 /// path among them).  See `timer_tick` for the TSC-monotone update logic.
+///
+/// # Reading this counter
+///
+/// This is the *raw ISR-published* count.  It advances only while the LAPIC
+/// periodic timer ISR (`timer_tick`) is delivering; if that timer stops
+/// (e.g. a KVM vCPU whose LAPIC injection is suppressed — Intel SDM Vol. 3A
+/// §11.5.4) the value **freezes** while wall time keeps moving.  Therefore any
+/// code that makes a *time-dependent decision* — a timeout, a deadline, a
+/// quiescence window, a throttle, a rate limiter — MUST read [`get_ticks`]
+/// (which returns `max(TICK_COUNT, tsc_floor)` and so keeps advancing off the
+/// invariant TSC, Intel SDM Vol. 3B §17.17) and NOT this field directly.  Read
+/// `TICK_COUNT` directly only for diagnostics that specifically want the
+/// ISR-published value (liveness checks, ISR-vs-TSC skew probes).
 pub static TICK_COUNT: AtomicU64 = AtomicU64::new(0);
 
 /// TSC value at boot — captured by [`set_tsc_calibration`] right after the
--- a/kernel/src/sched/mod.rs
+++ b/kernel/src/sched/mod.rs
@@ -843,7 +843,12 @@ fn reap_dead_threads_sched() {
     // without changing which frames are freed or when.  See
     // `LAST_KSTACK_DRAIN_TICK`.
     {
-        let now = crate::arch::x86_64::irq::TICK_COUNT.load(Ordering::Relaxed);
+        // `get_ticks()`, not raw `TICK_COUNT`: the throttle gate advances
+        // `LAST_KSTACK_DRAIN_TICK` by swapping in `now`; if `now` froze (dead
+        // LAPIC timer) the gate would admit exactly one drain and then wedge
+        // forever, stalling the quarantine reaper.  The TSC-derived tick keeps
+        // it ticking at ~TICK_HZ regardless of LAPIC delivery.
+        let now = crate::arch::x86_64::irq::get_ticks();
         if kstack_drain_due(&LAST_KSTACK_DRAIN_TICK, now) {
             drain_pending_kstack_free();
         }
@@ -1136,7 +1141,7 @@ mod d582_push_prov {
         let s = slot(phys);
         s.pfn.store(pfn, Ordering::Relaxed);
         s.tick
-            .store(crate::arch::x86_64::irq::TICK_COUNT.load(Ordering::Relaxed), Ordering::Relaxed);
+            .store(crate::arch::x86_64::irq::get_ticks(), Ordering::Relaxed);
         s.pusher_tid.store(pusher_tid, Ordering::Relaxed);
     }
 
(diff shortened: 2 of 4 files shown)


me1iissa (commented):
## Independent pre-merge review — VERDICT: APPROVE-WITH-NITS

Reviewed adversarially against the 6 requested dimensions, base `23fd66ae` → head `4c8bf265` (+174/−13). The core change is correct: every *production* time-dependent decision reader now goes through `get_ticks()`, and all producer/consumer stamp pairs are converted together, so there is **no mixed-clock comparison in production**. Two test-scaffolding / doc nits and two optional micro-nits below; none block merge.

### 1. Unit-safety of each switch — PASS
Verified every producer/consumer pair reads the same clock:
- `push_tick` producers `push_dead_stack` (`sched/mod.rs:1417`) + `quarantine_kstack_free` (`:1625`) → `current_tick_for_quiesce()`=`get_ticks()`; consumers `entry_is_quiesced` (`:1300`), drain probe (`:1686-1693`), test-mode pop diag (`:1492`) → same helper. Delta `now − push_tick` is honest.
- Drain throttle `LAST_KSTACK_DRAIN_TICK` (`:1584`): the only value ever swapped in is `get_ticks()` (`:851`); producer==consumer. Init `u64::MAX` still first-drains correctly.
- Cross-check of the *un-switched* scheduler due-wake path: `due_wake_scan` callers already pass `get_ticks()` (`:658`, `:711`) against TSC-derived `wake_tick` deadlines — pre-existing and correct, so the PR's "already immune" scope claim holds and the audit did **not** miss it.

### 2. NOT-switched TLB site — PASS (reasoning correct)
`mm/tlb.rs:790` `enqueue_tick` stays raw `TICK_COUNT`; the drain compares per-CPU `CPU_LAST_TICK` (`:888-892`), which is fed by `timer_tick`'s post-CAS `tick` — the raw ISR-published value (`arch/x86_64/irq.rs:1006`→`:1117` `on_cpu_tick(tick)`). Both sides are the same raw-counter unit; switching only `enqueue_tick` to `get_ticks()` would desync them (get_ticks would race a frozen per-CPU stamp). `git diff --stat` confirms the diff does **not** touch `tlb.rs`, so nothing feeding that gate changed.

### 3. ISR-context tolerance — PASS
All converted sites run in thread/scheduler context, none in the timer ISR: the reaper runs at the top of `schedule()` (IF=0 but not ISR; the timer ISR never reschedules kernel contexts), ramfs `now_secs` in syscall context, `d582` under a diag feature. `get_ticks()` takes **no locks** (three atomic loads + `rdtsc` + one 64-bit divide), so it is safe even in the IF=0 reaper context. The old raw load's cheapness was never a correctness requirement here.

### 4. Monotonicity edges — PASS
`get_ticks()`=`max(published, tsc_floor)` is monotone non-decreasing; both operands are monotone, so a later stamp is always ≥ an earlier one → `now ≥ push_tick` always (no underflow; RHS also uses `saturating_add`). No pre/post-conversion mixing within a boot (whole binary compiled consistently). No wrap risk (64-bit invariant TSC). Cross-CPU TSC skew can only make `now` transiently *smaller* → withholds longer (a leak, never an under-wait/UAF), and Gate 1 (`CPU_SWITCH_GEN`) remains the primary correctness signal.

### 5. Test 688 quality — PASS with a NIT
Genuine regression guard, not a tautology: forces `TICK_COUNT=1`, asserts the quiescence source + gate report ≥ an **independently-computed** `tsc_floor` (would fail against the pre-fix raw read). Restores via monotone `fetch_max(saved)` on every post-store exit path (incl. the failure returns). Skips cleanly pre-calibration / early boot.
- NIT: discrimination is strongest on **SMP=1**. On SMP=2 a sibling `timer_tick` can monotone-CAS `TICK_COUNT` back to `tsc_floor` inside the store→read window, so a *buggy* raw read could be masked and still pass (it never false-*fails*). That's acceptable — the whole-machine-frozen regime the fix targets is itself a single-core phenomenon — but the "deterministic on SMP" note in the commit is about no-false-fail, not discrimination; worth stating that explicitly.

### 6. Scope honesty — PASS
Diff == audit table exactly: `irq.rs` is doc-only, six `sched/mod.rs` conversions, `ramfs::now_secs`, Test 688. A repo-wide sweep of remaining raw `TICK_COUNT.load` readers matches the PR's stated exclusions precisely — `mm/w215_diag`, `mm/stack_prov`, `arch/db582`, `mm/w215_crc`, `subsys/linux/f3_code_dr_watch` (all diag-gated), the `irq.rs` clock internals + `timer_tick` writer, and `mm/tlb.rs` — nothing else. No speculative extras; the 50× REFUTED framing is consistent with prior independent evidence.

---

### Nits (non-blocking)

- **[NIT — top, test-only] `test_294` Case D mixes clocks the way this PR guards against.** `test_runner.rs:47790` builds `recent = TICK_COUNT.load() − 3` (raw) and asserts the entry is **WITHHELD**, but `entry_is_quiesced` now reads `get_ticks()`. Under healthy delivery `get_ticks() ≈ raw` (<1 tick) so it passes; but under the *exact* dead-timer burst this PR exists to handle, `get_ticks()` can lead raw by ≥13 ticks, flipping `escape_ok` true → eligible → Case D spuriously FAILS. (`test_659` block E at `:47938` is eligibility-only, so get_ticks-ahead only helps → safe.) Fix is trivial and the PR already added the helper: source `now` in `test_294` from `test_current_tick_for_quiesce()`/`get_ticks()` so its push_tick is in the gate's units. Low-probability + test-only ⇒ NIT, not REQUEST-CHANGES.
- **[NIT — doc] `sched/mod.rs:617`** `due_wake_scan` doc still says "`now` is the current `TICK_COUNT`" though both callers pass `get_ticks()`. Pre-existing and outside the touched lines, but since this PR is precisely about the raw-vs-derived distinction, tightening it would be consistent.
- **[NIT — optional perf]** the drain throttle now pays `get_ticks()` (rdtsc + divide + 3 atomic loads) at the top of **every** `schedule()`, even when `PENDING_KSTACK_FREE` is empty (the common case — emergency-tier frees are rare); previously one atomic load. A cheap relaxed emptiness check before the `get_ticks()`+swap would keep the hot path at a single load. Micro-opt, not required.

Public-spec citations only (Intel SDM Vol. 3A §11.5.4 / §6.14, Vol. 3B §17.17). No merge performed.

PR #695, opened 2026-07-02
tooling: route scrings-dump through the guest-RAM log ring + fix the scrings cr= parser

me1iissa:
## What

Two W101 op-trace **capture-tooling** defects, both surfaced during the Phase A live capture. No change to recorded data, kernel behaviour, or the exit-time crash dump — this is entirely in the on-demand frozen-ring **dump path** and the harness parser.

### 1. `scrings-dump` starved the guest (`kernel/src/kdb.rs`, `kernel/src/syscall/ring.rs`)

The frozen ring can hold ~10^5 entries. Emitting them over COM1 pays roughly one VM-exit per byte on the 16550 THR (Intel SDM Vol. 3C §25.1.3), so a deep dump monopolised the vCPU for ~28 min and wedged kdb *mid-capture*.

`scrings-dump` now routes each composed line through the near-zero-overhead guest-RAM log ring (`drivers::log_ring`): a single reservation + `memcpy` per line — no lock, no `outb`, no VM-exit. A 10^5-entry dump completes in well under a second without perturbing a live capture. Retrieve the lines out of band with `qemu-harness.py log-ring flush` (re-emits for the `scrings` parser) or `log-ring drain` (JSON).

- Bounded to `log_ring::CAP` (4 MiB): a dump larger than that keeps only the **newest** lines — the entries nearest the freeze, which is exactly what the chain reconstruction walks backwards from.
- `com1` fallback retained: `scrings-dump com1` forces the complete-but-slow COM1 path for when the vCPU is free.
- The exit-time crash dump keeps COM1 unconditionally so it survives even if the log ring is never drained.

### 2. The harness `scrings` parser matched **zero** entries (`scripts/qemu-harness.py`)

The kernel dumper has always emitted `cr=<caller_rip>` between `rip=` and `a1=`, but the `_SC_RING_LINE` regex omitted it, so every optrace frozen-ring dump parsed as empty. The regex now captures `cr` as an optional named group (slotted between `rip=` and `a1=`) and switches to named groups so it stays back-compatible with any cr-less line.

### Tests (`scripts/qemu-harness-smoke.py`)

New host-only `run_scrings_parse_check` (< 1 s, no QEMU): parses a synthetic dump with the real `cr=`/ns/tid/wake fields plus a legacy cr-less line, and asserts the `scrings-dump` argv shaping (default log-ring route, `com1` fallback, `pid`+`com1` in either order).

## Validation

- **Parser cross-check on real capture data**: reconstructed the exact kernel serial-line format from the archived Phase A dumps and parsed with both this parser and the reference parser. Byte-identical results: **271 entries (pid5)**, **131072 entries (pid1)**, field-for-field agreement on `cr`/`ret`/`ns`/`tid`/`wake`. (Previously: 0 entries.)
- Host-only smoke suite passes (`--staleness-only`), including the new check.
- Kernel builds with the capture feature set (`firefox-test-core,kdb,syscall-trace`) in 24.8 s.

## Compatibility

All fields additive. `dump_for_pid` / `dump_all_tracked` gained a `route` argument and a returned count, but both are kdb-internal — **no harness-facing breaking change**.

### DIFF
--- a/kernel/src/kdb.rs
+++ b/kernel/src/kdb.rs
@@ -1468,21 +1468,47 @@ fn op_optrace_depth(req: &str, out: &mut String) {
     }
 }
 
-/// `scrings-dump` — non-destructively emit the frozen ring(s) to serial so
-/// the harness `scrings` parser can ingest them.  With `pid`, just that pid;
-/// without, every tracked pid.
+/// `scrings-dump` — non-destructively emit the frozen ring(s) so the harness
+/// `scrings` parser can ingest them.  With `pid`, just that pid; without,
+/// every tracked pid.
+///
+/// Routes through the guest-RAM log ring by default (a single reservation +
+/// memcpy per line — no per-byte UART VM-exits), so dumping a deep ring does
+/// not monopolise the vCPU or wedge kdb during a live capture.  Retrieve the
+/// lines out of band with `qemu-harness.py log-ring flush` (re-emits to the
+/// serial log for `scrings`) or `log-ring drain` (JSON).  The log ring is
+/// bounded (`log_ring::CAP`, 4 MiB ≈ 20K lines): a larger dump keeps only the
+/// newest lines — the entries nearest the freeze, which is what the chain
+/// reconstruction needs.  Pass `com1` to force the complete-but-slow COM1
+/// path instead (use only when the vCPU is free).
 #[cfg(any(feature = "firefox-test-core", feature = "test-mode"))]
 fn op_scrings_dump(req: &str, out: &mut String) {
     use core::fmt::Write;
+    let com1 = extract_field(req, "com1")
+        .map(|s| matches!(s.as_str(), "1" | "true" | "on" | "yes"))
+        .unwrap_or(false);
+    let route = if com1 {
+        crate::syscall::ring::DumpRoute::Com1
+    } else {
+        crate::syscall::ring::DumpRoute::LogRing
+    };
+    let route_name = if com1 { "com1" } else { "log-ring" };
     match extract_field(req, "pid").and_then(|s| parse_u64(&s)) {
         Some(pid) if pid != 0 => {
-            crate::syscall::ring::dump_for_pid(pid);
-            let _ = write!(out, r#"{{"dumped":true,"pid":{}}}"#, pid);
+            let entries = crate::syscall::ring::dump_for_pid(pid, route);
+            let _ = write!(
+                out,
+                r#"{{"dumped":true,"pid":{},"entries":{},"route":"{}"}}"#,
+                pid, entries, route_name
+            );
         }
         _ => {
-            crate::syscall::ring::dump_all_tracked();
-            let n = crate::syscall::ring::tracked_rings().len();
-            let _ = write!(out, r#"{{"dumped":true,"pids":{}}}"#, n);
+            let pids = crate::syscall::ring::dump_all_tracked(route);
+            let _ = write!(
+                out,
+                r#"{{"dumped":true,"pids":{},"route":"{}"}}"#,
+                pids, route_name
+            );
         }
     }
 }
--- a/kernel/src/syscall/ring.rs
+++ b/kernel/src/syscall/ring.rs
@@ -618,6 +618,49 @@ fn nr_name(nr: u64) -> &'static str {
     }
 }
 
+/// Where a ring dump's `[SC-RING-*]` lines are routed.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum DumpRoute {
+    /// Classic COM1 `serial_println!` — every line pays ~one VM-exit per byte
+    /// on the 16550 THR (Intel SDM Vol. 3C §25.1.3).  Complete, but slow: a
+    /// deep frozen ring can monopolise the vCPU for minutes and wedge kdb.
+    /// Used by the crash-path exit dump (always reaches the log) and as the
+    /// `scrings-dump --com1` fallback.
+    Com1,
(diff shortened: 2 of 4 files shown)


me1iissa (commented):
## Independent review — APPROVE (0 blockers)

Reviewed the diff as committed at `a36d4a90` (not the description). Static analysis + host-side validation + reference cross-check on the archived Phase A capture. No QEMU boot (not required for this change).

### What I validated

- **Parser regression is real and the fix is correct.** The pre-PR `_SC_RING_LINE` regex does not match a line carrying `cr=<caller_rip>` between `rip=` and `a1=` — reproduced: it yields **0** matches on a real captured line. The new named-group regex matches both `cr=` and legacy cr-less lines. No catastrophic backtracking: parsing an 81 MB / 1M-line noisy buffer completes in ~0.58 s (the literal `[SC-RING] i=` prefix fails non-matching lines fast; the optional `(?: cr=…)?` group is disambiguated by the surrounding ` a1=` literal).
- **Field-for-field cross-check on real data.** Reconstructed the exact kernel `emit_ring()` serial format from the archived 271-entry pid5 capture and parsed it through the new harness parser: **271 entries, all 15 fields (i/tick/name/rip/cr/a1..a6/ret/ns/tid/wake) byte-identical** to the independent reference parser. Reproduces the author's claim.
- **Host smoke check passes** (10/10 assertions), and it is wired into the `--staleness-only` CI tier (added before the early-exit), so it runs in the cheap host-only cycle. The synthetic block exercises `cr=`/ns/tid/wake plus a legacy cr-less line and the `scrings-dump` argv shaping (default/log-ring, `com1`, `pid`+`com1` in either order).
- **Log-ring route keeps the newest lines and reports drops.** The ring is a monotonic-cursor byte ring (mask wrap), so a dump larger than `CAP` (4 MiB) overwrites the oldest bytes and retains the newest — the entries nearest the freeze, which is what the chain reconstruction walks back from. Truncation is **not silent**: both `log-ring drain` (JSON `dropped_bytes`/`oversize_drops`/`resident_bytes`) and `log-ring flush` (`[LOG-RING-FLUSH] begin … dropped_bytes=…`) report it at drain time.
- **No lock inversion / no hot-path perturbation.** `log_ring::record` is lock-free (atomic reservation + raw writes), so holding `RINGS.lock()` across the `LogRing` emit introduces no new lock ordering. Frozen rings are immutable during a dump: `begin()` returns `None` once `is_frozen_at()` trips, and it otherwise contends the same `RINGS.lock()` the dump holds, so no torn entries. The new code runs only on the kdb-triggered dump / exit-dump paths — nothing new lands in the syscall hot path.
- **Additivity + callers.** kdb JSON gains `entries`/`route` (additive; `dumped`/`pid`/`pids` preserved); the `dump_for_pid`/`dump_all_tracked` signature change (added `route`, returned count) is kdb-internal — the only callers in the tree are `op_scrings_dump` and `dump_all_tracked → dump_for_pid`. Feature-gating is clean: the module and every new symbol compile under `firefox-test-core`/`test-mode`; `syscall-trace` is a separate opt-in this change does not depend on. Exit-time dump correctly stays on COM1 unconditionally.

### Non-blocking findings (recommend a fast follow / operator briefing)

1. **MAJOR (discoverability): the default-route change silently breaks the historical `scrings-dump` → `scrings` sequence.** `cmd_scrings` reads only the serial log; with the default now `LogRing`, an on-demand `scrings-dump` no longer lands in the serial log until `log-ring flush` is run. The `scrings` command help/docstring is not updated to say so. Exit dumps are unaffected (still COM1), so this only bites the deliberate frozen-ring flow — but that is exactly the recapture path. Suggest: document the required `log-ring flush` in the `scrings-dump` kdb help + `cmd_scrings` docstring, or have `scrings` optionally trigger a flush.
2. **MAJOR (multi-pid eviction for the "dump ALL children" need): the bare all-pids `scrings-dump` can evict earlier pids before they are drained.** `dump_all_tracked` writes every tracked pid's ring into the single 4 MiB log ring in sequence with no flush between; when the aggregate exceeds `CAP` (pid1 alone at 131072 entries ≈ far more than 4 MiB of text), the earlier-dumped pids' lines are overwritten before a drain can recover them, leaving only the tail of the last pid(s). The drop is reported at flush, but the `scrings-dump` JSON `entries` count reflects lines *written*, not lines *surviving in the ring*, which can mislead. Correct usage for the recapture is per-pid `scrings-dump <pid>` + `log-ring flush` between each. Suggest: document that cadence (the stated need is "dump ALL children's rings"), and/or auto-flush per pid inside `dump_all_tracked`'s harness wrapper, and/or surface an overflow signal in the JSON.

### Nits

- `record_line_to_ring` truncates a line longer than its 1 KiB stack buffer *silently* (the `RingBuf` loop just `break`s), and because that clamps the payload to `MAX_RECORD` the `record()` `OVERSIZE_DROPS` counter never fires for these lines. Safe today — the longest emitted line is the `[SC-RING-BYTES]` hex line at ~552 B (READ_BYTES=256 → 512 hex chars + prefix), well under 1 KiB — but it is a latent silent truncation if `READ_BYTES` is ever raised. Worth a one-line note on the coupling.
- `scrings-dump com1` still holds `RINGS.lock()` across the slow COM1 emit (the same starvation this PR removes from the default path), gated behind the explicit opt-in and documented "use only when the vCPU is free." Accepted tradeoff, called out for the record.

### Confidence / scope

High confidence (~0.9) on the correctness verdict. Not checked: a live boot (out of scope for this tooling change); re-parsing the 44 MB pid1 dump (identical format to pid5, covered by the pid5 cross-check + smoke + backtracking test); a cargo build (feature-gating verified by cfg inspection; author reports a 24.8 s build of the capture feature set).

PR #696, opened 2026-07-02
tooling: fd-map TCP 4-tuples + paged kdb scrings dump (retire the log-ring route)

me1iissa:
Probe-cycle hardening from the 2026-07-02 W101 load-phase recapture. Addresses task #8: three defects the recapture found live (host-only smoke missed them — no concurrent firehose) plus the fd-map peer-host gap the verdict flagged.

## ⚠️ Breaking change (harness op semantics)

`scrings-dump` changes again, one PR after #695 introduced its log-ring route. It **no longer routes through the guest-RAM log ring**; its default is now the COM1 serial emit (parse with `scrings`). The `com1` positional and the log-ring `route` field are removed. The fast, complete, firehose-immune path is the new **`scrings-pull`** subcommand. Downstream dispatches calling `scrings-dump [pid] com1` should drop `com1` (COM1 is the only `scrings-dump` route now) and prefer `scrings-pull`. Harness co-changes are in this PR so nothing external breaks silently.

## 1. fd-map TCP 4-tuples (`net/socket.rs`, `kdb.rs`)

The kdb `fd-map` op resolved only the AF_UNIX peer table, so an AF_INET/INET6 socket showed `peer_socket_id:none` with no address — the gap-ending TCP send's peer HOST was **unbindable** in the verdict. `op_fd_map` now branches on the fd's `UNIX_SOCKET_FLAG` (bit 23 — the same predicate `is_unix_socket_fd` uses; the UNIX/INET tables have independent id spaces, so a table-membership guess could mis-key) and for INET sockets emits `af`, `l4` (tcp/udp), `local`/`remote` as `a.b.c.d:port`, `local_port`, `remote_port`, `bound`, `connected`, via a new read-only `socket_inet_snapshot` (getsockname(2)/getpeername(2)-shaped).

## 2. Paged kdb scrings dump — the log-ring route is retired (`ring.rs`, `kdb.rs`, harness)

PR #695 routed `scrings-dump` through the shared 4 MiB guest-RAM log ring. The recapture **proved that unusable under a live `syscall-trace` session**: the ~220K rec/s `[SC]` firehose shares that one ring and laps it in ~0.18 s, evicting dump lines before drain; and a >4 MiB dump (a 32768-entry pid1 ring ≈ 5.5 MB) can't fit the ring at all. A dedicated ring fixes only eviction; firehose-suppression only eviction; a bigger drain cap only the response size — **none makes the route both complete and firehose-immune.**

Replaced with a **paged direct-kdb dump**: the new `scrings-page` op reads the frozen ring in bounded windows straight into the kdb response (each < `MAX_RESP_BYTES`); the harness `scrings-pull` subcommand loops `from = next` until `eof` — the same offset-cursor protocol the base64 PNG read uses. Firehose-immune (no shared ring), capacity-independent (paging), fast (kdb-TCP, not per-byte COM1 PIO — seconds vs the ~8 min a COM1 pid1 dump took). `scrings-dump` stays as the COM1 serial fallback for when kdb is wedged. The `[SC-RING]` line format is now a single `format_entry` shared by both paths so they can't drift; `DumpRoute`/`record_line_to_ring` removed.

**Operational envelope (known dependency):** a full pid1 dump is ~180 kdb round-trips, each subject to the same storm-starvation that could wedge kdb post-load. This is acceptable — dumps are taken at freeze/idle windows where kdb is responsive, and the COM1 `scrings-dump` fallback covers a wedged kdb — but a partial pull must be **loud, not silent**: `scrings-pull` marks `all_complete=false`, lists each incomplete pid's `have`/`total`/`resume_from`/`error`, prints a stderr warning, and exits non-zero. It never returns a truncated set as if whole.

## 3. Harness `_kdb_recv` max_bytes (`qemu-harness.py`)

Incorporates capture-driver's banked fix: `_kdb_recv` hard-broke at 128 KiB with no trailing newline, truncating large responses into "malformed response". Now takes `max_bytes` (default 128 KiB; the paged pull and log-ring drain raise it).

## Validation

- **Kernel builds** under `firefox-test-core,kdb,syscall-trace`; base commit went fully green in CI (all 10 jobs incl. kernel-smoke QEMU boot, all feature combos, cargo clippy).
- **Paged reassembly cross-checked against the REAL archived rings**: simulating the `scrings-page` transport and driving the actual `_scrings_pull_one` loop reassembles **271 (pid5)** and **131072 (pid1)** entries, field-for-field, across 3 and **943 pages** respectively — no boundary loss, `complete=True`.
- **Host-only smoke** (`run_scrings_parse_check`) covers: the `cr=` parser guard, paging to EOF (5000 entries / 29 pages), a short/final single-page ring, and **resume-after-partial** (injected mid-pull kdb error → asserts INCOMPLETE + correct `resume_from`). The multi-page reassembly is the firehose-immunity guard that would have caught the #695 defect.

## Scope

All JSON fields additive. `dump_for_pid`/`dump_all_tracked` drop their `route` param (kdb-internal). **Deferred** (off the critical path after this rework): the `log-ring drain` 32 KiB firehose-response cap — scrings no longer uses the log ring, so that cap now only bounds raw-firehose draining; flagged as a small follow-up rather than bundled here.

### DIFF
--- a/kernel/src/kdb.rs
+++ b/kernel/src/kdb.rs
@@ -439,6 +439,8 @@ pub fn dispatch(req: &str, out: &mut String) {
         "optrace-depth"    => op_optrace_depth(req, out),
         #[cfg(any(feature = "firefox-test-core", feature = "test-mode"))]
         "scrings-dump"     => op_scrings_dump(req, out),
+        #[cfg(any(feature = "firefox-test-core", feature = "test-mode"))]
+        "scrings-page"     => op_scrings_page(req, out),
         "futex-ghost-hist" => op_futex_ghost_hist(req, out),
         // cond-autopsy: one-shot musl pthread_cond/mutex wake-target-vs-
         // wait-addr report.  Composes the live struct dump + parked waiters +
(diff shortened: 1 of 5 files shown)


me1iissa (commented):
## Independent review — APPROVE (0 blockers)

Reviewed the diff as committed at `edc610a6` (stacked cleanly on master `6c485797` = #695). Static analysis + host-side execution + a reassembly cross-check on the real archived capture.

### What I validated

- **fd-map INET 4-tuple (net/socket.rs, kdb.rs).** `socket_inet_snapshot` scopes the `SOCKETS` lock to the scalar copy and releases it before `tcp::lookup_local_ip`, so `SOCKETS` and the TCB lock are never held simultaneously — no AB-BA against the net fast path (mirrors `socket_local_addr`). The UNIX-vs-INET hinge is sound: INET fds are flagged `0x4000_0000 | (sock_type & 0x03)` (bit 23 clear) and AF_UNIX fds `0x4000_0000 | UNIX_SOCKET_FLAG` (bit 23 = `0x0080_0000` set); the type bits (0..2) cannot collide with bit 23, so `snap.flags & UNIX_SOCKET_FLAG` reliably discriminates, and both mis-flag failure modes are graceful (`af=unknown` / peer resolution skipped, never a mislabel).
- **Paged dump (ring.rs, kdb.rs).** `format_entry` is now the single source of truth; its line format is byte-identical to the prior `emit_ring`, so the existing `scrings` parser handles both transports unchanged. `page_for_pid` guarantees forward progress (emits ≥1 entry before the byte-budget break, so `from` always advances; pages ≤ total), uses enumerate-before-skip for absolute continuous `i=`, and — critically — checks the budget *between* entries, so every page's `lines` block is whole-lines and reassembly is seamless. `op_scrings_page` budgets the raw block at 24 KiB; with the hex/digit-dominated line content, `j_str` escaping grows the block by well under the ~7 KiB headroom to the 32 KiB `MAX_RESP_BYTES` cap. The `next`/`eof` cursor terminates on a missing pid (total=0 → eof) and on over-paging.
- **Harness.** `_kdb_recv max_bytes` is additive (128 KiB default preserved for existing callers; drain raises to 64 MiB, paged pull to 1 MiB/page) — this is the fix for the recv-truncation defect the recapture found. `_scrings_pull_one` terminates on `eof`/`emitted==0`/the 200K-page ceiling, synthesises the BEGIN/END frame host-side, and the parser counts actual `[SC-RING]` lines (robust even if the synthesised `entries=` disagreed with reality).
- **Tests.** The host smoke suite passes 12/12, including the new 5000-entry / 29-page reassembly guard, which drives the *real* `_scrings_pull_one` loop through a `_fake_kdb` that mirrors the kernel byte-budget contract and asserts completeness + monotonic `i=` + boundary fields. I reproduced the smoke run myself.
- **Reassembly cross-check on real data.** I simulated `page_for_pid` (24 KiB budget, exact line format) over the archived 271-entry pid5 ring, drove `_scrings_pull_one`, and the reassembled dump is **field-exact vs the reference parser (0 mismatches, `i=` continuity intact), 3 pages**. My pid1 page-count estimate (~1000 at 187 B/line avg) matches the author's reported 943.
- **Cleanup / gating.** `DumpRoute` and `record_line_to_ring` are fully retired (no dangling references anywhere). `op_scrings_page` is gated identically to `op_scrings_dump` (`firefox-test-core`/`test-mode`); `op_fd_map` + `socket_inet_snapshot` + `SocketType` are always compiled, so the fd-map change builds in any kdb config.

### Nits (non-blocking)

1. **24 KiB→32 KiB headroom is content-dependent.** The budget is enforced on the *raw* block; `j_str` escaping happens after. Real content (hex is escape-free; paths are normal filesystem strings with 2 quotes) grows by << the 7 KiB headroom, so this is safe in practice. But a pathological escape-heavy raw block is not *hard*-bounded — if one ever exceeded 32 KiB it would be silently truncated at `MAX_RESP_BYTES` (line 305) into malformed JSON. Belt-and-suspenders: after `j_str`, if `out.len() > MAX_RESP_BYTES`, drop the last entry and report a smaller `emitted`/`next` so the host re-pages it. Low priority.
2. **`page_for_pid` holds `RINGS.lock()` across up to `count` heap-allocating `format_entry` calls.** Pre-existing pattern (old `emit_ring` under `dump_for_pid` did the same); bounded by the 24 KiB budget (~130–180 entries), no contention on the smp1 capture, brief on smp2. Acceptable for a diagnostic; noted, not a change request.
3. **`log-ring drain` 32 KiB response cap is deferred** (PR priority #4). Correct call — off the critical path now that scrings no longer uses the log ring.

### Confidence / scope

High confidence (~0.9). Not checked: a live boot (out of scope for this tooling change); a cargo build (gating verified by cfg inspection; author reports it builds under `firefox-test-core,kdb,syscall-trace`); re-parsing the 44 MB pid1 dump (identical line format to pid5, covered by the pid5 cross-check + smoke guard).

me1iissa (commented):
## Re-review of the loud-partial commit (`9f1cb9e9`) — verdict holds: APPROVE

Second commit is harness+smoke only (0 kernel files; the approved kernel side `edc610a6` is untouched). It resolves the LEAD's two caveats cleanly.

**Loud-partial logic — correct.** `_scrings_pull_one` now returns a status dict; `complete` requires `eof` AND `have == total`. Every early exit `break`s with `error` set; the `while/else` correctly attributes only the `max_pages` ceiling; the `complete and have != total` guard surfaces a boundary bug instead of hiding it (and, as a bonus, it also catches a hypothetical malformed-page parse-drop from the kernel-side escape-headroom nit — that would now read as INCOMPLETE, not a silent truncation). Partial data is still returned in `dumps`, but never masquerades as whole: `all_complete=false` + a `partial_pids` list (`have`/`total`/`resume_from`/`error`) + a stderr warning + `sys.exit(2)`. A per-pid transport exception is caught, warned, and the loop continues so one pid's failure doesn't drop the others.

**Verified myself:** both `_scrings_pull_one` callers updated to the dict; smoke passes **15/15**, including the three new cases — EOF paging (5000/5000, `complete=True`, 29 pages, monotonic), short single-page ring (`complete`, 1 page), and resume-after-partial (injected kdb error at page 2 → `complete=False`, error named, **`resume_from == have == 353`**, confirming the contiguous-from-0 cursor claim). The breaking change (`scrings-dump` COM1 default, `com1` positional removed) is now called out in both the commit message and the PR body.

**One informational nit:** `resume_from` is reported but there is no `--from` flag on `scrings-pull` to auto-resume — resume is a manual re-invocation (or a future flag). Fine for a diagnostic; noted only.

No blockers. Recommend merge once CI is green.

me1iissa (commented):
## Final-tip confirm (`6ba91d3c`, nit-1 hard-bound) — verdict holds: APPROVE

Confirmed `6ba91d3c` is the real `origin/tooling/w101-probe-hardening` tip (parent `9f1cb9e9`, cleanly stacked on master `6c485797`) — `gh pr view` was showing the cached `9f1cb9e9` head. Smoke passes **15/15** on this tip (now 46 pages at the 15 KiB budget). The `j_str` ≤2× reasoning is right: the raw block is printable ASCII + `\n`, and `\n`/`\r`/`\t` are matched before the `<0x20` arm so the 6-byte `\uXXXX` branch is never taken.

**One precision caveat (non-blocking, for the record).** The stated proof ("15 KiB raw → ≤30 KiB") omits that `page_for_pid`'s budget check is *post-append*, so a page can overshoot the budget by up to one max-size entry. With `PATH_BYTES=128` / `READ_BYTES=256` the largest single entry is ~0.7 KiB (real read) to ~0.95 KiB (a `{:?}`-escaped all-control-byte path), so the true raw max is ~16 KiB, not 15 KiB. Under the strict 2× ceiling:
- **Realistic FF paths** (normal filesystem strings, no control bytes): ~16.1 KiB → ~32.30 KiB + envelope — **under 32768 by ~466 B**. Safe.
- **Adversarial 128-control-byte path**: ~16.3 KiB → ~32.80 KiB — **~30 B over** the strict ceiling.

That adversarial case is doubly unreachable: (a) it needs a 128-control-byte open() path, which FF never produces; and (b) it needs the whole preceding 15 KiB to also escape at 2×, but that bulk is `[SC-RING]` header text escaping ~1.03×, so the real escaped size is ~16.5 KiB regardless. And even if it *did* overflow, the new loud-partial guard is the backstop — a `MAX_RESP_BYTES` truncation yields malformed JSON → `_scrings_pull_one` hits its error path → the pull is flagged INCOMPLETE, never a silent corruption.

Net: safe as-is with a ~16 KiB real margin; better than my suggested drop-and-repage. If you want the proof airtight for *any* input (not just real FF paths), drop the budget to 14 KiB (clears the strict ceiling by ~2 KiB even in the adversarial case). Optional. **No blockers — merge as-is.**

Issue #21, opened 2026-04-23
/dev/urandom and /dev/random


me1iissa:
## Background
`/dev/urandom` and `/dev/random` are required by virtually every application that uses cryptography, including NSS, OpenSSL, musl's `arc4random`, and the Go runtime. These device nodes need to be backed by the kernel's CSPRNG.

## What to do
1. Wire `/dev/urandom` reads to the kernel CSPRNG (the RNG added in earlier waves)
2. Wire `/dev/random` to the same CSPRNG (Linux semantics: block only until the pool is seeded, then never block again — the GRND_RANDOM behaviour is legacy)
3. Register the device nodes in devfs at `/dev/urandom` and `/dev/random`
4. Seed the CSPRNG at boot using the RDRAND instruction (with RDSEED fallback) and/or TSC jitter

## Acceptance criteria
- [ ] `dd if=/dev/urandom bs=32 count=1 | xxd` produces 32 bytes of non-zero output
- [ ] Two consecutive reads produce different output
- [ ] `open("/dev/random", O_RDONLY)` does not block indefinitely
- [ ] Headless test

## Files likely touched
- `kernel/src/drivers/random.rs` (new or extend)
- `kernel/src/vfs/devfs.rs`
- `kernel/tests/test_runner.rs`

## Notes / spec references
- `man 4 random`
- Linux `drivers/char/random.c`
- RFC 4086 — Randomness Requirements for Security

Issue #20, opened 2026-04-23
/proc/cpuinfo — real CPU info from CPUID


me1iissa:
## Background
`/proc/cpuinfo` currently returns static placeholder content. Many programs (including Firefox's process model detection) read this file to determine CPU count, model name, feature flags (e.g. `sse4_2`, `avx2`), and cache sizes.

## What to do
1. At boot, execute `CPUID` leaf 0x0, 0x1, 0x7, 0x80000002-4 on each logical CPU and store results
2. Generate a `/proc/cpuinfo` entry per logical CPU with correct fields: `processor`, `vendor_id`, `model name`, `cpu MHz`, `cache size`, `flags`
3. CPU MHz: read from `IA32_APERF`/`IA32_MPERF` MSRs or use the PIT-calibrated value

## Acceptance criteria
- [ ] `grep "model name" /proc/cpuinfo` prints a non-empty string matching the host CPU
- [ ] `grep -c "^processor" /proc/cpuinfo` equals the number of logical CPUs
- [ ] `grep "flags" /proc/cpuinfo` contains `lm` (64-bit) and `fpu`
- [ ] Headless test validates the above

## Files likely touched
- `kernel/src/vfs/procfs.rs`
- `kernel/src/arch/x86_64/cpuid.rs` (new or extend)

## Notes / spec references
- `man 5 proc` — `/proc/cpuinfo`
- Intel SDM Vol. 2A — CPUID instruction
- OSDev CPUID: https://wiki.osdev.org/CPUID

PR #698, opened 2026-07-02
mm/tlb: service incoming shootdowns during the ACK spin — break the cross-CPU shootdown mutual livelock

me1iissa:
## Summary

The SMP=2 cross-CPU TLB-shootdown ACK timeouts (measured ~425/load, symmetric, every one at the full `ACK_BOUND`) are a **kernel mutual livelock**, not the KVM dead-timer artifact the census attributed them to. This PR fixes it with the standard cross-CPU shootdown reentrancy pattern.

### Mechanism
A shootdown issues from inside an **IF-masked** critical section — the `#PF` handler is a 64-bit interrupt gate (`idt.rs`: `IDT[14]` type `0x0E` → IF cleared on entry), and the CoW / make-writable / mprotect / MAP_SHARED write-through / shm-detach paths call `shootdown_*` with interrupts still disabled — then spins up to `ACK_BOUND` (~10 ms) for each target to ack. While that sender spins with IF=0, a fixed-mode `0xF0` IPI a **peer** sends *to* it stays pending in the IRR and its ISR cannot run (Intel SDM Vol. 3A §10.6.1). Under concurrent CoW both cores are frequently IF-masked in their own shootdowns simultaneously → each fails to ack the other → **both spin to the full bound**.

Evidence (probe `12c38116`, split-irqchip, `ffrender-artifacts/apicv-probe-2026-07-02/`): 425 timeouts, symmetric (222 cpu0→cpu1 / 203 cpu1→cpu0), all `iters=10000000`, **no dead-timer marker that run** (decoupled from timer death). GDB caught both vCPUs `IF=0` in `schedule()` at the wedge — the task#13 scheduler wedge feeds the same IF=0-can't-ack mechanism.

## Fix
On each **no-progress** ACK-spin iteration the sender services its own incoming shootdown slot via a new lock-free helper `service_incoming_shootdown_and_drain(cpu)` — the invalidation body factored out of `handle_shootdown_ipi` (invlpg if the CR3 matches, ack the slot, then `drain_pending_force_flush`). This lets the peer's ack complete so it can in turn service ours, unwinding the livelock.

### Composition with #643's FORCE_FLUSH drain (the review question)
The two drains **compose in one call** — not additively. `handle_shootdown_ipi` already ran the incoming-service *then* `drain_pending_force_flush`; the helper is exactly that body. Routing the spin through it gives both. The helper does **NOT** EOI the LAPIC: only the ISR entry point may, since the in-spin caller isn't servicing a delivered interrupt (a spurious `lapic_eoi` there corrupts LAPIC priority).

### Correctness invariants
- **Lock-free / IF=0-safe:** the helper and `drain_pending_force_flush` take no locks and send no IPI (they invlpg + reload CR3 + do atomic slot ops), so calling them from the fault-path spin (which may hold `VMM_LOCK`) is safe — no nested lock-order hazard.
- **#643 soundness (`tlb.rs` `on_cpu_tick` note):** untouched. NK uses the FULL CR3-reload drain, never the clear-only `on_cpu_tick` path; `ACK_BOUND` is unchanged and `post_force_flush` still trails the spin — the invariant's preconditions all hold.
- **Slot disjointness:** the serviced `SHOOTDOWN_SLOTS[self]` is disjoint from the sender's `targets` (which exclude `self_mask`); the `compare_exchange(1→0)` on `pending` makes the speculative per-iteration call idempotent (cheap fall-through when nothing is pending).
- **Ordering:** Intel SDM Vol. 3A §4.10.4.2 (shootdown protocol) — invlpg completes before the ack clear (AcqRel CAS); §4.10.4.1 (MOV-to-CR3) covers the force-flush drain.

## SMP=1-inert (by construction)
On one CPU `targets = snapshot_active_mask(cr3) & !self_mask == 0`, so `shootdown_range_inner` returns before the ACK spin — the new service call never executes. The uniprocessor path is bit-for-bit unchanged.

## Diff
`kernel/src/mm/tlb.rs` (+148/-11, most is the moved invalidation body) and `kernel/src/test_runner.rs` (+66, Test 699).

## Verification
- Compiles clean under `firefox-test-core,kdb` and `kdb`-only (feature-combo cfg paths).
- **Test 699 PASSES** (live, test-mode): unit-exercises the reentrancy primitive — services the incoming shootdown, idempotent on a drained slot, composes the force-flush drain in one call.
- **Caveat:** the full local test-mode suite hits a **pre-existing** `DOUBLE_FAULT` at ~test 243 (before the TLB block) — confirmed on clean master by an A/B baseline (same signature, both KVM and TCG), so it is unrelated to this change. Test 699 was verified by running it ahead of that crash point (reverted before commit). Flagging the pre-existing test-mode crash separately.
- The genuine two-CPU livelock is not deterministically reproducible in the single-threaded runner; its behavioral proof is the **SMP=2 census A/B (425 → ~0)**, tracked as the post-merge verification task.

### DIFF
--- a/kernel/src/mm/tlb.rs
+++ b/kernel/src/mm/tlb.rs
@@ -621,6 +621,20 @@ fn shootdown_range_inner(cr3: u64, va_lo: u64, va_hi: u64) -> bool {
         if remaining == 0 {
             break;
         }
+        // Reentrancy: we are about to keep spinning with interrupts disabled
+        // (a shootdown issues from inside an IF-masked #PF / mm critical
+        // section).  A peer CPU may be concurrently spinning on OUR ack for a
+        // shootdown it sent to us; because IF=0 keeps the 0xF0 IPI pending in the
+        // IRR (Intel SDM Vol. 3A §10.6.1) the ISR cannot run, so that incoming
+        // request would go unserviced until we return and re-enable interrupts.
+        // If the peer is likewise IF-masked in ITS own shootdown, both sides
+        // spin to the full `ACK_BOUND` — a cross-CPU shootdown mutual livelock.
+        // Service our own incoming slot here so the peer's ack completes and it
+        // can in turn service ours.  `self_cpu` is disjoint from `targets`
+        // (targets excludes self_mask), so this touches a different slot than the
+        // poll above; the routine is lock-free and EOI-free (safe under IF=0),
+        // and finds `pending == 0` cheaply on the common no-incoming iteration.
+        service_incoming_shootdown_and_drain(self_cpu);
         core::hint::spin_loop();
         iters += 1;
     }
@@ -984,14 +998,27 @@ fn drain_pending_force_flush(cpu: usize) {
     }
 }
 
-/// IPI handler.  Invoked from [`crate::arch::x86_64::idt`] when the LAPIC
-/// delivers a [`TLB_SHOOTDOWN_VECTOR`] interrupt to this CPU.
+/// Service an incoming cross-CPU shootdown targeting `cpu` (if its slot is
+/// pending) and retire any deferred force-flush posted to it.
 ///
-/// Reads the per-CPU shootdown slot, performs the invalidation if the
-/// target CR3 matches the running one, and clears `pending`.  Always
-/// EOIs the LAPIC at the end.
-pub extern "C" fn handle_shootdown_ipi() {
-    let cpu = apic::cpu_index();
+/// This is the invalidation body shared by two callers: the [`TLB_SHOOTDOWN_VECTOR`]
+/// ISR ([`handle_shootdown_ipi`]) and the ACK-spin in [`shootdown_range_inner`].
+/// It performs ONLY local work — an `invlpg` of the published range if the
+/// running CR3 matches, the slot ack (compare-exchange `pending` 1→0), and
+/// [`drain_pending_force_flush`] — so it takes NO locks and sends NO IPI and is
+/// therefore safe to call with interrupts disabled, INCLUDING from inside
+/// another shootdown's own IF-masked ACK spin (the reentrancy the spin relies on
+/// to break a cross-CPU shootdown mutual livelock).  Per Intel SDM Vol. 3A
+/// §4.10.4.2 (TLB-shootdown protocol) the `invlpg` must complete before the ack
+/// clear (the AcqRel compare-exchange provides that order); §4.10.4.1 (MOV-to-CR3
+/// invalidation) covers the force-flush drain.
+///
+/// It does NOT EOI the LAPIC.  Only the ISR entry point may EOI: an in-spin
+/// caller is not servicing a delivered interrupt (IF=0 kept the 0xF0 IPI pending
+/// in the IRR, Intel SDM Vol. 3A §10.6.1), so a spurious `lapic_eoi` there would
+/// corrupt the LAPIC interrupt-priority state.
+#[inline]
+fn service_incoming_shootdown_and_drain(cpu: usize) {
     if cpu < apic::MAX_CPUS {
         let slot = &SHOOTDOWN_SLOTS[cpu];
         // Atomically claim the slot.  The single-writer rule on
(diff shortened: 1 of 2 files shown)


me1iissa (commented):
## Deep review (mm/TLB core) — APPROVE, 0 blockers

Reviewed `1a9ddad0` against master `47c1f227`. All six load-bearing claims verify against source. The reentrancy is correct.

### The six claims — each CONFIRMED against source
1. **EOI omission.** `service_incoming_shootdown_and_drain` has no `lapic_eoi`; only `handle_shootdown_ipi` (the ISR) calls `service_incoming_shootdown_and_drain(cpu)` then `apic::lapic_eoi()`. The in-spin caller never EOIs — correct: under IF=0 the 0xF0 IPI stays pending in the IRR (Intel SDM Vol. 3A §10.6.1), there is no delivered interrupt to acknowledge, and a spurious EOI would corrupt the LAPIC ISR/priority state.
2. **Slot disjointness — the key invariant, holds cleanly.** `targets = snapshot_active_mask(cr3) & !self_mask` (`mm/tlb.rs:525`), so `self_cpu ∉ targets`. The sender's *outgoing* requests live in `SHOOTDOWN_SLOTS[target]` (target ≠ self) and it polls those for the ack; the in-spin self-service touches only `SHOOTDOWN_SLOTS[self_cpu]` (this CPU's *incoming* slot, written by peers). Disjoint index spaces → the self-service cannot clear the sender's own outgoing request, no aliasing. In the mutual case (A→B, B→A) each CPU services the peer's incoming slot and releases the peer's ack; both acks complete, no cycle.
3. **#643 soundness invariant preserved.** The spin drains via `drain_pending_force_flush` = the FULL `flush_tlb()`+clear path (`mm/tlb.rs:969`), never `on_cpu_tick`'s clear-only path (`:859`). The PR does not move `post_force_flush` (production post stays in the timeout branch; the only new `post_force_flush` is inside Test 699's synthetic harness) and does not shrink `ACK_BOUND` — so the invariant's stated preconditions ("post trails the PTE mutation", ACK_BOUND unchanged) hold. Early (mid-spin) draining is strictly safe: it retires an *already-posted* (prior-timeout) force-flush earlier with a real full flush; the `cmpxchg(1→0)` means whichever of {spin, ISR, tick} wins does the flush and the others no-op — no double-service, no invariant coupling to the clear-only tick path.
4. **Lock-order under IF=0.** The helper takes no locks and sends no IPI: `cmpxchg`, `invlpg`, atomic stores, and `drain_pending_force_flush` (a `flush_tlb()` = MOV-to-CR3 + atomic flag clear, also lock-free). Called from a fault-path spin that may hold `VMM_LOCK`, it introduces no nested acquisition → no new deadlock edge.
5. **Refactor fidelity — verbatim.** The invalidation body (`cmpxchg` claim, cr3/va load, `local_invlpg_range`, firefox-test-core done-flag, `STAT_SHOOTDOWNS_HANDLED`, `drain_pending_force_flush`) is byte-identical to the old `handle_shootdown_ipi` body — the middle (invlpg) is not in any diff hunk, so it moved unchanged; only the function boundary, the EOI relocation, and comments changed.
6. **SMP=1-inert.** On one CPU `targets = active & !self_mask = 0` → `if targets == 0 { return true }` (`:526`) → the ACK spin never runs → the new call is never reached (also gated by the `SMP_ACTIVE` early-return at `:500`).

### Observations (non-blocking)
- **SDM §4.10.4.2 ordering (the LEAD's item b):** the ack is implicit in the `cmpxchg(1→0)`, which precedes the `invlpg` in program order, so a sender can observe `pending==0` before the servicing CPU's `invlpg` retires. This is **pre-existing master behavior, verbatim-moved — not introduced or worsened by #698**, and benign in practice: the `invlpg` is the next instruction with no intervening access to the range on the servicing core; frame-freeing callers are additionally covered by `quarantine_free`; non-freeing callers leave only a transient sub-instruction self-stale window with no access. (The firefox-test-core `SHOOTDOWN_DONE_FLAGS` is a stricter completion signal, but production polls `pending`; tightening that is out of scope here.) Calling the same body from the spin does not change this ordering.
- **Test 699 scope + gate.** The test correctly exercises the *primitive* (synthesize an incoming shootdown → service → assert `pending==0` + `shootdowns_handled` advanced → idempotent second call → composed force-flush drain), masks interrupts across the synthetic slot manipulation, and restores pristine state. It does **not** reproduce the genuine two-CPU mutual livelock (acknowledged; that is the census A/B, task #20). Reasonable for the single-threaded runner. Caveat: its pass is author-attested + static-verified here; the full test-mode suite green gate is unavailable due to the pre-existing test-243 DOUBLE_FAULT (task #21, A/B-confirmed unrelated). Recommend the merge gate = CI green (compile + feature-combo matrix) + census A/B (task #20) as the behavioral proof.

### Verdict
APPROVE — 0 blockers. Confidence ~0.9. Not run: no boots (Test 699 pass is author-attested + statically verified; behavioral livelock-break is the post-merge census A/B).

PR #700, opened 2026-07-03
proc: never hold THREAD_TABLE across schedule() in exit_group()

me1iissa:
## Summary

`exit_group()` resolved the caller's pid inside a `THREAD_TABLE` critical
section and, in the "thread no longer in the table" arm, called `schedule()`
while that guard was still live:

```rust
let pid = {
    let threads = THREAD_TABLE.lock();
    match threads.iter().find(|t| t.tid == tid) {
        Some(t) => t.pid,
        None => { schedule(); return; }   // guard still held
    }
};
```

`schedule()`'s dead-thread reap prologue re-acquires `THREAD_TABLE`
unconditionally. `THREAD_TABLE` is a non-reentrant spin lock, so entering
`schedule()` with the guard live makes the CPU spin forever on a lock it
already holds; because the build is `panic = "abort"`, the guard's `Drop` can
only run after `schedule()` returns — which it never does. Result: a
permanently-held `THREAD_TABLE` (lock byte stuck at 1) and a wedged CPU.

The fix copies the pid out of the locked region, drops the guard, then
branches — the whole scheduling path is now lock-free. The `Some` path is
behaviour-identical; only the lock scope is tightened.

## Scope / correctness

- **This is independent invariant-hardening, NOT a fix for the attempt-4 SMP=2
  census freeze.** That freeze's root attribution is separately disputed and
  under discriminator review; this PR should not be conflated with it.
- The window here is narrow: it needs the running thread to have already been
  removed from `THREAD_TABLE` by a sibling's group teardown while it is
  mid-exit. It is a **same-CPU** re-entrant self-deadlock, distinct from the
  census signature (both CPUs waiting, holder on neither).
- **SMP-independent** — the re-entrant self-deadlock fires on a single core as
  well; SMP=1 behaviour on the `Some` path is unchanged.
- Restores the invariant: no `THREAD_TABLE` guard may be live across a
  scheduling decision (dropping a lock before a call that re-acquires it is
  unconditionally sound — Intel SDM Vol. 3A §8.10.1, spin-lock ownership).

## Testing

- Compiles clean via the harness (`qemu-harness.py check`, rc=0).
- No deterministic test-mode regression is feasible: reaching the `None` arm
  requires the currently-running thread to be absent from `THREAD_TABLE`, which
  cannot be synthesised without destabilising the test-runner thread itself.
  Correctness is structural.

### DIFF
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -2433,11 +2433,27 @@ fn free_user_page_tables(cr3: u64) {
 /// killing the test runner thread itself.
 pub fn exit_group(exit_code: i64) {
     let tid = recover_current_tid();
+    // Resolve the pid inside a SHORT `THREAD_TABLE` critical section and copy
+    // it out, so the guard is released before any scheduling decision.  The
+    // `None` arm below reaches `schedule()`, whose dead-thread reap prologue
+    // re-acquires `THREAD_TABLE` unconditionally; entering `schedule()` while
+    // this guard were still live would re-enter a non-reentrant spinlock and
+    // wedge this CPU spinning on a lock it already holds (a lock the guard's
+    // `Drop` can no longer release, since `Drop` runs only after `schedule()`
+    // returns — which it never will).  Dropping the guard before the branch
+    // keeps the whole scheduling path lock-free.
     let pid = {
         let threads = THREAD_TABLE.lock();
-        match threads.iter().find(|t| t.tid == tid) {
-            Some(t) => t.pid,
-            None => { crate::sched::schedule(); return; }
+        threads.iter().find(|t| t.tid == tid).map(|t| t.pid)
+    };
+    let pid = match pid {
+        Some(p) => p,
+        None => {
+            // This thread is no longer recorded in `THREAD_TABLE` — a sibling's
+            // group teardown already reaped it while it was mid-exit.  There is
+            // nothing left to tear down; yield permanently with no lock held.
+            crate::sched::schedule();
+            return;
         }
     };
 


me1iissa (commented):
## Scheduler-lock-discipline review — APPROVE, 0 blockers

Reviewed `4dae80bd` against master `f0bf60fb`. Textbook lock-scope tightening; the fix is correct, minimal, and behavior-preserving. Reviewed as standalone in-class hardening (not the census-freeze fix — that's the separately-disputed held-across-*switch* class).

### The deadlock is real (justification confirmed)
`crate::sched::schedule()` (`sched/mod.rs:2171`) calls `reap_dead_threads_sched()` (`:871`), which unconditionally does `THREAD_TABLE.lock()` (`:922`). So the old `None`-arm `schedule()` under the live `threads` guard re-enters a non-reentrant `spin::Mutex` on the same CPU → self-deadlock; with `panic=abort` the guard's `Drop` never runs, so the lock byte is stuck at 1 and the CPU wedges. The fix is necessary, not cosmetic.

### The three points — each CONFIRMED
1. **Borrow/drop correctness.** The `pid` block is `{ let threads = THREAD_TABLE.lock(); threads.iter().find(…).map(|t| t.pid) }`. `find()` yields `Option<&Thread>` borrowing `threads`; `.map(|t| t.pid)` consumes it and returns an **owned** `Option<Pid>` (`Pid` = `u64`, `Copy`) — the transient `&Thread` borrow ends inside the `.map` call, so the block's value holds no borrow of `threads`, and `threads` drops at the closing `}`. The outer `match pid { Some(p) => p, None => { schedule(); return; } }` therefore runs with **no guard live on either arm**. This is the canonical "copy the scalar out under the lock, drop the lock, then decide" idiom. Provably correct.
2. **`None` arm reachability + behavior.** The `None` arm already existed pre-PR (the change only *moves* its `schedule()` out of the lock), so the PR is a strict improvement regardless of whether the arm is hot. The race sched-fixer describes (a sibling group-teardown reaps this thread's `THREAD_TABLE` entry while it is mid-`exit_group`) is a plausible way to reach it; "yield permanently with no lock held" is the right behavior — `schedule()`'s reap prologue is designed to run from a dying current thread, and the trailing `return` is unreachable-but-defensive (a dead current thread is never switched back in).
3. **Sweep — this is the only held-across-`schedule()` site (independently verified).** I enumerated every `schedule()` call site and read the lock context of each candidate. All use the correct `{ lock; mark Blocked/Sleeping; } schedule();` idiom or hold no lock: `ke/wait.rs:99/204/262` (inner `THREAD_TABLE` block closes before `schedule`; the DISPATCHER_REGISTRY guard is separately scoped — dropped before, re-acquired after), `ipc/waitlist.rs:367` (explicit `drop(bell)` before), `proc/mod.rs:2909` (sleep_ticks, block closes before), `syscall/mod.rs:2655` (waitpid, Blocked-commit dropped before), `proc/mod.rs:2117/2888` (dead-yield, no lock), `exit_group_inner` (comment-documented "no kernel locks held" at `schedule()`), and the virtio/pipe sites (lock-free completion/wait-outcome polls). A structural grep surfaced six more candidates — all six are false positives (matched `THREAD_TABLE.lock()`/`schedule()` text inside *comments*, e.g. proc/mod.rs:2503 and the syscall-trace docs, or adjacency between two scoped functions). The `exit_group` `None` arm was the lone real violator.

### Some-path unchanged
The diff replaces only the `pid`-resolution block; downstream uses of `pid` (the firefox-test diagnostic dump + teardown) are byte-identical on the branch. `Some` behavior is preserved.

### Verdict
APPROVE — 0 blockers. Confidence ~0.95. Not run: no boots (the change is a compile-checkable lock-scope tightening; CI feature-combo matrix covers build).

PR #702, opened 2026-07-03
mm: W215 write-protect #PF trap — name the in-place page-cache clobber writer (diagnostic, do-not-merge)

me1iissa:
## What

A **diagnosis-only** instrument (feature `w215-wptrap`) that names the residual W215 page-cache clobber writer by **trapping the store**, then freezing for a GDB autopsy. Nothing here changes default behaviour; do not merge until the writer is named (or the approach is retired).

## Why the passive shadows could not name it

The residual W215 corruption is an **in-place store into a live, reference-held page-cache frame** holding verified-correct library code (`libxul`/satellite `.text`). No allocator, free, or page-table event is hooked at the moment of that store, so the `w215_diag` per-phys provenance shadows (FREE/ALLOC/PROV) can classify the frame's *lifecycle* but can never name the *writer's RIP*. The only way to name an arbitrary in-place store is to **trap** it.

## Mechanism

1. At cache insert of a hot cluster code frame, **after** `insert_with_expected` has verified the frame content against the source bytes, the frame is made **read-only** on its higher-half direct-map alias (`PHYS_OFF + phys`). A verified code page has no legitimate kernel writer for the rest of its cache residency.
2. The next supervisor write to it raises a protection-violation `#PF` (P=1, W=1, U=0). The page-fault handler recognises it by `CR2 - PHYS_OFF`, records `[W215/WP-TRAP] writer_rip=… cr2=… phys=… rc=… tid=… pid=…`, then bugchecks (`W215_WP_TRAP`, `0xDEAD_000A`) to **freeze** the machine at the offending store.
3. Disarmed on eviction so a recycled frame is never left read-only.

## The load-bearing part — the huge-page split

The `[256 MiB, 1 GiB)` cluster is mapped by **2 MiB huge pages** in the higher-half map. Write-protecting one 4 KiB frame requires splitting the covering huge page. This reuses the **existing, boot-proven** split in `vmm::map_page` / `get_or_create_entry`, which fills every sub-PTE with the *same* physical base + flags the huge page carried — so **neighbouring frames keep their exact mapping and writability**; only the single target PTE loses its writable bit. A local `INVLPG` + cross-CPU shootdown retires the stale writable TLB entry.

`test_w215_wptrap_split_neighbour_safe` pins this invariant against the real primitive: after protect, the target is present+read-only and maps to its phys, **and** its neighbour is still present+writable and maps to *its* phys; after restore, the target is writable again.

## Coverage / limitations (diagnostic scope)

- Protects the higher-half direct-map alias **only** — the alias every kernel `memset`/`copy_nonoverlapping`/zero-fill path uses (the suspected in-place writer class). A writer through some *other* kernel mapping would not trap; user `.text` aliases are already `r-x` (read-only) so they are not the writer.
- Device **DMA** writes bypass CPU paging and cannot raise a `#PF`; the class under investigation is a CPU-side in-place store (the DMA-recycle class was addressed separately).
- Arming is bounded to a **64-slot** pool and the historical cluster window, so system perturbation is minimal.

## Byte-identical guarantee

Everything is gated behind `w215-wptrap` (implies `w215-diag`): the module, all call sites (`idt.rs` arm + PF hook, `cache.rs` disarm), the bugcheck code and its name arm. Verified: default (`--features ""`) and plain-`w215-diag` builds compile clean with **zero** reachable change.

## Reproduction status (honest)

W215 wrong-content is currently **non-reproducing on master** (the deep 14-flip saga is parked). This instrument arms the trap so that *if* the flaky in-place store recurs (historical trigger: SMP=2 + heavy Wikipedia load), the writer is named. If it never fires, the instrument still lands as tooling and the report says plainly "armed, no repro, no writer caught."

## Diff-size note

Larger than a normal diag change (new module + PF hook + huge-page-split reuse + arm/disarm + feature + test) — user-authorised for this investigation. Kept clean; no logical change split.

Cite: Intel SDM Vol. 3A §4.6, §4.10.4, §4.10.5; POSIX 1003.1-2024 mmap(2).

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -339,6 +339,15 @@ native-core-tests = []
 # preserving on-demand investigative access.  Implies `firefox-test`
 # because the diagnostic shares the firefox-test cache instrumentation.
 w215-diag = ["firefox-test"]
+# W215 write-protect #PF trap.  Arms a read-only mapping on the higher-half
+# direct-map alias of a verified cluster code page at cache insert; the next
+# supervisor write to it raises a protection-violation #PF that the page-fault
+# handler recognises and freezes on (naming the in-place clobber writer's RIP).
+# Implies `w215-diag` (shares the cluster-key / phys-window definitions and the
+# firefox-test cache instrumentation).  Diagnostic-only; default and plain
+# `w215-diag` builds are byte-identical (the module and all call sites are
+# gated on this feature).
+w215-wptrap = ["w215-diag"]
 # #582 torn-saved-RFLAGS-slot writer probe.  Arms a DR0 8-byte data-write
 # watch on context-switch victims' saved-RFLAGS slots
 # (`Thread::context.rsp + 0`) and classifies the writer of any store to
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -1444,6 +1444,28 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, frame: &mut InterruptF
         }
     }
 
+    // === W215 write-protect trap (w215-wptrap diagnostic) ===
+    //
+    // A supervisor write (U=0, W=1) to a *present* page (P=1) that lands on the
+    // higher-half direct map is a protection violation.  If the target frame is
+    // one this diagnostic write-protected — a verified cluster code page — the
+    // interrupted instruction is the in-place W215 clobber writer.  The
+    // faulting linear address IS the direct-map alias, so the physical frame is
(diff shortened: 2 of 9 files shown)


me1iissa:
Closing this do-not-merge W215 write-protect #PF trap diagnostic. W215 (the in-place page-cache clobber saga) is **non-reproducing** on fixed master following the SMP=2 teardown-safety chain (#703 PROCESS_TABLE-across-shootdown deadlock, #704/#705 CR3-drain liveness gate) — a 30-min SMP=2 Wikipedia content load armed 16 library code frames with WP-TRAP 0 / FREESHADOW 0 / TLB-TIMEOUT 0. The WP-trap tooling is preserved on a local branch if the saga ever recurs. No need to keep the diagnostic PR open.

PR #703, opened 2026-07-08
mm/pf: drop PROCESS_TABLE before the write-fault-path TLB shootdown — fix the SMP lock-vs-IPI deadlock (the [TLB/TIMEOUT] residual)

me1iissa:
## Summary

Fixes a **lock-vs-IPI cross-CPU deadlock** in `handle_page_fault` that is the dominant source of the residual `[TLB/TIMEOUT]` stream and the ~10× SMP=2 slowdown. On a live SMP=2 Wikipedia boot the fix drives `[TLB/TIMEOUT]` from the pre-fix ~0.92/s (369 by load-end) to **0**, and yields the healthiest SMP=2 boot the reliability program has seen.

## Root cause (GDB-autopsied on a live SMP=2 boot)

`handle_page_fault` takes `PROCESS_TABLE` (a spin mutex) on every fault to look up the faulting process's `VmSpace`, then — in the copy-on-write / write-fault fast path — mutates the leaf PTE and issues a cross-CPU TLB shootdown **while still holding it**. The shootdown spins (bounded, 10M iters) waiting for every peer CPU that maps the CR3 to ACK the IPI.

`#PF` is delivered through an interrupt gate, so a peer CPU handling its *own* fault runs `handle_page_fault` with `IF=0`. If it is spinning to acquire `PROCESS_TABLE` — the lock this CPU holds — it can never take the shootdown IPI (`IF=0` keeps vector `0xF0` pending in the IRR; Intel SDM Vol. 3A §10.6.1), so the initiator spins the full ACK bound and records a `[TLB/TIMEOUT]`. The bound lets it eventually proceed, but the pattern recurs on every concurrent write-fault under load.

Both vCPUs were caught `IF=0`:
- CPU A in `mm::tlb::shootdown_range` — the 10M-iter ACK-wait spin
- CPU B at the `lock cmpxchg`/`pause` spin acquiring `PROCESS_TABLE` inside `handle_page_fault`

Live kdb counters confirmed the concurrently-observed "#673 interlock livelock" was 99.9995% the **benign `Current`** case (`benign_current_defers` 12.5M vs `double_dispatch_defers` 60) — a downstream symptom of this deadlock, not an interlock defect. This is the ~80% residual that the `#698` ACK-spin reentrancy fix could not reach (that fix only rescues a target already inside the shootdown ACK-spin, not one wedged in a plain `PROCESS_TABLE` acquire).

## Fix

Release `PROCESS_TABLE` immediately before each of the four fast-path shootdowns (MAP_SHARED write-through, CoW private-copy install, single-owner make-writable, NX fixup). The PTE write is already published and nothing after it reads the process table; the shootdown and the CoW CAS-install take only Copy values (`cr3`, `page_addr`) and carry their own locking (`VMM_LOCK`/`mm_sem`) — exactly the pattern the file-backed demand path in the *same function* already uses (it drops `PROCESS_TABLE` before its VFS work).

Also includes (diagnostic enablers used to root-cause this):
- `600073c2` — split the #673 interlock accounting into benign `Current` vs genuine `OnStack` (silences the ~15M-line benign flood).
- kdb: expose `benign_current_defers`/`double_dispatch_defers` in `cpu-state` (one-call livelock classification).

## Verification (SMP=2, KVM, Wikipedia, 2048 MiB)

| Metric | Before fix (same phase) | With fix |
|---|---|---|
| `[TLB/TIMEOUT]` | ~113 at content-load → 369 by load-end (0.92/s) | **0** (held 0 through ~15 min / tick 91000) |
| `[SCHED/STARVE]` | 15+ | 1 (boot only) |
| `benign_current_defers` | 12.5M (CPUs wedged IF=0) | 2.3K |
| bugcheck / freeze / heap-corrupt | — | 0 |

Content procs (pid12/pid13 firefox-bin) run heavy and healthy (pf>400K, sc>9M, network flowing).

## Independent adversarial review

Confirmed SAFE for the PROCESS_TABLE class: the PTE writes are idempotent/atomic; the CoW install self-guards under `VMM_LOCK` (the W215/d4fb7fa two-frames-one-VA class is **not** reopened); no residual PROCESS_TABLE cycle; borrow placements sound.

## Tracked follow-ups (out of scope here — pre-existing, separate instances)

- **mm_sem lock-vs-IPI class (same shape, different lock):** `free_process_memory` (`proc/mod.rs:2215`), `free_vm_space` (`:2337`), and `clone_for_fork` (`vma.rs:874`) hold `mm_sem.write()` across their own shootdown, against a #PF-fast-path peer spinning on `mm_sem.read()` (IF=0). Empirically **not** the dominant contributor (TLB/TIMEOUT stayed 0 through heavy fork/teardown here), and `mm_sem` is load-bearing for the W216 exclusion invariant, so the fix is the `service_incoming_shootdown_and_drain` reentrancy pattern on the `mm_sem` acquire path, not a bare drop. Separate PR.
- **exit-teardown liveness gate:** `exit_thread`'s `all_dead → free_process_memory` (`proc/mod.rs:1985`) lacks the `any_sibling_on_cpu` cross-CPU gate that `exit_group_inner` has (`proc/mod.rs:2825-2841`) — a pre-existing narrow UAF window (the file-backed demand path already exposes a larger one). Minimal fix: mirror the gate. Separate PR.

Cite: Intel SDM Vol. 3A §10.6.1 (IPI delivery / a wedged target is the sender's responsibility), §4.10.4.2 (TLB shootdown protocol).

### DIFF
--- a/kernel/src/arch/x86_64/idt.rs
+++ b/kernel/src/arch/x86_64/idt.rs
@@ -1558,6 +1558,21 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, frame: &mut InterruptF
             let old_phys = pte & 0x000F_FFFF_FFFF_F000;
             let new_pte = old_phys | page_flags | PAGE_PRESENT;
             crate::mm::vmm::write_pte(cr3, page_addr, new_pte);
+            // Release PROCESS_TABLE before the cross-CPU shootdown.  The
+            // shootdown IPI needs every target CPU to run its handler to ACK,
+            // but any peer CPU that is itself in `handle_page_fault` spins on
+            // PROCESS_TABLE with interrupts disabled (#PF is an interrupt
+            // gate).  While we hold this lock that target can never take the
+            // IPI, so we spin the full 10M-iter ACK bound — a lock-vs-IPI
+            // cross-CPU deadlock that stalls every concurrent fault and shows
+            // up as the residual `[TLB/TIMEOUT]` stream on SMP.  The PTE write
+            // above is already published and nothing below reads the process
+            // table; the shootdown takes only the Copy `cr3`/`page_addr` — so
+            // drop the lock first, exactly as the file-backed demand path below
+            // already does before its VFS work.  (Intel SDM Vol. 3A §10.6.1:
+            // IPI delivery to a CPU that cannot service it is the sender's
+            // responsibility; here we remove the reason it cannot service.)
+            drop(procs);
             crate::mm::tlb::shootdown_page(cr3, page_addr);
             // NOTE: this does not set the PTE dirty bit nor mark the backing
             // page-cache frame dirty.  Inert today (VFS writeback is not yet
@@ -1627,6 +1642,13 @@ fn handle_page_fault(faulting_addr: u64, error_code: u64, frame: &mut InterruptF
                     }
                     // else: unrelated bump — proceed; the CAS guards the install.
                 }
+                // Release PROCESS_TABLE before the CAS-install + cross-CPU
+                // shootdown below (same lock-vs-IPI deadlock as the MAP_SHARED
+                // arm above).  All uses of `vm_space` — including the gen-abort
+                // re-check — are complete; the install is a self-guarding CAS
+                // (`map_page_in_cow_if_unchanged`) and the refcount ops carry
+                // their own locking, so none of this needs the process table.
+                drop(procs);
                 // The private copy (new_phys) takes sole ownership; mark it
                 // before publishing so a concurrent faulter that observes the
                 // installed PTE never sees a 0-refcount frame.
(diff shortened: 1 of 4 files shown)


me1iissa:
### SMP=2 verification — full-run outcome

The verification boot ran clean and healthy for **~19 min** (tick 116000) on SMP=2/KVM/Wikipedia:
- `[TLB/TIMEOUT]` = **0** throughout (was ~0.92/s pre-fix), `[SCHED/STARVE]` = 1 (boot only), **0** bugcheck/heap-corrupt/freeze.
- Throughput recovered far past the degraded baseline: `sc` reached ~9.6M (degraded SMP=2 runs stalled ~3M).

It did **not** render — it then hit the **separate, pre-existing, parked W101 reactor wake-drain** (all 50 pid1 threads `Blocked`, content procs `STUCK_IN_NR=7`/poll, reactor `rloop` spinning with `pf`/`sc` frozen). That gate is independent of this change (a TLB-shootdown lock-ordering fix cannot cause a userspace event-loop wake stall), and the machine reached it *faster and cleaner* than any prior SMP=2 run precisely because the TLB-timeout degradation is gone.

Net: this PR fully meets its stated target (eliminate the `[TLB/TIMEOUT]` deadlock / restore SMP=2 throughput). The render/G1 outcome remains gated by the pre-existing W101 wake-drain family, tracked separately.

PR #705, opened 2026-07-09
proc: unconditional free_process_memory in exit_group + tick-bounded CR3-drain wait

me1iissa:
Two commits forming one coherent teardown fix, following PR #704.

## 1. `exit_group_inner` frees unconditionally — close the skipped-free leak (`474ba50d`)

`exit_group_inner` guarded its `free_process_memory` call with an `any_sibling_on_cpu` scan that, when a sibling was on-CPU, **skipped the free entirely** (not deferred it) — and nothing ever retried, so the address space leaked (the orphan-reap sweep / `waitpid` later drops the still-`Some` `VmSpace`, whose `Drop` frees no page-table frames). PR #704 moved the cross-CPU teardown-race protection *into* `free_process_memory` (its CR3-drain wait), making that hand-rolled pre-check redundant — and its skip-behaviour now purely a leak. Call unconditionally and let the drain-wait handle the on-CPU sibling. (The CR3-keyed drain-wait is also strictly broader than the old PID-keyed scan — it incidentally closes a vfork/`CLONE_VM` gap the scan had.)

## 2. Bound the CR3-drain wait by TSC ticks, not spin count (`230baf1a`)

#704's drain-wait used a fixed `100_000_000`-iteration `PAUSE` bound. `PAUSE` latency varies by orders of magnitude across host/KVM/TCG, so a count sized to mean "seconds" can expire in **under one 10ms timer tick** on another platform. Live SMP=2 testing (a ~24-thread Firefox `exit_group`, once commit 1 began exercising this path) showed exactly that: the wait hit the bound and freed while a `Dead` sibling of the *same process* was still draining off the other CPU — the very UAF the wait exists to close. (Root confirmed from the serial: the un-draining CR3 was the process's own last Dead thread, not a shared address space.)

Bound by wall-clock ticks instead: sample `irq::get_ticks()` (TSC-derived — constant-rate TSC per Intel SDM Vol. 3B §17.17, so it advances even if this CPU's LAPIC timer is dead, guaranteeing termination) every ~1024 spins, give up after ~300ms. The normal drain completes in 1-2 ticks, so the safety valve now fires only for a genuinely wedged sibling. **This also hardens #704's merged drain-wait** (same mis-calibration, never triggered from `exit_thread`'s timing). Commit 1's correctness depends on this.

## Verification (SMP=2, KVM, Firefox → Wikipedia)

Exercised the exact triggering scenario — **254 group exits, 462 threads, 3 vfork parents** — with **spin-WARN = 0** (the pre-fix bound produced it; the tick bound eliminates it), 12 PIDs freed (leak closed), **0** bugcheck/panic/corruption, `TLB/TIMEOUT`=0. Independently reviewed **GO** (all invariants from #704's 3-round review preserved; termination guaranteed; hot path unchanged).

Cite: Intel SDM Vol. 3A §4.10 (CR3 validity), Vol. 3B §17.17 (constant-rate TSC).

### DIFF
--- a/kernel/src/proc/mod.rs
+++ b/kernel/src/proc/mod.rs
@@ -2201,15 +2201,21 @@ pub fn free_process_memory(pid: Pid) {
     // held: the not-yet-`Dead` caller (both call sites mark the caller `Dead`
     // only just before `schedule()`, after this returns) is preemptible and
     // resumes the wait after any preemption, and each sibling drains via its own
-    // CPU's timer, independent of this one.  `CR3_DRAIN_SPIN_BOUND` is a safety
-    // valve against a sibling that legitimately holds the CR3 for a long
-    // kernel excursion (this kernel does not preempt Ring 0, so a `Dead`
-    // sibling finishing a slow I/O-backed `#PF` keeps its CR3 until it reaches
-    // its own reschedule point) or is pathologically wedged — on expiry we log
-    // and proceed to free, no worse than the pre-fix unconditional free.  The
-    // bound is deliberately large (its wall-clock cost is on the order of
-    // seconds of `PAUSE`, not milliseconds) so it never fires for a legitimate
-    // in-flight fault.
+    // CPU's timer, independent of this one.  The wait is bounded by wall-clock
+    // TICKS, not a raw spin count: the sibling drains when its CPU takes a timer
+    // tick and reschedules off the now-`Dead` thread (~1-2 ticks), and a
+    // spin-count bound is unreliable — `PAUSE` latency varies by orders of
+    // magnitude across host / KVM / TCG, so a fixed count is either a
+    // multi-second stall or (as observed under KVM) expires before even one
+    // tick, abandoning the wait prematurely and proceeding with the CR3 still
+    // live.  `get_ticks()` is TSC-derived (Intel SDM Vol. 3B §17.17,
+    // constant-rate TSC), so it advances even if this CPU's LAPIC timer is dead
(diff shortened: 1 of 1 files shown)


me1iissa:
### Verification correction (soak continued past the first check)

An earlier snapshot showed 0 WARN; the fuller soak (~215s, **348 group exits, 462 threads, 3 vfork parents**) recorded **1** `[PROC] WARN ... after 30 ticks` — I want the record accurate.

Characterized: the WARN was `free_process_memory(pid=21)`, a Firefox process that **SIGABRT'd** (`exit_code=-6`); one of its own Dead threads held the CR3 for >300ms during teardown. Notably:
- **`SCHED/STARVE` total = 1 (boot only)** — so it was *not* scheduler starvation.
- **No fault/crash followed the free** (the potential UAF did not materialize; `PID 21 memory freed` → `waitpid: reaped PID 21` cleanly).

So this is the tick-bounded safety valve firing for a genuine anomaly (a thread holding the CR3 >300ms — a stuck thread or a stale CR3-active-mask entry; cause not yet isolated), which is exactly the behavior the review GO'd ("fires only for a genuinely wedged sibling; log+proceed no worse than pre-#704"). It is **not** the premature-abandonment bug this PR fixes — that was the spin-count bound expiring in *under one tick*; the tick bound now gives the normal ~1-2-tick drain a 15-30× margin, and 347/348 exits drained cleanly with no WARN.

Net: this PR closes the leak and fixes the premature abandonment; the one residual WARN is the designed valve on a rare anomaly (worth a separate look at whether it's a stuck thread vs a `note_cr3_unload` mask-leak, tracked separately), not a regression.

PR #717, opened 2026-07-13
mm: MAP_FIXED must evict a raced overlap, not spuriously ENOMEM (Phase-3 SMP TOCTOU)

me1iissa:
## Problem
`sys_mmap`'s MAP_FIXED path is a 3-phase sequence so it can honour POSIX/Linux mmap(2)'s rule that MAP_FIXED discards any overlapping existing mapping: Phase 2a removes the overlapping VMA records under `PROCESS_TABLE`, Phase 2b clears the PTEs **lock-free** (its cross-CPU TLB shootdown must not run under the lock), Phase 3 re-acquires the lock and inserts.

The lock is dropped between Phase 2a/2b and Phase 3. On SMP a sibling thread on the shared CR3 can map into the just-cleared range in that window, so Phase 3's `insert_vma` fails with `Overlap`. The retry there was gated on `!is_fixed` (it re-picks a fresh base — valid only for a kernel-chosen address), so a **MAP_FIXED** placement fell straight through to ENOMEM. But mmap(2) forbids MAP_FIXED from failing on overlap: *"the overlapping part of the existing mapping(s) will be discarded."*

Under heavy concurrent dynamic-linker load (Firefox mapping `libsqlite3` segments via MAP_FIXED, SMP=2) this surfaced as `[MMAP-ERR] insert_vma failed ... attempts=0`; the linker then wrote the intended-writable page — now a stale read-only reservation — and took an unhandled write `#PF` (SIGSEGV) in the content process.

## Fix
On a Phase-3 `Overlap` for a MAP_FIXED (non-NOREPLACE) placement, **evict the raced mapping and retry at the same base** — the MAP_FIXED analog of the already-closed non-FIXED TOCTOU. The eviction mirrors the Phase 2a/2b split exactly: remove the VMA record under `PROCESS_TABLE`, then **drop the lock** before `unmap_and_free_range_in` (which issues a synchronous cross-CPU TLB shootdown and takes `mm_sem` internally). **No lock is ever held across the shootdown** — the lock-across-shootdown deadlock class stays closed. Retry is bounded (2×, then ENOMEM) so a sibling that keeps re-racing cannot livelock. MAP_FIXED_NOREPLACE is excluded (its overlap is a genuine caller error).

## Verification (SMP=2 Wikipedia, KVM)
**Before** (no fix): the race fires → `[MMAP-ERR] pid=6 ... flags=0x12 ... insert_vma failed attempts=0` → content-proc write-`#PF` SIGSEGV.

**After** (this fix, same amplified timing, 2/2 boots): the same race fires →
`[MMAP] pid=6 MAP_FIXED re-evicted a raced overlap at base=0x7efff0512000 len=0x9000 (attempt 1)` → **0 `[MMAP-ERR]`, 0 crash, reached `[FFTEST] DONE`** (no deadlock).

Also: 4 additional SMP=2 boots with no regression/deadlock; existing MAP_FIXED coverage (Tests 44, 107c) preserved; the restructure is behaviour-preserving for the non-FIXED re-pick, MAP_FIXED success, ESRCH, and ENOMEM paths.

## Scope
This removes one content-proc crash face (and, with it, the downstream `error_code=0x7` write-fault into the stale r-x reservation). SMP=2 heavy-page (Wikipedia) rendering remains flaky due to a separate memory-corruption class that is out of scope here.

Ref: POSIX.1-2017 / Linux mmap(2), MAP_FIXED.

### DIFF
--- a/kernel/src/syscall/mod.rs
+++ b/kernel/src/syscall/mod.rs
@@ -3552,95 +3552,142 @@ pub(crate) fn sys_mmap(addr_hint: u64, length: u64, prot: u32, flags: u32, fd: u
     // syscall result rather than returning directly so the reap below always
     // runs (Phase 2a above may have queued an inode free even on the rare
     // process-exited-between-phases path).
+    //
+    // Two placement TOCTOU windows are closed here.  Both are opened because
+    // Phase 1 (base selection) and Phase 2b (MAP_FIXED PTE clear) release
+    // PROCESS_TABLE before this point, so on SMP a sibling thread on the shared
+    // CR3 can claim the range in between:
+    //
+    //   * non-FIXED — the base was picked by find_free_range /
+    //     find_free_stack_range in Phase 1 under a lock since released.  A
+    //     racing sibling mmap can take the gap, so a bare insert here would
+    //     wrongly return ENOMEM while free address space remains — a spurious
+    //     failure POSIX mmap(2) forbids for a kernel-chosen address.  Re-pick
+    //     and re-insert under the lock we now hold (find→insert is atomic while
+    //     held, so the retry cannot itself race).  Safe because a non-FIXED
+    //     mmap installs NO page-table entries — the base lives only in the VMA
+    //     record.
+    //
+    //   * MAP_FIXED — POSIX/Linux mmap(2): "if the memory region specified by
+    //     addr and len overlaps pages of any existing mapping(s), then the
+    //     overlapping part of the existing mapping(s) will be discarded."  So
+    //     MAP_FIXED must EVICT any overlap and place at the caller's exact
+    //     base — it must not fail on overlap.  Phase 2a/2b already cleared the
+    //     range, but a sibling can race a mapping back into it in the
+    //     lock-dropped window; the insert then hits Overlap.  Evict that raced
+    //     mapping and retry (bounded, ≤2×, then ENOMEM — a wrong success is
+    //     worse than ENOMEM, and the bound is livelock-safe against a sibling
+    //     that keeps re-racing).  The eviction MIRRORS the Phase 2a/2b split:
+    //     remove the VMA record under PROCESS_TABLE, then DROP the lock before
+    //     `unmap_and_free_range_in`, which issues a synchronous cross-CPU TLB
+    //     shootdown and acquires mm_sem internally — holding PROCESS_TABLE (or
+    //     mm_sem) across it would reintroduce the lock-across-shootdown
+    //     deadlock class closed on the #PF path.  MAP_FIXED_NOREPLACE is
+    //     excluded: its overlap is a genuine caller error, not a race to evict.
     let result = 'phase3: {
-        let mut procs = crate::proc::PROCESS_TABLE.lock();
-        let proc = match procs.iter_mut().find(|p| p.pid == pid) {
-            Some(p) => p,
-            None => break 'phase3 -3, // ESRCH (process exited between phases)
-        };
-        let space = match proc.vm_space.as_mut() {
-            Some(s) => s,
-            None => break 'phase3 -3,
-        };
+        enum Step { Done(Option<u64>), Evict }
 
-        // Close the Phase-1/Phase-3 placement TOCTOU for kernel-chosen
-        // (non-FIXED) mappings.
-        //
-        // For a non-FIXED mmap the base was picked by find_free_range /
-        // find_free_stack_range in Phase 1 under a PROCESS_TABLE lock that was
-        // then RELEASED before this point.  On SMP a concurrent mmap into the
-        // SAME address space (sibling thread / shared CR3) can claim that gap
-        // in the window, so `insert_vma` here would fail with Overlap and we
-        // would wrongly return ENOMEM even though free address space remains —
-        // a spurious failure that POSIX mmap(2) forbids for a kernel-chosen
-        // address.  (Observed live under SMP=2: a sibling thread's memfd
-        // MAP_SHARED mmap lost this race, the userland SharedMemory map failed,
-        // and a downstream NULL dereference of the unmapped region followed.)
-        //
-        // Re-pick the gap and re-insert under the lock we now hold; with the
-        // lock held the find→insert pair is atomic, so the retry cannot itself
-        // race.  This is safe because a non-FIXED sys_mmap installs NO page
-        // table entries (frames are demand-faulted later): the chosen base
-        // lives only in the VMA record, so re-choosing it here is complete —
-        // the return value and every subsequent fault use the recorded base.
-        //
-        // MAP_FIXED placements keep the caller-supplied base unchanged (no
-        // re-pick): their Overlap is a genuine caller error and Phase 2b has
-        // already edited the page table for that exact base.
         let mut cur_base = base;
-        let mut attempts = 0u32;
-        let installed: Option<u64> = loop {
-            let vma = VmArea {
-                base: cur_base,
-                length,
-                prot,
-                flags,
-                backing: backing.clone(),
-                name,
-            };
-            match space.insert_vma(vma) {
-                Ok(()) => break Some(cur_base),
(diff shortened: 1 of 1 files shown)


me1iissa:
## VERDICT: APPROVE

Independent adversarial review, code-level (no QEMU boots — CI + the author's before/after already exist). I studied the actual call graph rather than the PR prose, traced every exit path, and tried to construct a concrete wrong-mapping / crash / deadlock interleaving. I could not. The core fix is correct, POSIX-conformant, deadlock-safe, and bounded. Approving, with two low-severity nits I recommend clearing before merge (one of them pollutes a serial marker the current investigation depends on) and two informational notes.

### What I verified holds

1. **Lock/shootdown discipline is correct (traced, not trusted).** The `'evict` iteration takes `PROCESS_TABLE` only inside the inner block; `space.remove_range(cur_base, length)` runs under it, then the block ends (`procs` dropped → lock released) *before* the `Step::Evict` arm calls `unmap_and_free_range_in(cr3, cur_base, length)`. That routine acquires `mm_sem` (read) and spins on a synchronous cross-CPU TLB shootdown (`shootdown_range`) internally, and takes no `PROCESS_TABLE`. So no lock — `PROCESS_TABLE`, `mm_sem`, or VMA/space lock — is ever held across the shootdown IPI wait. This keeps the lock-across-shootdown deadlock class (the `handle_page_fault` PROCESS_TABLE-across-shootdown and the `mm_sem`+PROCESS_TABLE #PF-reentrancy lineage) closed. The eviction is a faithful mirror of the existing Phase-2a/2b split. Even the `break 'evict None` (proc-gone) exits run `procs`'s destructor, so the lock is released on every path.

2. **POSIX MAP_FIXED semantics are honoured.** mmap(2): the overlapping part of any existing mapping "will be discarded," and the placement is at the caller's exact address. The fix evicts the raced overlap and re-inserts at the unchanged `cur_base` (== `base` for the FIXED path — the base is never re-picked for FIXED, only for the kernel-chosen non-FIXED arm). Partial-overlap correctness is delegated to the same `remove_range` Phase 2a uses, which handles all four cases (contained / hole-punch split / right-shrink / left-shrink) with file-offset fix-ups; `unmap_and_free_range_in` only clears PTEs inside `[cur_base, cur_base+length)`, so a sibling mapping that extends outside the request keeps its non-overlapping remnant. I walked the before/after/left/right/contained interleavings — all resolve to the correct final VMA + PTE state.

3. **Bound is livelock-safe and non-FIXED is behaviour-identical.** FIXED evicts at `fixed_evict_attempts` 0 and 1, then a 3rd overlap falls to `Err(_) => Done(None)` → ENOMEM: at most 2 shootdowns, no unbounded loop. The non-FIXED re-pick arm keeps its `attempts < 8` bound and its `find_free_range`/`find_free_stack_range` logic verbatim; the two counters are cleanly independent (non-FIXED only ever runs one `'evict` iteration; FIXED never touches `attempts`). `note_mmap_placement` is still called once, under the lock, with the same arguments, so Test 107c (MAP_FIXED must not drag `mmap_hint`) is preserved.

4. **No torn/half-inserted reservation and no new refcount hazard.** On the exhaustion-ENOMEM path the range is owned by the sibling's re-raced VMA (not a hole from our side); on success the full range is our single VMA. `unmap_and_free_range_in` only frees a frame when its refcount reaches 0, and a cache-backed frame retains the cache's own ref, so the file-backed safety argument is unchanged. `reap_pending_inodes()` runs on every exit path (success, exhaustion, and proc-gone), so an shm inode unpinned by an evict's `remove_range` is never leaked.

The structural approach is sound for this kernel: because the overlap-unmap issues a cross-CPU shootdown that cannot run under the placement lock, the atomic "munmap-overlap + insert under one write lock" model is unavailable, and the drop-lock 3-phase structure necessarily opens this TOCTOU. A bounded evict-and-retry is the right adaptation.

### Findings (ranked)

**F1 — LOW/MEDIUM — ESRCH is silently reclassified to ENOMEM and emits a false `[MMAP-ERR]` line.**
`kernel/src/syscall/mod.rs:3600` and `:3604`:
```
None => break 'evict None, // ESRCH (exited between phases)
...
None => break 'evict None,
```
Pre-fix these were `break 'phase3 -3` (ESRCH). Now `break 'evict None` funnels the proc-not-found and `vm_space == None` cases into `installed == None`, which returns **-12 (ENOMEM)** and prints
`[MMAP-ERR] pid=… insert_vma failed: … fixed_evict_attempts=0` (`:3688`) — even though `insert_vma` was never attempted, so the message is factually wrong. The PR description's "behaviour-preserving for … ESRCH … paths" claim does not hold for this path. Functional impact is near-nil (the thread whose process was torn down mid-syscall is being killed and its return value is moot), but the spurious `[MMAP-ERR]` is emitted on exactly the serial marker the in-flight sig-B / heavy-page verification keys on — a benign process-exit race would now masquerade as an insert failure and could derail confirmation that this very fix worked. Cheap fix: carry a sentinel (e.g. `Option<i64>` or a small enum) so the proc-gone / space-none breaks return -3 without the log, distinct from the exhaustion ENOMEM. Recommend clearing before merge.

**F2 — LOW / cosmetic — dead `'phase3` label.**
`kernel/src/syscall/mod.rs:3587` — `let result = 'phase3: {` — no `break 'phase3` remains inside the block (the old `break 'phase3 -3` became `break 'evict None`). This is an unreferenced label and will trip the `unused_labels` lint (warn-by-default; harmless to green CI unless warnings are denied). Drop the label. (Flagged from reading, not a compile.)

**F3 — LOW / PRE-EXISTING (not introduced here) — MAP_FIXED_NOREPLACE loses EEXIST on a raced overlap.**
`is_noreplace` conflicts are detected in Phase 1 (`:3302`, returns EEXIST) under the lock, but the lock is dropped before Phase 3, and the evict arm is correctly gated `!is_noreplace`. A sibling that races a mapping into the range in the Phase-1→Phase-3 window therefore reaches `insert_vma` → Overlap → `Err(_) => Done(None)` → **ENOMEM**, where mmap(2) MAP_FIXED_NOREPLACE specifies **EEXIST**. This is pre-existing (the same fall-through existed before #717) and astronomically rare, so it does not block; noting it because the evict rework touches the adjacent arm and it would be a natural one-line improvement to route the NOREPLACE Overlap to -17 while here.

**F4 — INFO — the raced-MAP_FIXED evict path has no automated test.** Tests 44 and 107c cover single-threaded MAP_FIXED placement and hint behaviour, not a two-thread race into the just-cleared range. A deterministic in-kernel reproduction is hard (it needs the SMP Phase-2a→Phase-3 window), and the sibling non-FIXED TOCTOU it mirrors also landed without a race test, so absence is consistent with precedent and not a blocker. If a `test_runner.rs` two-thread stressor that drives concurrent same-VmSpace MAP_FIXED at a contended base is feasible, it would be worth adding as a follow-up.

**F5 — INFO (acceptable, documented) — a 3× consecutive re-race still returns ENOMEM for a MAP_FIXED.** The ≤2 bound means a hostile/unlucky sibling that re-maps the range three times in the window yields ENOMEM rather than an eventual success. This trades strict "MAP_FIXED never fails on overlap" for livelock safety, is explicitly documented in the new comment, and is strictly better than the pre-fix behaviour (which returned ENOMEM on the *first* race). Acceptable.

### Bottom line
The fix does what it claims: it removes the MAP_FIXED Phase-3 TOCTOU that spuriously ENOMEM'd a raced overlap, without reopening the shootdown-under-lock deadlock class, and it correctly preserves the non-FIXED, MAP_FIXED-success, and hint paths. No correctness, crash, or deadlock defect found. **APPROVE.** Please address F1 (and ideally F2) before merge — F1 because it emits a misleading `[MMAP-ERR]` into the serial stream the active investigation is reading.

Refs: POSIX.1-2017 mmap(2) (MAP_FIXED / MAP_FIXED_NOREPLACE overlap-discard and EEXIST semantics); Intel SDM Vol. 3A §4.10.4–5 (TLB shootdown / page-table visibility ordering).

PR #718, opened 2026-07-14
harness: coerce session pid to int in _pid_alive (fix list/status crash on stale session JSON)

me1iissa:
## Problem
A session JSON written by an older harness version stored the pid as a string. `_pid_alive` passed it straight to `os.kill` → `TypeError: 'str' object cannot be interpreted as an integer`, crashing `list` (and any subcommand walking sessions) whenever such a record is present in `~/.astryx-harness/`.

## Fix
Coerce the pid to `int` at `_pid_alive` entry; a non-numeric pid is treated as not-alive, so a malformed record degrades to a dead session instead of an unusable subcommand. `_pid_running` inherits the guard via its `_pid_alive` gate.

Additive — no JSON output-field changes.

## Verification
`python3 scripts/qemu-harness.py list` crashed before the fix with the stale record present; after, it returns the live-session JSON correctly (verified against a running session).

### DIFF
--- a/scripts/qemu-harness.py
+++ b/scripts/qemu-harness.py
@@ -911,12 +911,27 @@ def _session_path(sid: str) -> Path:
 def _events_path(sid: str) -> Path:
     return HARNESS_DIR / f"{sid}.events.jsonl"
 
+def _normalize_pid(value):
+    """Session JSONs written by older harness versions may carry the pid as
+    a string, and signal sites pass it straight to os.kill — normalize at
+    the read boundary. Malformed or non-positive values become None so a
+    corrupt record reads as a dead session (os.kill with 0/negative pids
+    would signal a process group, never a single stale QEMU)."""
+    try:
+        pid = int(value)
+    except (TypeError, ValueError):
+        return None
+    return pid if pid > 0 else None
+
 def _load_session(sid: str) -> dict:
     p = _session_path(sid)
     if not p.exists():
         _err(f"Unknown session: {sid}")
     with p.open() as f:
-        return json.load(f)
+        data = json.load(f)
+    if "pid" in data:
+        data["pid"] = _normalize_pid(data["pid"])
+    return data
 
 def _save_session(data: dict):
     p = _session_path(data["sid"])
@@ -924,12 +939,11 @@ def _save_session(data: dict):
         json.dump(data, f)
 
 def _pid_alive(pid) -> bool:
-    # Session JSONs written by older harness versions can carry the pid as a
-    # string; coerce before signalling so `list`/`status` don't crash on a
-    # stale record. A non-numeric pid means a malformed record → not alive.
-    try:
-        pid = int(pid)
-    except (TypeError, ValueError):
+    # Defence in depth for callers that read session JSON without
+    # `_load_session` (e.g. the `list` sweep does its own json.load): a
+    # malformed pid reads as not-alive rather than crashing the subcommand.
+    pid = _normalize_pid(pid)
+    if pid is None:
         return False
     try:
         os.kill(pid, 0)
--- a/scripts/qemu-harness.py
+++ b/scripts/qemu-harness.py
@@ -923,7 +923,14 @@ def _save_session(data: dict):
     with p.open("w") as f:
         json.dump(data, f)
 
-def _pid_alive(pid: int) -> bool:
+def _pid_alive(pid) -> bool:
+    # Session JSONs written by older harness versions can carry the pid as a
+    # string; coerce before signalling so `list`/`status` don't crash on a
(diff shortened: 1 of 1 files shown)


me1iissa:
## VERDICT: APPROVE

Code-level review of the `_pid_alive` int-coercion. The fix does exactly what it claims — `list`/`status` no longer crash on a stale session JSON whose `pid` was persisted as a string — and it does so safely. I walked all `_pid_alive`/`_pid_running` call sites and the `os.kill` semantics for the coercion edge cases. Two low-severity, out-of-scope follow-up notes below; neither blocks.

### What holds up

**Adversarial angle (1) — no caller relies on the old TypeError.** Every `_pid_alive` call site consumes it as a bool: the read paths (`list`/`status`/`snap`/`events` status lines, e.g. `alive = _pid_alive(pid) if pid else False`) and the signal paths (`if pid and _pid_alive(pid): os.kill(...)`). None wrap it in a `try` expecting a raise; the pre-fix crash was an *unhandled* TypeError propagating out of the command handler. `int(pid)` failing → `return False` (not-alive) is the correct degradation for a malformed record. `_pid_running` inherits it through its `if not _pid_alive(pid): return False` gate, so a string pid there also degrades cleanly (and its later `/proc/<pid>/stat` read tolerates a numeric-string pid via f-string interpolation).

**Adversarial angle (2) — no group-signal hazard, and a `pid <= 0` guard is not required for safety.** `_pid_alive` hardcodes `os.kill(pid, 0)`; per POSIX kill(2) a null signal delivers nothing — it only does existence/permission checking — so even a coerced `0`/negative can never signal a process group *through this function*. The real-signal sites (the SIGTERM→SIGKILL escalation in `cmd_stop`, the livelock auto-reap, the oracle-stub teardown) receive the **un-coerced** `sess.get("pid", 0)` value, so a group-kill there would require a genuine negative *int* in the JSON — which real harness data never produces (it persists the positive QEMU child pid, as int or, in stale records, as the string of that positive int). A corrupt `"pid": "-1"` string reaches `os.kill("-1", SIGTERM)` → TypeError, **not** a group kill; and `os.kill(0, ...)` is unreachable at the signal sites because the `if pid` guard makes int `0` falsy. So a `pid <= 0` guard is optional hardening, not a safety requirement.

**Adversarial angle (3) — coercion surface is correct.** Verified: `None`→TypeError (caught), `"12345"`→12345, `"3.5"`→ValueError (caught), `3.0`→3, `True`→1 (harmless). The `(TypeError, ValueError)` tuple covers exactly the raising cases.

The fix also *improves* the `stop` path for the common stale case as a bonus: a string-pid record whose process is already gone now yields `os.kill(int(pid), 0)` → ProcessLookupError → `_pid_alive` returns False → the `os.kill(pid, SIGTERM)` is skipped, so `stop` cleans up instead of crashing (pre-fix it crashed at the `_pid_alive` check).

### Findings (both LOW, out of the stated scope, neither a regression)

**F1 — the coercion is local to `_pid_alive`; the direct real-signal sites still pass the raw pid.** In `cmd_stop` / the livelock auto-reap / the oracle-stub teardown, the pattern is `if pid and _pid_alive(pid): os.kill(pid, signal.SIGTERM)` with `pid = sess.get("pid", 0)` un-coerced. For the *rare* case of a stale string-pid record whose process is **still alive**, `_pid_alive` now returns True and control flows into `os.kill("<string>", SIGTERM)` → TypeError, which the surrounding `except (ProcessLookupError, PermissionError)` does not catch. This is not a regression (pre-fix that same record crashed one line earlier, in the `_pid_alive` check) and the common dead-process stale record is handled cleanly, but the PR's implicit "stale string-pid records no longer crash the harness" holds only for the read paths, not the signal paths. Cheap complete fix: normalize `pid` to int once at the read boundary — e.g. a `_sess_pid(sess)` helper, or coercing in `_load_session` — so every site (read and signal) sees an int. Optional.

**F2 — optional `pid <= 0` hardening.** After `pid = int(pid)`, a `if pid <= 0: return False` would make corrupt `"0"`/`"-1"` records report not-alive rather than a bogus "alive" (`os.kill(0, 0)` / `os.kill(-1, 0)` succeed as existence checks and currently return True). Cosmetic — the status readout, not safety.

### Bottom line
Correctly and safely fixes the stated `list`/`status` crash; no caller depended on the old exception; the sig=0 contract rules out any group-signal risk and real data cannot coerce to 0/negative. **APPROVE.** F1 (extend the int-normalization to the signal sites, ideally at the read boundary) is the one worth a quick follow-up.

Ref: POSIX.1-2017 kill(2) (null-signal existence check; pid ≤ 0 process-group targeting).

PR #719, opened 2026-07-14
mm: don't write-protect the kernel identity map on kernel_cr3 fork/mprotect (task#4 WP guard)

me1iissa:
## Summary

A process running on the kernel's own page tables (`actual_cr3 == kernel_cr3`) is only ever the in-kernel test runner: a `vm_space == None` process runs on `kernel_cr3`, and its lazily-created VmSpace (`sys_mmap`'s `from_existing_cr3` path) adopts `kernel_cr3`. For such a caller, two mm paths would retag the kernel's 1:1 identity map (PML4[0]):

- **`VmSpace::clone_for_fork`** write-protects every present leaf of the parent for copy-on-write. On `kernel_cr3` those leaves ARE the identity map, so the pass clears `PAGE_WRITABLE` on the kernel's own mappings. The running kernel writes through the identity map (e.g. the BSP bootstrap stack); the next store to a now-read-only identity page faults, and a `#PF` that re-pushes onto the same read-only stack escalates to `#DF` (a `getrandom`/`sysinfo` kernel-context write to such a page instead fault-loops).
- **`sys_mprotect`** retags PTEs in place and would clear `PAGE_WRITABLE` on an identity page the same way.

## The guard (identity-range-scoped, not cr3-wholesale)

A shared discriminator `mm::vmm::is_identity_map_phys(va, phys)` — a present leaf whose masked physical address equals its virtual address, inside the low 1:1 window (`< 4 GiB`).

- **fork** (`clone_for_fork`, all three WP sites — 4 KiB / 2 MiB / 1 GiB): when forking on `kernel_cr3`, identity leaves are left writable and copied verbatim instead of write-protected; every other (non-identity, high-VA) leaf is CoW'd exactly as before.
- **mprotect** (`sys_mprotect`): a range that covers an identity page under `kernel_cr3` is refused with `-EINVAL` before any VMA-list mutation.

**Refcount accounting:** identity leaves take **no reference** — the child holds a writable alias to a kernel-owned frame it must never free, so `page_ref_inc` is skipped for them, mirroring the write-protect skip. This is required for balance: teardown decrements per VMA (`free_process_memory` walks `vm_space.areas`; `free_user_page_tables` frees only structure pages and skips huge leaves), and the identity map is covered by no VMA, so a kept increment would never be decremented (~1M leaked increments per `clone_for_fork(kernel_cr3)`). Skipping the increment keeps inc and dec both absent = balanced. (Corrected from the initial revision, which kept the increment on a false "accounting unchanged" premise — see the F1 fix below.)

**Why identity-scoped:** a normal user address space never maps `phys == va` (its pages are backed by unrelated frames) and never satisfies `actual_cr3 == kernel_cr3`, so this is inert for every real `fork(2)`/`mprotect(2)`. An in-kernel mprotect test that maps a HIGH user VA into `kernel_cr3` and mprotects it still succeeds because that page is not identity (Test 49). A naive `cr3 == kernel_cr3` wholesale refuse would break that test.

## Test

**Test 49b** exercises the discriminator (6 cases), the mprotect guard end-to-end (identity page → `EINVAL`, its PTE stays writable, a HIGH-VA page on the same `kernel_cr3` still retags to `0`), and — after the F1 fix — the **fork guard directly**: it drives a real `clone_for_fork(kernel_cr3)` (after unmapping the anon page so the walk is identity-only) and asserts a known identity frame's `page_ref_count` is unchanged across the fork AND the child's teardown. That assertion fails if the increment ever returns.

## Review round 2 (F1–F4)

- **F1 (was blocking): fixed the identity-frame refcount leak** — skip `page_ref_inc` for identity leaves (2 MiB sub-loop + 4 KiB), and corrected the false balance comment.
- **F2**: added the direct `page_ref_count`-across-fork+teardown assertion above.
- **F3**: the mprotect huge-leaf pre-check now tests both 1 GiB and 2 MiB alignment (`read_pte` does not report the level), so a 1 GiB identity leaf cannot be mis-masked.
- **F4**: documented that `IDENTITY_WINDOW_TOP` is coupled to the bootloader's 4 GiB identity extent and must move in lockstep.

## Verification

- **SMP=1, TCG (`--no-kvm`, matches CI):** Tests 49 / 49b / 297 pass; Test 49b reports the identity frame's refcount balanced (0) across fork+teardown. No new deterministic failures — the pre-existing environmental/allow-listed set only (TCC/busybox/glibc-interp missing from the local disk image; `test236` fragmentation; `test_522` allow-listed). (`kuser_shared` is a pre-existing RTC-read-race flake — its SystemTime derives from the 1-second-resolution RTC, `nt::kuser_shared::update_time_fields` → `rtc::read_unix_time`; it passed on re-run and in every prior run, and this change touches no time path.)
- **SMP=2 test-mode flake A/B:** reported separately (the #714 fault-rate metric) — 0 faults on both arms; the SMP=2 gate is the parked W101 pipe-test wedge, not faults. This PR is independent of that CI-containment decision.

Refs: `fork(2)`, `mprotect(2)`, Intel SDM Vol. 3A §4.10.5 (paging-structure change propagation), §4.6 (access rights).

### DIFF
--- a/kernel/src/mm/vma.rs
+++ b/kernel/src/mm/vma.rs
@@ -796,7 +796,8 @@ impl VmSpace {
     /// Also syncs `self.cr3 = actual_cr3` so subsequent VmSpace operations
     /// (demand-paging, CoW handling) use the correct page tables.
     pub fn clone_for_fork(&mut self, actual_cr3: u64) -> Option<Self> {
-        use crate::mm::vmm::{PAGE_PRESENT, PAGE_WRITABLE, PAGE_HUGE, ADDR_MASK};
+        use crate::mm::vmm::{PAGE_PRESENT, PAGE_WRITABLE, PAGE_HUGE, ADDR_MASK,
+                             get_kernel_cr3, is_identity_map_phys};
         use crate::mm::pmm;
         use crate::mm::refcount::page_ref_inc;
 
@@ -840,6 +841,23 @@ impl VmSpace {
             );
             self.cr3 = actual_cr3;
         }
+
+        // task#4 identity-map guard.  A fork whose page tables ARE the kernel's
+        // own (`actual_cr3 == kernel_cr3`) is only ever the in-kernel test
+        // runner (a `vm_space == None` process runs on `kernel_cr3`).  For such
+        // a fork the CoW write-protect pass below would clear `PAGE_WRITABLE`
+        // on the kernel's 1:1 identity map (PML4[0]); the running kernel writes
+        // through that map (e.g. the BSP bootstrap stack), so the next kernel
+        // store to a now-read-only identity page faults — a #PF that re-pushes
+        // onto the same read-only stack and escalates to #DF, or a getrandom /
+        // sysinfo kernel-context write that fault-loops.  When this flag is set
+        // the leaf loops below leave identity leaves (`phys == va`, low VA)
+        // writable and CoW every other (non-identity, high-VA) leaf normally.
+        // A real user address space never satisfies `actual_cr3 == kernel_cr3`
+        // — its PML4[0] is built empty and never shares a page-table page with
+        // the kernel identity map — so this is inert for every real fork(2).
+        let kernel_cr3 = get_kernel_cr3();
+        let fork_on_kernel_cr3 = kernel_cr3 != 0 && actual_cr3 == kernel_cr3;
 
         // Per-fork [FORK-COW] dump (START line + one line per VMA, ~100 VMAs
         // per Firefox process).  ALWAYS-ON historically — this taxed every
@@ -904,8 +922,16 @@ impl VmSpace {
 
                     // 1 GB huge page — write-protect in both, no CoW split.
                     if pdpte & PAGE_HUGE != 0 {
-                        let flags_ro = (pdpte & !ADDR_MASK) & !PAGE_WRITABLE;
                         let phys_1g  = pdpte & !0x3FFF_FFFFu64;
+                        let va_1g    = ((pml4_idx as u64) << 39) | ((pdpt_idx as u64) << 30);
+                        if fork_on_kernel_cr3 && is_identity_map_phys(va_1g, phys_1g) {
+                            // Kernel identity 1 GiB page: keep it writable in
+                            // both parent and child (copy the PDPTE verbatim).
+                            *parent_pdpt.add(pdpt_idx) = pdpte;
+                            *child_pdpt .add(pdpt_idx) = pdpte;
+                            continue;
+                        }
+                        let flags_ro = (pdpte & !ADDR_MASK) & !PAGE_WRITABLE;
                         *parent_pdpt.add(pdpt_idx) = phys_1g | flags_ro;
                         *child_pdpt .add(pdpt_idx) = phys_1g | flags_ro;
                         continue;
@@ -928,9 +954,23 @@ impl VmSpace {
                         // 2 MB huge page — write-protect in both and ref-count sub-pages.
                         if pde & PAGE_HUGE != 0 {
                             let phys_2m     = pde & 0x000F_FFFF_FFE0_0000u64;
-                            let flags_ro    = (pde & !ADDR_MASK) & !PAGE_WRITABLE;
-                            *parent_pd.add(pd_idx) = phys_2m | flags_ro;
-                            *child_pd .add(pd_idx) = phys_2m | flags_ro;
+                            let va_2m       = ((pml4_idx as u64) << 39)
+                                            | ((pdpt_idx as u64) << 30)
+                                            | ((pd_idx as u64) << 21);
+                            let (parent_pde, child_pde) =
+                                if fork_on_kernel_cr3 && is_identity_map_phys(va_2m, phys_2m) {
+                                    // Kernel identity 2 MiB page: keep writable
+                                    // in both (copy the PDE verbatim).  The
+                                    // sub-page ref-count accounting is unchanged
+                                    // from the normal path so teardown stays
+                                    // balanced — only the writable bit differs.
+                                    (pde, pde)
+                                } else {
+                                    let flags_ro = (pde & !ADDR_MASK) & !PAGE_WRITABLE;
+                                    (phys_2m | flags_ro, phys_2m | flags_ro)
+                                };
+                            *parent_pd.add(pd_idx) = parent_pde;
+                            *child_pd .add(pd_idx) = child_pde;
                             for sub in 0..512u64 {
                                 page_ref_inc(phys_2m + sub * 0x1000);
                             }
@@ -952,13 +992,24 @@ impl VmSpace {
                             if pte & PAGE_PRESENT == 0 { continue; }
 
                             let phys       = pte & ADDR_MASK;
-                            let flags_ro   = (pte & !ADDR_MASK) & !PAGE_WRITABLE;
+                            let va         = ((pml4_idx as u64) << 39)
+                                           | ((pdpt_idx as u64) << 30)
+                                           | ((pd_idx as u64) << 21)
+                                           | ((pt_idx as u64) << 12);
 
-                            // Write-protect parent PTE in place.
-                            *parent_pt.add(pt_idx) = phys | flags_ro;
+                            // Kernel identity leaf under kernel_cr3: keep the
+                            // parent (and child) PTE writable — never CoW the
+                            // kernel's own 1:1 map.  Otherwise write-protect in
+                            // place as usual.
+                            let entry = if fork_on_kernel_cr3 && is_identity_map_phys(va, phys) {
+                                pte
+                            } else {
+                                phys | ((pte & !ADDR_MASK) & !PAGE_WRITABLE)
+                            };
 
-                            // Child PTE: same physical page, read-only.
-                            *child_pt.add(pt_idx) = phys | flags_ro;
+                            // Parent PTE in place; child PTE: same physical page.
+                            *parent_pt.add(pt_idx) = entry;
+                            *child_pt.add(pt_idx)  = entry;
 
                             // Keep page alive until both mappings are gone.
                             page_ref_inc(phys);
--- a/kernel/src/mm/vma.rs
+++ b/kernel/src/mm/vma.rs
@@ -957,22 +957,32 @@ impl VmSpace {
                             let va_2m       = ((pml4_idx as u64) << 39)
                                             | ((pdpt_idx as u64) << 30)
                                             | ((pd_idx as u64) << 21);
-                            let (parent_pde, child_pde) =
-                                if fork_on_kernel_cr3 && is_identity_map_phys(va_2m, phys_2m) {
-                                    // Kernel identity 2 MiB page: keep writable
-                                    // in both (copy the PDE verbatim).  The
-                                    // sub-page ref-count accounting is unchanged
-                                    // from the normal path so teardown stays
-                                    // balanced — only the writable bit differs.
-                                    (pde, pde)
-                                } else {
-                                    let flags_ro = (pde & !ADDR_MASK) & !PAGE_WRITABLE;
-                                    (phys_2m | flags_ro, phys_2m | flags_ro)
-                                };
+                            let identity =
+                                fork_on_kernel_cr3 && is_identity_map_phys(va_2m, phys_2m);
+                            let (parent_pde, child_pde) = if identity {
+                                // Kernel identity 2 MiB page: keep it writable in
+                                // both (copy the PDE verbatim) AND take no
+                                // reference on its sub-pages.  The child holds a
+                                // writable alias to a kernel-owned frame it must
+                                // never be able to free, so it owns no reference;
+                                // the sub-page page_ref_inc below is skipped,
+                                // mirroring the write-protect skip.  Identity
+                                // leaves are covered by no VMA, so teardown
+                                // (free_process_memory walks VMAs;
+                                // free_user_page_tables skips huge leaves)
+                                // decrements none of them — skipping the inc
+                                // keeps inc and dec both absent = balanced.
+                                (pde, pde)
+                            } else {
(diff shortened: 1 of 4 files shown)


me1iissa:
## VERDICT: REQUEST_CHANGES

Independent, unseeded, code-level review (no QEMU; I did not read the other review first). The WP guard's core purpose is achieved correctly and safely — it cannot fire for a real user address space, it is identity-range-scoped rather than cr3-wholesale, and the mprotect side is well-designed and end-to-end tested. But the **fork path has a real refcount leak on the kept-writable identity leaves**: they are still `page_ref_inc`'d while the VMA-based teardown can never decrement them. That directly contradicts the PR's own "teardown stays balanced" claim, and it fires live on every Test 297 run. One substantive change requested; details and a minimal fix below.

### Blocking finding

**F1 — identity leaves are `page_ref_inc`'d on the fork but never decremented on teardown → a refcount leak on every `clone_for_fork(kernel_cr3)`.**

In `kernel/src/mm/vma.rs` the guard keeps identity leaves writable but leaves the refcount accounting on the *unconditional* normal-path arm:

- 2 MiB path (`vma.rs:961-975`): the guard yields `(pde, pde)` (writable, correct), then the `for sub in 0..512 { page_ref_inc(phys_2m + sub*0x1000) }` at `:974-975` runs **unconditionally** — 512 increments per identity huge page.
- 4 KiB path (`vma.rs:1004-1015`): `entry = pte` (writable, correct), then `page_ref_inc(phys)` at `:1015` runs **unconditionally**.

The comment at the 2 MiB site asserts "the sub-page ref-count accounting is unchanged from the normal path so teardown stays balanced." The premise is false for the identity case. Teardown balance on the normal path relies on *every incremented leaf being covered by a VMA*, because `free_process_memory` decrements per **VMA**, not per raw PTE:

- `proc/mod.rs:2495-2533` walks `vm_space.areas` and decrements `[vma.base, vma.base+vma.length)`.
- `free_user_page_tables` (`proc/mod.rs:2680-2705`) frees only page-table **structure** pages and explicitly `continue`s past huge leaves (`:2690`, `:2696`) — it never decrements leaf data frames.

But the identity map is **not a VMA**: the runner's VmSpace is built by `from_existing_cr3` with `areas: Vec::new()` (`vma.rs:690`), and the child clones that VMA list (`vma.rs:1042-1053`) — neither covers `[0, 4 GiB)`. So the ~1M identity-frame increments performed by the fork's PML4[0] walk have **no matching decrement anywhere in teardown**.

This is live, not hypothetical. Test 297 calls `fork_process(pid=current_pid())` (`test_runner.rs:49737`) on the in-kernel runner, which by that point holds a lazily-created VmSpace on `kernel_cr3` (Test 49b asserts `space_cr3 == kcr3`). That drives `clone_for_fork(kernel_cr3)` → the PML4[0] identity walk (four 1 GiB PDPT slots × 512 × 512 sub-pages ≈ 1,048,576 `page_ref_inc`) → `exit_group_pid(child)` → VMA-based teardown that decrements none of them. Every suite run leaks ~1M refcount increments.

Why it matters even though it's test-only and non-crashing:
- **It contradicts spec item (d)** ("not disturb refcount balance … on teardown") and the code comment.
- For any frame in `[1 MiB, 4 GiB)` that is PMM-managed and later recycled as a user page, the leaked `+1` is a **persistent refcount offset**: when its legitimate references later drop, `page_ref_dec` never reaches 0, so `pmm::free_page` is never called and the frame is permanently pinned. (Direction is over-reference only — no premature free / UAF, so no crash — but it is a genuine memory + invariant leak.)
- Under `firefox-test-core`, each `page_ref_inc` fires a `w215_diag::prov_record(KIND_REFINC)` (`refcount.rs:133-136`), so one kernel_cr3 fork spams ~1M REFINC provenance records and inflates refcounts on exactly the low-4 GiB frames the active W215 investigation inspects — it pollutes the surface being used to hunt the very corruption class this guard sits near.

Why task#4 owns this even though the `page_ref_inc` predates it: pre-guard, `clone_for_fork(kernel_cr3)` write-protected the identity map and the kernel faulted before/at completion, so the child was never cleanly torn down and the imbalance never manifested. The guard makes the fork+teardown complete, which is what turns the pre-existing unconditional increment into a live leak.

**Minimal fix:** skip the increment for identity leaves, mirroring the WP skip — the child holds a writable alias to a kernel-owned frame it must never be able to free, so it should hold no reference on it. Both directions then stay absent (no inc, no dec) and balance is restored:

- 2 MiB: gate the `for sub … page_ref_inc` loop on `!(fork_on_kernel_cr3 && is_identity_map_phys(va_2m, phys_2m))` (reuse the value already computed for the PDE decision).
- 4 KiB: only `page_ref_inc(phys)` when the leaf was not the identity branch.

This is safe against the CoW fault path because identity leaves are kept writable and therefore never take a write-protection fault, so they need no share-count. If you instead judge the leak acceptable as bounded test-only behaviour, at minimum correct the "teardown stays balanced" comment and document the intentional non-accounting — but the skip is only a few lines and removes the defect outright.

### Non-blocking findings

**F2 (LOW) — the fork guard is not exercised by Test 49b.** Test 49b's own docstring (`test_runner.rs:14718`) states a live `clone_for_fork(kernel_cr3)` is "deliberately NOT exercised here." 49b tests the predicate (6 cases) and the *mprotect* guard end-to-end (EINVAL + PTE-stays-writable + high-VA-still-succeeds) and fails loudly if it cannot reach `kernel_cr3` — that part is solid and non-vacuous. Fork-side coverage is only *implicit* via Test 297 remaining green, and green there proves "no #DF," not "refcounts balanced" (see F1). A direct fork-side assertion (e.g. snapshot a target identity frame's `page_ref_count` across a driven kernel_cr3 fork+child-teardown and assert it returns to baseline) would have caught F1 and would make the fork guard non-vacuously tested.

**F3 (LOW, defensive) — the mprotect guard mis-masks a 1 GiB huge leaf.** `syscall.rs` (the identity pre-check) branches only on `PAGE_HUGE` and applies the **2 MiB** masks `(pte & 0x000F_FFFF_FFE0_0000, p & !0x1F_FFFF)` to any huge PTE. `read_pte` returns a 1 GiB leaf as a huge PDPTE, which this would mask wrong. Inert today (the identity map is 2 MiB pages, so no 1 GiB identity leaf is reachable), but it is a latent gap if the bootloader ever emits 1 GiB identity pages.

**F4 (LOW) — `IDENTITY_WINDOW_TOP = 4 GiB` hard-couples to the bootloader's identity extent.** If the bootloader ever identity-maps beyond 4 GiB, `is_identity_map_phys` silently under-covers (those leaves would be WP'd/retagged). The tests only probe up to 0x1000_0000, so a >4 GiB regression wouldn't be caught. A comment cross-referencing the single source of truth (or deriving the bound rather than hard-coding it) would harden this.

**F5 (INFO, benign) — the kernel_cr3 fork child gets a writable alias to all of low physical memory.** Keeping the identity PDEs writable in the child means the forked child maps `[0, 4 GiB)` phys writable through its own tables. Test-only (only kernel_cr3 forks produce such a child) and the Test 297 child is never scheduled before `exit_group_pid`, so benign — noting it because it is an isolation property a future reader should know is intentional.

### Verified safe (adversarial angles that held)

- **Can never fire for a real user AS.** Doubly gated: `cr3 == kernel_cr3` (only the runner) *and* `phys == va < 4 GiB`. A real user PML4[0] is built **empty** — `new_user` deliberately does not copy the identity map (`vma.rs:739-751`, with a NULL-deref rationale; the older doc-comment above `new_user` claiming a deep-clone is stale and contradicts the code) — so a real fork walks only user VMA leaves (high VA, `phys != va`, VMA-covered) and CoW is never silently disabled. Even a coincidental low `phys == va` user leaf is saved by the `cr3` gate.
- **mprotect EINVAL rejects nothing legitimate.** The guard is skipped entirely unless `cr3 == kernel_cr3`; the runner's own high-VA mmap (`phys != va`) is allowed, which Test 49b sub-check (c) verifies. EINVAL for an unsupported range is consistent with mprotect(2).
- **No TOCTOU.** The mprotect pre-check runs before any VMA mutation and reads a static, single-threaded (kernel_cr3 runner) identity map; there is no window in which an identity leaf's classification can change between check and mutation.
- **Predicate is correct across the fork page sizes for the actual 2 MiB identity map** (1 GiB / 2 MiB / 4 KiB VA reconstruction all correct), and refcount **direction** is over-reference only, so F1 cannot produce a use-after-free.

Fixing F1 (or, at minimum, correcting the balance comment and documenting the intentional leak) is the one thing gating approval; the rest is sound.

Refs: POSIX.1-2017 fork(2) & mprotect(2); Intel SDM Vol. 3A §4.10.5 (paging-structure change propagation before frame reuse).

me1iissa:
## RE-VERDICT: APPROVE

Re-review of delta `222309f4` against my prior F1-F4. All four are addressed; the blocking finding (F1) is fixed cleanly, and the new fork-balance test is non-vacuous and safe. Verdict flips to **APPROVE**.

### F1 — refcount leak — FIXED, predicates provably congruent

The fix computes the identity classification **once** and reuses that single binding for both decisions, so the WP-skip and the inc-skip cannot diverge:

- 2 MiB (`vma.rs:960-987`): `let identity = fork_on_kernel_cr3 && is_identity_map_phys(va_2m, phys_2m);` then `if identity { (pde, pde) } else { … }` for the PDE **and** `if !identity { for sub … page_ref_inc(…) }` for the accounting — same `identity`.
- 4 KiB (`vma.rs:1011-1034`): `let identity = fork_on_kernel_cr3 && is_identity_map_phys(va, phys);` drives both `entry` and `if !identity { page_ref_inc(phys); }`.

Because both arms read the identical `identity` value, a divergence is impossible by construction — this is the strongest possible form of "exactly congruent." Balance is now correct in every case: identity leaf → no WP, no inc; teardown decrements it via neither the VMA walk (no covering VMA) nor `free_user_page_tables` (skips huge leaves) → inc and dec both absent = balanced. Non-identity leaf → WP + inc, VMA-covered dec at teardown = balanced. The 1 GiB arm is unchanged (it never inc'd). The false "accounting unchanged so balanced" comment is corrected to state the skip-both rationale.

### F2 — fork-side test coverage — now real, non-vacuous, and leak-safe

New Test 49b sub-check (d) drives an actual `clone_for_fork(kernel_cr3)` and asserts a known identity frame's refcount is unchanged across fork **and** child teardown. I checked it can't pass vacuously and can't leave dangling state:

- **Non-vacuous.** `ident_va` is a verified-present identity leaf, so it lies inside a 2 MiB identity huge page that `clone_for_fork(kcr3)`'s PML4[0] walk necessarily visits. Pre-fix, that path's `for sub in 0..512 { page_ref_inc(phys_2m + sub*0x1000) }` increments all 512 sub-frames including the one at `phys == ident_va`, so `page_ref_count(ident_va)` would read `before+1` (`page_ref_count` indexes `pfn(phys)`, and `phys == va` for identity frames makes the VA-keyed read correct). The assertion `after_fork == before` therefore fails on the old code and passes only with the skip — it genuinely exercises the guarded path.
- **`drop(vs)` is safe even though `vs.cr3 == kernel_cr3`.** `VmSpace::Drop` (`vma.rs:1516-1557`) frees **no** page tables — it only conditionally evicts the MM_REGISTRY / MM_GEN_REGISTRY entries, and only when `Arc::strong_count == 2` (self + registry). Here the kernel_cr3 sem is also held by the runner's real VmSpace (and the registry), so `strong_count > 2` and the entry is left intact. So the handle drop neither frees kernel_cr3's tables nor perturbs the registry for later tests. The child (a distinct cr3) is torn down by `free_vm_space`, which early-returns on `cr3 == kernel_cr3` and walks the child's (empty) VMA list — decrementing no identity frame, consistent with the `after_free == before` assertion.
- **No new flakiness.** The driven fork write-protects the parent's non-identity low leaves in place — the mechanism the original docstring called "flaky" — but this is the *same* operation Test 297 already performs (`fork_process(runner)` → `clone_for_fork(kcr3)`), which the runner already survives post-#716. So 49b(d) adds a refcount assertion around an operation the suite already exercises, not a new hazard. `from_existing_cr3` uses empty `areas`, so the fork/teardown touch only page tables, not VMA-pinned inode state.

### F3 — mprotect huge-leaf coverage — 1 GiB now covered; residual is unreachable

The pre-check now OR's the 2 MiB and 1 GiB alignments (`syscall.rs`), so a 1 GiB identity leaf is refused. On the specific false-positive you asked about: the OR *can* in principle match a non-identity huge leaf whose VA and phys fall in the same 1 GiB-aligned block (`hit1` true while `hit2` false — e.g. a hypothetical 2 MiB leaf `va=0x40200000 → phys=0x40400000`). But it is **unreachable**: the guard runs only on `kernel_cr3`, whose entire low 4 GiB is the phys==va identity map — there are no non-identity huge leaves there. And even if one somehow existed, the only effect is a spurious `EINVAL` on a runner mprotect (a loud failure, never silent corruption). Acceptable as-is; a size-aware check (walk the level to know 1 GiB vs 2 MiB) would remove the theoretical imprecision if you want to harden it later.

### F4 — coupling documented

`IDENTITY_WINDOW_TOP` now carries the "single source of truth, raise in lockstep with the bootloader extent or the guard under-covers" note. Addressed.

### Scope / non-blocking notes

- The delta touches only the fork refcount arms (`vma.rs`), a doc comment (`vmm.rs`), the mprotect pre-check (`syscall.rs`), and the test — **no clock/RTC/kuser_shared/SystemTime path**. The separately-logged "SystemTime did not advance" RTC-read flake is unrelated to this PR; nothing here should be gated on it.
- F3's two-mask OR imprecision (theoretical, unreachable) and a size-aware alternative are the only residuals, and neither blocks.

The blocking finding is resolved correctly, the fix is now covered by a non-vacuous test, and the supporting findings are addressed. **APPROVE.**

Refs: POSIX.1-2017 fork(2) & mprotect(2); Intel SDM Vol. 3A §4.10.5.

PR #721, opened 2026-07-16
mm/cache, vfs: ramfs/tmpfs MAP_SHARED↔read coherence (supersedes #635; F1 SMP-UAF fix)

me1iissa:
Supersedes #635 (closed in favour of this branch). Independent review of #635 returned REQUEST_CHANGES with one blocking finding (F1) and one re-validate ask (F2); this successor branch applies the F1 fix and re-validates on current master with SMP=2.

## What this change does (unchanged from #635)

Restores ramfs/tmpfs `MAP_SHARED` <-> `read(2)` coherence in both directions. An in-memory filesystem keeps a file's bytes in two places once a page is demand-faulted: the inode's own buffer (served to `read(2)`/`write(2)`/`stat`) and the page-cache frame an `mmap(MAP_SHARED|PROT_WRITE)` PTE aliases. A store through that PTE lands only in the cache frame; the inode buffer is never updated. Two halves close the gap:

- Writeback-on-eviction (`mm/cache`): before an in-memory-FS cache frame leaves the cache (`cache::evict` and the collision eviction in `cache::insert`), flush its bytes back into the inode buffer, so a later cache-miss `read(2)` observes the mmap write.
- Cache-authoritative read (`vfs::fd_read`): for in-memory filesystems, overlay any still-resident cache page's current bytes onto the read result, so an mmap write held in a resident frame is observed by `read(2)`.

Scoped to ramfs/tmpfs via `FileSystemOps::is_in_memory()` (default false). Block-backed filesystems re-read the on-disk image (single source of truth) and are unaffected.

The mechanism is complementary to #637, not redundant: #637's `WriteThrough` makes a `MAP_SHARED` store land in the cache frame in place (co-mappers see it via a cache HIT) but never reconciles that frame with the inode buffer that `FileSystemOps::read` serves, so the reverse direction (mmap write -> subsequent `read(2)`) is still incoherent on master. Tests 641 C/E/F encode exactly that contract and fail on master today.

Per POSIX.1-2017 `mmap(2)` (MAP_SHARED writes visible to `read(2)`) and `write(2)`.

## F1 (blocking) fix — reference-less overlay read was an SMP use-after-free

The #635 overlay looked a resident page up with the ref-less `cache::lookup`, which releases the cache lock before returning a bare physical address, then copied from that frame into the **user** buffer with no reference held. On SMP a concurrent `cache::evict` / collision-`insert` / `truncate_range` on another CPU could drop the frame's last reference and free+recycle it mid-copy, so the copy would pull a recycled frame's bytes (another file's data) into the reader's buffer — a `read(2)` correctness violation and an information leak, reachable only in the SMP eviction regime the overlay exists to serve.

Fix (`vfs/mod.rs`, commit 8be691e):

- Acquire a guard reference under the cache lock with `cache::lookup_and_acquire` (increments the refcount while the cache lock is held), so the frame cannot be freed or recycled for the duration of the copy even if a sibling CPU evicts the entry.
- After the copy, release the guard with `page_ref_dec`. If a concurrent eviction dropped the cache's own reference during the copy, the guard is the last holder and the frame reaches refcount zero here; route it through `tlb::quarantine_free` — the same deferred, TLB-safe free the eviction path uses (a sibling CPU may still hold a translation). Discarding the decrement result would leak the frame in that race.
- The SAFETY comment is corrected to state the actual invariant (the acquired guard reference), not the false "cache holds a reference" claim.

Holding the cache lock across the copy is not an option: `buffer` is the user buffer, so a user-faulting copy under the cache spinlock would deadlock the page-fault path. The acquire-under-lock / copy-without-lock / put-after (freeing on zero) shape is the standard page-cache read discipline.

Refs: POSIX.1-2017 `read(2)` (returned bytes are the file's data), `mmap(2)`; Intel SDM Vol. 3A section 4.10.5 (TLB / frame-reuse ordering).

## F2 (re-validate) — eviction-writeback lock order on current master

The eviction-writeback adds an `fs.write` (MOUNTS + inode mutex) into the collision-eviction path of `cache::insert`, which runs on the #PF demand-fault path. #635's CI ran against pre-#703/#708 master. On current master the file-backed #PF path drops PROCESS_TABLE before any MOUNTS/FS work and holds no `mm_sem`; the writeback runs only after the cache lock is released (lock order cache -> MOUNTS -> inode, no nesting), and ramfs `write` touches only the inode Vec (no re-entry into the cache/fault path). The SMP=2 641E run below exercises exactly this collision-eviction writeback and clears it empirically.

## Validation (harness-only)

Rebased onto current master (the review asked for a099a254; master advanced by one unrelated commit, #720, zero overlap with the 5 FS files touched here — rebased onto the newer tip for a clean CI base).

### SMP=1 TCG (`--no-kvm`, matches CI) — full suite; Test 641 A–F all pass

```
  TEST: cross-proc MAP_SHARED memfd write-visibility (Test 641)
  A no-eviction: consumer aliases producer frame + sees write ✓
  B write-after-map: later producer write visible via existing mapping ✓
  C eviction-then-remap: producer write survived eviction via writeback-to-inode ✓
  D fallocate(0,4096): backing inode grown to 4096 ✓
  E collision-eviction: production cache::insert eviction wrote sentinel back to inode ✓
  F fd_read overlay: read(2) observed the resident frame's mmap write ✓
[TEST-JSON] {"name":"cross-proc MAP_SHARED memfd write-visibility (Test 641)","result":"pass","elapsed_ticks":2}
```

Suite: `[TEST SUITE] ✗ 5 TESTS FAILED` (440 pass / 5 fail). None of the 5 are in VFS/cache/ramfs/tmpfs/ext2; each is outside this change's subsystem:
- `TCC compile` (`Cannot read /disk/bin/tcc: NotFound`) and `busybox basic` (`/disk/bin/busybox: NotFound`) — CI-allowlisted missing-binary env-fails.
- `glibc_hello` (`create_user_process failed: InterpNotFound`) — the dynamic linker is absent from the local test image; passes on a fully-staged CI image.
- `test236` (`pmm::alloc_pages(64) failed`) — contiguous PMM allocation (memory subsystem); this change is refcount- and allocation-neutral.
- `test_522_tcp_rx_poll_bell` (`read 0 bytes, expected 18`) — TCP RX (networking), a known flaky (dedicated `w223/ci-allowlist-test522` branch).

(The three `/disk` env-fails are an artifact of the local test image, not a regression — see note below.)

### SMP=2 (2 CPUs online) — Test 641 C and E (the eviction paths where both the coherence gap and the F1 race are reachable)

```
[SMP] AP 1 online (after SIPI #1), 2 CPU(s) online
  C eviction-then-remap: producer write survived eviction via writeback-to-inode ✓
  E collision-eviction: production cache::insert eviction wrote sentinel back to inode ✓
  F fd_read overlay: read(2) observed the resident frame's mmap write ✓
[TEST-JSON] {"name":"cross-proc MAP_SHARED memfd write-visibility (Test 641)","result":"pass","elapsed_ticks":2}
```

Test 641 passes in full (A–F) with two CPUs online. The F1 overlay path (case F) and both eviction-writeback paths (C, E) run clean under SMP with no UAF, deadlock, or `[REFCOUNT/DEC-UNDERFLOW]`.

### F2 empirical clearance

Case E on the SMP=2 run is exactly the production collision-eviction writeback: it faults a frame in, stores the sentinel into it, then `cache::insert`s a colliding key so the collision eviction fires and writes the evicted frame's bytes back into the inode buffer via `fs.write` (MOUNTS + inode mutex) on the #PF-adjacent path. It passed with no deadlock and no refcount underflow — the lock-order interaction the review asked to re-validate on current master (post-#703/#708/#719) is clear.

Note on the local image: the SMP runs used a private copy of `build/data.img`. During worktree setup the harness auto-regenerated the data image (flagged separately); the regenerated image is missing a few staged `/disk` binaries (`tcc`, `busybox`, glibc interpreter), which accounts for the three `/disk` env-fails above. These are image-provisioning artifacts, not code regressions — the FS coherence gate (Test 641) and every other VFS/cache/ext2/ramfs/tmpfs test pass.

## Metadata hygiene

Both original #635 commits carried a real-email author and a session-URL trailer; both are corrected on this branch (author/committer normalized to the noreply identity, session trailers stripped, `Co-Authored-By` preserved).

Original reviewer to re-verdict before merge.

### DIFF
--- a/kernel/src/mm/cache.rs
+++ b/kernel/src/mm/cache.rs
@@ -98,6 +98,91 @@ pub fn lookup(mount_idx: usize, inode: u64, page_offset: u64) -> Option<u64> {
         .map(|e| e.phys)
 }
 
+/// PHYS → kernel-higher-half direct-map offset.  Same constant as
+/// `mm/pmm.rs`; the page cache reads/writes cached frames through this map.
+/// Per Intel SDM Vol. 3A §4.10.5 the higher-half identity map covers every
+/// PMM frame, so `PHYS_OFFSET + phys` is always a valid kernel VA for an
+/// allocated frame.
+const PHYS_OFFSET: u64 = 0xFFFF_8000_0000_0000;
+
+/// Write a cache frame back to its in-memory filesystem's inode buffer.
+///
+/// ## Why this exists — ramfs/tmpfs MAP_SHARED dual-storage coherence
+///
+/// An in-memory filesystem (ramfs / tmpfs) keeps a file's bytes in TWO
+/// places once a page is demand-faulted:
+///
+///   1. the inode's own buffer (`RamInode::File::data`), which is what
+///      `FileSystemOps::read` / `write` / `stat.size` operate on; and
+///   2. the page-cache frame the demand-fault path installs, which a
+///      `mmap(MAP_SHARED)` PTE aliases directly.
+///
+/// A `MAP_SHARED + PROT_WRITE` store lands ONLY in the cache frame — the CPU
+/// writes the frame through the aliasing PTE; the kernel is never on that
+/// write path, so the inode buffer is never updated.  If the frame is then
+/// evicted (and freed), a subsequent `read(2)` (which goes to the inode
+/// buffer via `FileSystemOps::read`) returns the STALE, never-updated bytes
+/// — for a freshly-`ftruncate`d memfd, all zeros.  This is the
+/// dual-storage incoherence that empties Firefox's WebRender display-list /
+/// blob transport.
+///
+/// This helper restores coherence by flushing the cache frame's current
+/// bytes back into the inode buffer just before the frame leaves the cache,
+/// so the inode buffer becomes the authoritative copy of any mmap writes for
+/// any later cache-miss `read`.  For block-backed filesystems (ext2 / fat32)
+/// `is_in_memory()` is false and this is a no-op — those filesystems re-read
+/// the on-disk image and never alias a writable cache frame in this way, so
+/// they are unaffected.
+///
+/// Per POSIX.1-2017 mmap(2): a `MAP_SHARED` write "shall change the
+/// underlying file object" and be visible to a subsequent `read(2)`.
+///
+/// ## Lock order
+///
+/// This MUST be called with NO cache lock held.  It snapshots the
+/// filesystem handle via [`crate::vfs::fs_at`] (which takes and immediately
+/// drops `MOUNTS`), then dispatches `FileSystemOps::write`, which on
+/// ramfs/tmpfs takes only the inode mutex.  Lock order is therefore
+/// cache → (released) → MOUNTS → (released) → inode — no cycle, no nesting.
+/// Callers that hold the cache lock collect `(mount_idx, inode, page_offset,
+/// phys)` while locked and invoke this after releasing the lock, exactly as
+/// the quarantine-free deferral does.
+///
+/// `phys` must still be a live frame (the caller holds a reference, e.g. the
+/// cache's own reference that has not yet been dropped) so the direct-map
+/// read is valid.  Best-effort: a write error (e.g. a read-only tmpfs, which
+/// has no MAP_SHARED writers anyway) is ignored.
+fn writeback_frame_to_inode(mount_idx: usize, inode: u64, page_offset: u64, phys: u64) {
+    let fs = match crate::vfs::fs_at(mount_idx) {
+        Some((fs, _)) => fs,
+        None => return,
+    };
+    if !fs.is_in_memory() {
+        return;
+    }
+    // Snapshot the frame's 4 KiB through the higher-half direct map.  The
+    // frame is alive (caller holds a reference), so the read is valid.
+    let mut frame = [0u8; 4096];
+    let src = (PHYS_OFFSET + phys) as *const u8;
+    // SAFETY: `phys` is a live, page-aligned PMM frame; the higher-half
+    // identity map covers it (Intel SDM Vol. 3A §4.10.5).  We copy exactly
+    // one 4 KiB page into a stack buffer.
+    unsafe {
+        core::ptr::copy_nonoverlapping(src, frame.as_mut_ptr(), 4096);
+    }
+    // Write the page's bytes back into the inode buffer at this file offset,
+    // bounded to the file's current size so we do not zero-extend a sealed /
+    // shrunk file.  ramfs `write` grows the buffer as needed; we cap the
+    // write at `stat.size` so a partial last page does not artificially
+    // extend the file past its real end.
+    let file_size = fs.stat(inode).map(|s| s.size).unwrap_or(0);
+    if page_offset >= file_size {
+        return; // page is entirely past EOF — nothing authoritative to flush.
+    }
+    let span = core::cmp::min(4096u64, file_size - page_offset) as usize;
+    let _ = fs.write(inode, page_offset, &frame[..span]);
+}
+
 /// Look up a cached page and atomically acquire a guard reference on it.
 ///
 /// The reference count is incremented while the cache lock is still held, so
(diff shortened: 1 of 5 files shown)


me1iissa:
## RE-VERDICT: APPROVE

Fix-cycle re-review of #721 (successor to #635) against my own findings, with the hardest look at the one deviation from my prescription — the `page_ref_dec == 0 → tlb::quarantine_free` on the release. The deviation is **correct, and better than my sketch**; F1 is properly closed; metadata is clean; F2 holds. Approving.

### F1 — SMP use-after-free — FIXED, and the quarantine deviation is right

The overlay now takes a guard reference under the cache lock (`cache::lookup_and_acquire`), copies, then releases with `page_ref_dec` and — on a zero result — `tlb::quarantine_free`. I verified the whole refcount interaction on current master, not just the shape:

- **The guard closes the window.** `lookup_and_acquire` increments the refcount *while holding the cache lock* (its own doc: "prevents any concurrent `cache::insert` … from driving the count to zero"). So for the entire copy the frame's rc ≥ 1 (our guard). No concurrent evictor can observe rc == 0, therefore none can free/recycle the frame mid-copy — the original UAF is gone. The SAFETY comment is corrected to name the guard, not the false "cache holds a reference."

- **My illustrative `let _ = page_ref_dec` WAS a real leak — the fixer is right.** If a concurrent `cache::evict_if_phys` / collision-`insert` / `truncate_range` drops the cache's own reference during the copy, the guard becomes the last holder and *our* dec is the 1→0 transition. Discarding that result (my sketch) leaves a rc-0 frame that is never returned to the PMM — a leak in exactly the race the overlay exists to serve. Routing the zero result through `quarantine_free` fixes it. Good catch; the sketch was illustrative and incomplete, and the implementation is the complete form.

**Answering the four adversarial questions directly:**

1. **Double-free / can the eviction path also free it?** No. The refcount transition 1→0 happens exactly once (`page_ref_dec`'s CAS), and while our guard is held no other path can reach 0. Cross-checking every concurrent path on master: `insert_with_expected`'s collision eviction frees (quarantine) *only* when its own `new_rc == 0` (cache.rs — conditional), which it cannot reach while our +1 is live; `evict_if_phys` (the production evictor) and `truncate_range` only `page_ref_dec` the cache's ref and never free; the standalone `cache::evict` whose caller frees unconditionally is **test-only** and single-threaded. So the unique 0-observer is our post-copy dec, and only it frees. No double-free.

2. **Does `quarantine_free` tolerate/expect this caller?** Yes. It is the generic deferred-free-vs-stale-TLB primitive that `insert_with_expected` and `unmap_and_free_range_in` already use. A frame whose rc just hit 0 while a sibling CPU may still hold a translation (the mmap'd-then-torn-down case) is exactly its contract — a bare `pmm::free_page` would be the wrong choice here, so the deviation is not just leak-safe but TLB-correct (Intel SDM 3A §4.10.5).

3. **Re-insert between evict and dec → wrong free?** No, handled by the same dec-on-0 test. If the *same* phys is re-inserted on that key, `cache::insert` re-increments it under the lock, so our later dec returns ≥ 1 and we do **not** free (correct — the cache references it again). If a *different* frame is inserted, our old phys has only our guard, our dec returns 0, and freeing it is correct (it is genuinely unreferenced). Either branch is right.

4. **Is the leak in my sketch real?** Yes (see above). Confirmed.

No premature-free either: `dec == 0` means no cache ref *and* no live mmap PTE, so a still-mapped frame (rc ≥ 1 via its PTE) is never freed by the overlay.

### Metadata — clean

All three commits (`8be691e9`, `6febbfee`, `718bc5ad`) are authored **and** committed as `184648288+me1iissa@users.noreply.github.com`; the Opus co-author trailer uses `noreply@anthropic.com`; no `Claude-Session` trailer and no real-email/session-URL anywhere in the three bodies. The June-era leak (real email + session trailer on the original #635 commits) is not carried into r2.

### F2 — cleared (static + empirical)

The collision-eviction writeback (`insert_with_expected` → `writeback_frame_to_inode` → `fs.write`) runs lock-free w.r.t. the cache lock, and current master's file-backed #PF path drops PROCESS_TABLE before any MOUNTS work and holds no mm_sem, so there is no new #703/#708 cycle. The reported SMP=2 Test 641 **case E** pass (production collision-eviction, no deadlock, no DEC-UNDERFLOW) on post-#703/#708/#719 master empirically confirms it.

### On the validation evidence — code-correctness carries F1; a contention run is optional

Your read is right that the SMP=2 641 A–F pass is **coherence + no-regression**, not a contention exercise: cases C/E/F drive the eviction/overlay wires single-threaded on the runner, so they never place a concurrent evictor inside the overlay's copy window — the F1 race is not reproduced by them. That does not block approval, because F1 is correct **by construction**: the guard reference makes rc ≥ 1 across the copy (so no UAF), the CAS makes 1→0 unique (so no double-free), and dec-on-0→quarantine frees exactly once (so no leak). This is the same acquire/copy/release discipline the demand-fault path already uses, and DEC-UNDERFLOW stayed 0 in the SMP=2 run (which would trip on a double-dec). A deterministic contention test (force an evict of the target key mid-overlay) would be worthwhile hardening but is hard to make non-flaky and is not required to merge.

### Verdict

F1 is fixed correctly — the quarantine deviation is the right call and closes a real leak my sketch would have left. Double-free, re-insert, and premature-free are all excluded by the guard + dec-on-0 discipline; metadata is clean; F2 holds statically and in the SMP=2 case-E run. **APPROVE** (merge on CI green).

Refs: POSIX.1-2017 read(2)/mmap(2); Intel SDM Vol. 3A §4.10.5.

PR #722, opened 2026-07-16
harness: make data.img regen safe (atomic rename + refuse-while-open + no write-through-symlink)

me1iissa:
## What & why

Incident 2026-07-16: the harness staleness check auto-ran `create-data-disk.sh --force` from a worktree whose `build/data.img` was a **symlink** to the canonical `/home/ubuntu/AstryxOS/build/data.img`. The regen rewrote the canonical image **in place, through the symlink**, while two running QEMU sessions held it open as `snapshot=on` backing — the live guests then risked reading a mixed-layout ext2, poisoning a memory-corruption investigation.

## Fixes (all additive)

`scripts/create-data-disk.sh`:
1. **Atomic replace** — the image is built into a temp file in the **same directory** as the resolved target and `rename(2)`'d over it. A reader holding the old inode via an open fd keeps reading a consistent old image; only new opens see the new one. For a symlinked target the rename lands on the resolved real file, preserving the symlink.
2. **Refuse-while-open** — the resolved target is checked for open holders via `fuser(1)` (`lsof(8)` fallback); any holder matching a running harness session (`~/.astryx-harness/*.json` pid) is named with its sid. Held-open regen is refused (**exit 3**) unless `--force-inuse`. The refusal stays even though (1) makes it safe for existing readers, because silently swapping the image under an active investigation invalidates *future* boots' comparability — the message points the caller at stopping the fleet or isolating with a private `--output` copy.
3. **No write-through-symlink to foreign territory** — a `build/data.img` symlink whose target resolves **outside the invoking tree** is refused (**exit 3**) with no override; the caller must use a private copy. A within-tree symlink is honored (resolved target replaced atomically, link preserved).

`scripts/qemu-harness.py` (start-path staleness auto-regen):
- Refuses up front when `build/data.img` symlinks a shared target outside the worktree (boots the untouched stale image).
- Surfaces a script-level refusal (exit 3) as **additive JSON**: `_regen_data_img` gains `refused`/`refused_reason`; `firefox_variant_info.regen_refused_reason` records it. The auto path **never** passes `--force-inuse`, so automatic regeneration always fails safe. No field renames.

## Verification (host-side only — no QEMU)

New `scripts/create-data-disk-safety-smoke.sh` drives the script in a throwaway isolated tree (no staging, ~2s) and asserts all three guards; wired into `build.yml`'s `tooling-smoke` job. Local run: **20 passed, 0 failed** —
- (a) atomic-rename: target inode changes; a background `sleep`-holder's fd still `cmp`-matches the pre-regen image; new target differs.
- (b) refuse-while-open triggers on a held fd (exit 3, names holder + override) and `--force-inuse` overrides.
- (c) foreign-symlink refuses even with `--force-inuse` (foreign inode + content untouched); within-tree symlink allowed and its resolved target replaced atomically.

Existing `qemu-harness-smoke.py --staleness-only` still passes.

Diff ~243 non-test lines (~1.6× the ~150 soft budget) — the overrun is two self-contained bash holder-detection/naming helpers plus surfacing the refusal at both harness regen sites; clean, no scope creep.

Do not merge — coordinator routes review/merge.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -238,7 +238,9 @@ jobs:
       # in-place / write-through-symlink regen that clobbered a shared data.img
       # under running QEMU sessions.
       - name: create-data-disk write-safety smoke (atomic rename + refuse guards)
-        run: bash scripts/create-data-disk-safety-smoke.sh
+        # --require-tools: a runner missing e2fsprogs/psmisc/lsof FAILS here
+        # instead of silently self-skipping and reducing coverage while green.
+        run: bash scripts/create-data-disk-safety-smoke.sh --require-tools
 
   kernel-build:
     name: bootloader + kernel builds
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -225,6 +225,20 @@ jobs:
       # render), or an injection path that corrupts the ext2 image.
       - name: patch-gui-gl smoke (software-GL image provisioning + ext2 integrity)
         run: python3 scripts/patch-gui-gl-smoke.py
+      # Boot-free smoke for the create-data-disk.sh write-safety guard.  Drives
+      # the script in a throwaway isolated tree (no staging, ~2s) and asserts
+      # the three guards from the 2026-07-16 incident: (a) regen builds to a
+      # temp file and renames(2) over the target so a process holding the old
+      # image open keeps reading it; (b) a held-open target is refused (exit 3)
+      # unless --force-inuse; (c) a data.img symlinked to a target outside the
+      # tree is refused with no override, while a within-tree symlink is honored
+      # and its resolved target replaced atomically.  Self-skips any sub-check
+      # whose tool (mke2fs / fuser|lsof) is absent so a minimal runner can't
+      # false-fail.  Failure mode this prevents: a refactor re-introducing the
+      # in-place / write-through-symlink regen that clobbered a shared data.img
+      # under running QEMU sessions.
+      - name: create-data-disk write-safety smoke (atomic rename + refuse guards)
+        run: bash scripts/create-data-disk-safety-smoke.sh
 
   kernel-build:
     name: bootloader + kernel builds
--- a/scripts/create-data-disk-safety-smoke.sh
+++ b/scripts/create-data-disk-safety-smoke.sh
@@ -1,0 +1,195 @@
+#!/usr/bin/env bash
+#
+# Host-side smoke test for the create-data-disk.sh write-safety guard
+# (incident 2026-07-16: a --force regen wrote THROUGH a worktree symlink and
+# rewrote the canonical build/data.img in place while running QEMU sessions
+# held it open as snapshot=on backing).
+#
+# One-shot, non-interactive, no QEMU: it drives create-data-disk.sh in an
+# isolated throwaway tree (only the script is present, so every install-*.sh
+# staging step is skipped) and asserts the three guards:
+#
+#   (a) atomic replace   — regen builds to a temp file and renames(2) over the
+#                          target: the target inode CHANGES, and a process that
+#                          held the OLD image open keeps reading the old bytes.
+#   (b) refuse-while-open — a held-open target is refused (exit 3) unless
+#                          --force-inuse is passed.
+#   (c) foreign symlink  — a data.img symlinked to a target outside the tree is
+#                          refused (exit 3) with no override; the foreign file
+#                          is left untouched.  A within-tree symlink is allowed
+#                          and its resolved target is replaced atomically.
+#
+# Exit 0 = all checks passed (or cleanly skipped for a missing tool); non-zero
+# = a guard regressed.  Refs: rename(2), fuser(1), lsof(8), mke2fs(8).
+set -uo pipefail
+
+SELF_DIR="$(cd "$(dirname "$0")" && pwd)"
+CDD="${SELF_DIR}/create-data-disk.sh"
+SIZE_MB=128   # >= inode-table floor for the script's hardcoded -N 200000
+
+PASS=0
+FAIL=0
+pass() { echo "  PASS: $*"; PASS=$((PASS + 1)); }
+fail() { echo "  FAIL: $*"; FAIL=$((FAIL + 1)); }
+
+if [ ! -f "${CDD}" ]; then
+    echo "SKIP: create-data-disk.sh not found at ${CDD}"
+    exit 0
+fi
+if ! command -v mke2fs >/dev/null 2>&1; then
+    echo "SKIP: mke2fs (e2fsprogs) not installed — cannot build test images"
+    exit 0
+fi
+HAVE_INUSE_TOOL=false
+if command -v fuser >/dev/null 2>&1 || command -v lsof >/dev/null 2>&1; then
+    HAVE_INUSE_TOOL=true
+fi
+
+WORK="$(mktemp -d)"
+cleanup() { rm -rf "${WORK}" 2>/dev/null || true; }
+trap cleanup EXIT
+
+# Isolated fake-root: only create-data-disk.sh present, so ROOT_DIR resolves
+# here and every `[ -f "${ROOT_DIR}/scripts/install-*.sh" ]` guard is false —
+# no glibc/Firefox staging runs and the build is ~0.3s.
+mkroot() {
+    local r="$1"
+    mkdir -p "${r}/scripts" "${r}/build/disk"
+    cp "${CDD}" "${r}/scripts/create-data-disk.sh"
+    printf 'VERSION=1\n' > "${r}/build/disk/marker.txt"
+}
+run_cdd() {  # run_cdd <root> <output-or-empty> <extra-args...>
+    local r="$1"; shift
+    local out="$1"; shift
+    local args=(--build-dir="${r}/build" --force "${SIZE_MB}")
+    [ -n "${out}" ] && args=(--output="${out}" "${args[@]}")
+    bash "${r}/scripts/create-data-disk.sh" "${args[@]}" "$@" \
+        > "${r}/last.log" 2>&1
+}
+
+echo "== create-data-disk safety smoke (work=${WORK}) =="
+
+# ── (a) atomic replace + old-reader isolation ────────────────────────────────
+echo "[a] atomic replace via temp+rename"
+A="${WORK}/a"; mkroot "${A}"; IMG="${A}/build/data.img"
+if run_cdd "${A}" "${IMG}"; then
+    if dumpe2fs -h "${IMG}" >/dev/null 2>&1; then
+        pass "v1 build produced a valid ext2 image"
+    else
+        fail "v1 image is not a valid ext2 filesystem"
+    fi
+    grep -q 'via temp+rename' "${A}/last.log" \
+        && pass "build logged the temp+rename path" \
+        || fail "build did not use temp+rename"
+    I1="$(stat -c %i "${IMG}")"
+    MODE="$(stat -c %a "${IMG}")"
+    [ "${MODE}" = "644" ] && pass "image mode is 0644" \
+        || fail "image mode is ${MODE}, expected 644"
+    cp "${IMG}" "${WORK}/ref_v1.img"
+    # Hold the OLD inode open, then regen (needs --force-inuse since it's held).
+    sleep 30 < "${IMG}" &
+    HP=$!
+    sleep 0.3
+    printf 'VERSION=2\n' > "${A}/build/disk/marker.txt"
+    run_cdd "${A}" "${IMG}" --force-inuse
+    RC=$?
+    [ "${RC}" -eq 0 ] && pass "held-open regen with --force-inuse succeeded" \
+        || fail "held-open regen with --force-inuse rc=${RC} (expected 0)"
+    I2="$(stat -c %i "${IMG}")"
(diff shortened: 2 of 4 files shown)


me1iissa:
VERDICT: APPROVE (no blockers; 3 minor robustness notes)

Adversarial code-level review of the three write-safety guards, the additive harness JSON, and the shipped smoke. Host-tooling only — no QEMU boot. I ran the smoke suite directly (20/20 on a host with fuser+lsof+mke2fs). Findings are minor and non-blocking; the incident pattern (auto-`--force` regen writing through a worktree symlink into the canonical image held open by running QEMU) is closed by defense-in-depth on both the script and harness sides.

## Fix 1 — atomic replace: correct and empirically proven

- The temp is genuinely in the **resolved** target's directory: `IMG_DIR="$(dirname -- "${IMG_FINAL}")"` then `mktemp "${IMG_DIR}/.data.img.tmp.XXXXXX"` (create-data-disk.sh:1149-1151), and `mv -f "${IMG_TMP}" "${IMG_FINAL}"` (:1182). Same directory ⇒ same filesystem ⇒ `mv` uses rename(2) atomically. The lead's "temp in /tmp would break atomicity" concern does not apply — the temp is co-located with the resolved real file, not the symlink and not /tmp.
- **No truncate-in-place before the rename**: every `dd`/`mke2fs` targets `${IMG_TMP}` (:1154/1164/1173); the only post-rename references to `${DATA_IMG}` are read-only (`patch-gui-gl.sh --verify`, :1203/1208). A reader holding the old inode is never disturbed.
- Symlinked target: the rename lands on `IMG_FINAL` (the resolved real file), so the symlink itself is preserved.
- Smoke section (a) proves this end-to-end: it holds fd 0 open on the v1 image (`sleep 30 < IMG`), regens, and `cmp`s `/proc/<pid>/fd/0` against a saved v1 copy — the old reader still reads v1 bytes while the target inode changed (create-data-disk-safety-smoke.sh:88-110). This is a real isolation test, not a hollow assertion.

## Fix 2 — refuse-while-open: correct; auto-path fails safe

- `_image_open_pids` uses `fuser` then falls back to `lsof -t`; holders are mapped to harness sids via `~/.astryx-harness/*.json` `pid` fields (create-data-disk.sh:270-306). Held-open regen exits 3 unless `--force-inuse` (:367-385).
- **The auto path genuinely cannot reach `--force-inuse`**: the string appears in qemu-harness.py only inside a comment (line 2095), never in an argv. `FORCE_INUSE` defaults false and is set only by the explicit `--force-inuse` arg (create-data-disk.sh:62, 216). Verified by grep.
- Smoke section (b) confirms exit-3 refusal + inode-untouched + `--force-inuse` override, and self-skips when neither tool is present (smoke:118-145).

## Fix 3 — no write-through-foreign-symlink: boundary computed soundly

- The tree root is derived from the **script's own location** (`ROOT_DIR="$(cd "$(dirname "$0")/.." && pwd)"`, create-data-disk.sh:48), not CWD or git-toplevel. Running a worktree's own copy roots the boundary at that worktree — exactly what catches the incident (a worktree link to `/home/ubuntu/AstryxOS/build/data.img` resolves outside the worktree root ⇒ refused).
- Spoofing/`..`/relative-symlink resistance is real: both sides fully canonicalize before comparing — bash `readlink -f` on `IMG_FINAL` and `ROOT_REAL` (:330, 346) with a trailing-slash prefix match (`"${ROOT_REAL}/"*`, :349-350) that avoids the `/home/a` vs `/home/abc` collision; the harness uses `os.path.realpath` + `os.path.commonpath` which is component-aware (qemu-harness.py:2340-2349). Refused even with `--force-inuse`.
- The harness has its own pre-check that skips the regen call entirely when data.img symlinks outside the worktree (qemu-harness.py:3869-3891), so refusal is fast and structured — belt-and-suspenders with the script's guard.
- Smoke section (c) confirms foreign-symlink refusal (override ignored), foreign inode+content byte-untouched, symlink intact, and within-tree symlink honored with atomic replacement of the resolved target (smoke:147-189).

## Additive JSON — confirmed no renames

`_regen_data_img` gains `refused`/`refused_reason` (qemu-harness.py:2089-2097, populated from the exit-3 sentinel at :2126-2136); `firefox_variant_info` gains `regen_refused_reason` (:3880, :3922). Both are pure additions — no existing field renamed. cmd_start treats a refusal distinctly from a build failure and boots the untouched stale image (:3917-3929), so auto-regen always fails safe.

## Findings (all minor / non-blocking)

1. **Foreign-target check is gated on the *leaf* being a symlink.** Both the bash guard (`if [ -L "${DATA_IMG}" ]`, create-data-disk.sh:348) and the harness (`os.path.islink(data_img)`, qemu-harness.py:2331) only fire when `build/data.img` *itself* is a symlink. If an **ancestor directory** is the symlink (e.g. `build/` linked out) and `data.img` is a regular file within it, `IMG_FINAL`/`realpath` still resolves to foreign territory but the foreign check is skipped — the atomic rename would then write through the ancestor link. Refuse-while-open (fix 2) still catches it *if* the file is held open and a holder tool is present. Same incident class, lower likelihood (the documented worktree pattern symlinks the leaf file, which is what happened). Consider comparing `IMG_FINAL` against `ROOT_REAL` unconditionally (not only when the leaf is a symlink). Hardlinks are correctly left as an accepted residual — the atomic rename replaces the directory entry, so a hardlinked data.img regen does not clobber the shared inode; the harness docstring documents this honestly (qemu-harness.py:2320-2326).
2. **Refuse-while-open fails OPEN when neither `fuser` nor `lsof` exists.** `_image_open_pids` returns empty ⇒ no holders detected ⇒ regen proceeds. This is *backstopped* — fix 1 keeps existing readers consistent (reader integrity is never lost), and fix 3 independently blocks the actual incident (foreign symlink). The only casualty on a tool-less host is future-boot comparability. Also, `fuser`/`lsof` as non-root cannot see holders owned by other users (harness sessions are same-user, so benign in practice). Worth a one-line note in the guard that holder detection is best-effort.
3. **CI smoke depends on runner-preinstalled tools.** The `tooling-smoke` job does not explicitly `apt-get install psmisc lsof e2fsprogs`; it relies on ubuntu-latest having them (the sibling patch-gui-* smokes rely on the same). The smoke self-skips absent-tool sections and exits 0, so if a future runner image dropped psmisc, guard (b)'s coverage would silently evaporate while CI stayed green. Suggest an explicit install (or a hard assert that the tools exist) in the tooling-smoke job so the safety net can't quietly lose coverage.

## Summary

The three guards are correct, layered (script + harness both refuse foreign symlinks), and fail-safe on the automatic path. The atomic rename is the load-bearing fix and is empirically verified to isolate held-open readers. Merge-ready as-is; the three notes are follow-up hardening, not merge gates. Commit hygiene is clean (noreply author identity, no session URL in the body).

Refs: rename(2), fuser(1), lsof(8), ext2 on-disk layout.

me1iissa:
Folded in the review notes (commit `0878b559`):

**1. Ancestor-directory-symlink bypass (primary)** — replaced the leaf-only symlink gate (`-L ${DATA_IMG}` / `is_symlink`) with a **lexical-vs-physical** comparison in both the bash script and the harness pre-checks: refuse when the requested path aims **inside** the tree lexically (`os.path.abspath` / a pure-bash lexical normaliser — no symlink resolution) but the fully-resolved path lands **outside**. That catches a redirect via the leaf **or any ancestor directory**, and — deliberately not a pure-unconditional `IMG_FINAL` vs `ROOT_REAL` check — still allows an explicitly-foreign `--output=/tmp/x.img` (the private-copy escape hatch the refusal messages themselves recommend). Harness gains additive `_data_img_symlink_info.write_through_outside_wt`; both the staleness and variant-mismatch pre-checks key off it. Hardlinks remain a benign residual (atomic rename replaces the dir entry, not the shared inode) — honest docstring kept.

**2. CI tool assert (cheap insurance)** — added `--require-tools` to the smoke (missing e2fsprogs/psmisc/lsof → hard FAIL instead of silent skip) and pass it in `build.yml`'s invocation.

**3. (optional)** — one-line note that refuse-while-open holder detection is best-effort (fails open without fuser/lsof; can't see other-user holders as non-root), backstopped by the atomic rename + fix 3.

**Regression-locked**: smoke now 26 checks incl. ancestor-dir-symlink refusal (foreign inode+content untouched) and an explicit-foreign `--output` allow (guards against over-refusing the escape hatch). Local 26/26 pass incl. `--require-tools`; `write_through_outside_wt` unit-verified across leaf/ancestor/explicit-foreign/within-tree; `qemu-harness-smoke.py --staleness-only` still green. CI `tooling smoke` green (32s) on 0878b559.

me1iissa:
VERDICT: APPROVE (re-review of delta 0878b559)

Re-verified all three of my prior notes against the delta, with special scrutiny on the deliberate deviation from the "unconditional IMG_FINAL vs ROOT_REAL" prescription I'd relayed. **The deviation is not just acceptable — it is the better design**, and it fully closes my primary finding. I ran the shipped smoke (26/26) and three independent probes the smoke cannot self-demonstrate. No QEMU (host-tooling only).

## Primary — ancestor-dir-symlink bypass: CLOSED, via a superior rule

The author replaced the leaf-only `-L "${DATA_IMG}"` gate with a **lexical-vs-physical** rule (create-data-disk.sh:330-393; harness qemu-harness.py:2358-2371): refuse iff the requested path aims INSIDE the tree *lexically* (abspath / pure-bash normalize, no symlink resolution) but the fully-RESOLVED path lands OUTSIDE. I worked all four adversarial cases you raised:

1. **Catches BOTH redirect shapes.** Leaf symlink: `DATA_IMG_LEXICAL` stays in-tree (no symlink follow) while `readlink -f` lands foreign ⇒ refuse. Ancestor-dir symlink (`build/` linked out, regular `data.img` within): identical outcome — lexical in-tree, physical foreign ⇒ refuse. Confirmed by code trace **and** the new smoke case (create-data-disk-safety-smoke.sh:196-215) which builds a real in-tree `linkdir` → foreign dir with a regular `data.img` inside and asserts exit 3 + foreign inode/content byte-untouched. My primary note is closed.

2. **The "lexically-outside ⇒ allowed" branch is the escape hatch, not a bypass.** It is only reachable when the caller *explicitly* names a foreign path (`--output=/tmp/x.img`, foreign `--build-dir`, or `ASTRYXOS_DATA_IMG`). The incident (auto-regen) path always uses the default in-tree `${BUILD_DIR}/data.img`, which is lexically in-tree, so **any** symlink redirect off it is refused — the auto path cannot reach the allow branch. And even an explicit foreign target that is itself a symlink to a live image is still caught by fix 2 (refuse-while-open on the resolved path) when that image is actually held open. So there is no path by which the incident recurs.

3. **The pure-bash lexical normalizer is correct** without touching the fs. I unit-tested it in isolation: `/a/b/../c→/a/c`, `/a/b/→/a/b`, `/a/./b→/a/b`, `/a/../..→/`, `/a//b///c→/a/b/c`, `/→/`, `/a/b/c/../../d→/a/d`, and `/has space/x/../y→/has space/y` (spaces preserved via `local IFS=/`). `..` correctly pops the last component (dense array, `unset` of the last index, re-append lands correctly); `..` at root is a no-op. The harness mirror uses `os.path.abspath` (lexical) vs `os.path.realpath` (physical), the exact Python equivalent.

4. **Explicit foreign `--output` still works.** Smoke (:216-226) asserts `--output=<outside tree>` returns 0 and produces a valid ext2 image. This is precisely why the author's deviation beats my prescription: an unconditional physical `IMG_FINAL vs ROOT_REAL` compare would have **wrongly refused** the private-copy escape hatch that the refusal messages themselves recommend. Conceding the point — lexical-vs-physical is the correct rule.

Both harness refusal gates (qemu-harness.py:3888, :3997) now key on `write_through_outside_wt`; the only surviving reference to the old `is_symlink AND target_outside_wt` is a docstring describing what it superseded. `target_outside_wt` is retained as an informational field but no longer gates a refusal — no residual leaf-only gap.

## CI tool-assert: hard-fails loudly (verified)

`--require-tools` routes a missing tool through `missing_tool()`, which under the flag emits `FAIL (required)` and increments `FAIL`, so the terminal `[ "${FAIL}" -eq 0 ] || exit 1` exits non-zero (create-data-disk-safety-smoke.sh:45-51). Independently verified by hiding `mke2fs` from `PATH`: `--require-tools` → `FAIL (required)` + **exit 1**; without the flag → `SKIP` + exit 0; a bad arg → usage + exit 2. build.yml:241 passes `--require-tools`, so a runner that dropped e2fsprogs/psmisc/lsof fails the job rather than going green with reduced coverage. My CI note is closed — it fails, does not warn.

## Best-effort holder-detection note: addressed

`_image_open_pids` now carries a comment stating it fails OPEN without fuser/lsof and cannot see other-user holders as non-root, and that the atomic rename (fix 1) + write-through refusal (fix 3) are the correctness backstop while this refusal is a future-boot-comparability policy guard (create-data-disk.sh:275-279). Accurate.

## Smoke: 26/26, genuine

Ran clean at 26 passed / 0 failed with `--require-tools`. The two new cases exercise real paths (actual ancestor symlink + real foreign file with inode/content assertions; real foreign `--output` producing a dumpe2fs-valid image), not tautologies.

## Bottom line

All three prior notes are resolved, the primary via a rule that is stronger than what I proposed. No new findings. Merge-ready — APPROVE on CI green (tooling smoke already green on this SHA, and I reproduced 26/26 + the `--require-tools` hard-fail locally). Rebase not required (host-tooling only, no kernel surface).

Refs: rename(2), fuser(1), lsof(8), realpath(3)/POSIX pathname resolution.

PR #723, opened 2026-07-17
sched: per-CPU runqueue load balancing — pull-on-idle + periodic push (Perf P2 phase 3d) — supersedes #649

me1iissa:
## Perf P2 #19 — phase 3d (per-CPU runqueue load balancing)

**Supersedes #649.** #649 was the top of a stacked series (phases 3a→3b→3c→3d) whose base branch is gone; its earlier phases have since landed on master independently:

- phase 3a (authoritative per-CPU pick + O(1) force gate) — merged
- phase 3b (load-aware wakeup-target selection) — merged
- phase 3c (reschedule IPI, vector 0xF2) — merged as **#709** (`7a06fe25`)
- the flaky-`vdso_map` CI allow-fail — already in `ci/allow-fail.json`

So this successor carries **only the phase-3d load-balancing change**, rebased onto current `master` and layered on top of the resched-IPI code *as-merged*. It is a pure addition (`+238` `sched/percpu.rs`, `+133` `test_runner.rs`, zero deletions).

### What it does

Evens out the per-CPU runqueues so one CPU does not pile up a backlog while another sits idle. When the busiest runqueue's depth exceeds the least-loaded one's by more than `BALANCE_IMBALANCE_THRESHOLD` (2), the most-overdue **eligible** thread is migrated from the busiest to the least-loaded runqueue. *Pull-on-idle* is the special case where the least-loaded CPU is empty.

- **No new lock surface.** Balancing runs inside `mirror_maintain`, which already holds every per-CPU runqueue guard in ascending CPU-index order (the documented migration lock order) plus the thread table. The migration dequeues the victim from the source runqueue and enqueues it on the destination under those held guards, then reassigns `last_cpu` + `mirror_slot` so the deterministic placement and the next delta pass agree.
- **Poke via the existing phase-3c mask.** The destination CPU is flagged in the same `resched_mask` the wake edge fills; the post-guard IPI loop (added by #709) drains it, so the pulled thread is picked up promptly instead of waiting a tick.
- **Throttled** to once every `BALANCE_EVERY_PASSES` (16) passes so the victim scan never dominates the hot path.
- **Migration guards:** never migrate a pinned thread (`cpu_affinity = Some(_)`), a thread mid context-switch (`!ctx_rsp_valid`), or the currently-Running thread (automatic — a Running thread carries no `Some(busiest)` mirror slot after the delta loop clears it).
- **SMP=1 is a bit-for-bit no-op** — the whole pass is gated on `ncpus > 1`.

### Rebase — conflict resolutions a reviewer should scrutinise

1. **`sched/percpu.rs` auto-merged cleanly.** The balance helpers (`pick_balance_core`, `test_pick_balance`, `pick_balance_victim`, the constants/statics) landed after `test_wake_target`; the in-`mirror_maintain` balance block landed **between** the per-rq `min_deadline` recompute and the gated audit — the identical relative position it occupied in #649 (delta loop → `min_deadline` → balance → gated audit). The block's `resched_mask |= 1 << dst` composes with the #709 wake-edge mask and drains through the same post-guard IPI loop. The percpu.rs code is **byte-identical** to #649's phase-3d commit (verified by `git range-diff`).

2. **`test_runner.rs` — test renumbered 652 → 707.** Master independently reused test number **652** (`test_652_percpu_audit_image_off_stack`) and added tests 653–657 after `test_651_resched_ipi`. To avoid two "Test 652" labels in the serial/CI output, the new load-balance test is renumbered to **Test 707** (the next free slot above the current max, 706). This is the only code change vs. #649: the dispatch call, the `fn` name, the `NAME` string, and the comment header all move 652 → 707; the seven test-case bodies are unchanged.

3. **`test_runner.rs` — shared-trailer disentangle.** The 3-way merge tangled the identical `test_pass!(NAME); true }` trailer shared by master's last new test (the TSC-deadline cadence test) and #649's load-balance test. Resolved by explicitly closing master's last test with its own trailer, so both functions are well-formed.

### Validation

Per the current host constraints this was validated by `cargo +nightly check` (via the harness `check`, correct bare-metal target) for the default kernel and the `firefox-test-core` profile; full boot/test validation is GitHub CI. Test 707 proves the pure decision (`test_pick_balance`): threshold gating, busiest→emptiest direction, most-overdue victim, pull-on-idle, no-eligible-victim skip, uniprocessor no-op, 3-CPU global-extreme selection, and the deadline-tie lower-tid tie-break.

### DIFF
--- a/kernel/src/sched/percpu.rs
+++ b/kernel/src/sched/percpu.rs
@@ -828,7 +828,7 @@ fn pick_balance_core(
     // Most-overdue eligible victim that is HOMED on the busiest CPU.  Ties on
     // the absolute deadline are broken toward the LOWER tid so the choice is
     // deterministic and independent of iteration order — this keeps the pure
-    // `pick_balance_core` (driven by Test 652 over a candidate slice) and the
+    // `pick_balance_core` (driven by Test 707 over a candidate slice) and the
     // live `pick_balance_victim` (driven over the thread table) selecting the
     // identical victim even when two equally-overdue threads compete.
     let mut victim: Option<(Tid, u8, u64)> = None;
@@ -849,7 +849,7 @@ fn pick_balance_core(
     victim.map(|(tid, prio, _)| (busiest, emptiest, tid, prio))
 }
 
-/// Test-facing replica of the load-balance decision (Test 652), over
+/// Test-facing replica of the load-balance decision (Test 707), over
 /// caller-supplied loads + candidates so the migration policy is provable
 /// without spinning threads or touching the live `RQS`.
 #[doc(hidden)]
@@ -861,21 +861,28 @@ pub fn test_pick_balance(
     pick_balance_core(loads, ncpus, candidates)
 }
 
-/// Live load-balance victim selection: build the eligible-candidate list from
-/// the locked table (applying the migration guards) and the held-guard loads,
-/// then run [`pick_balance_core`].  Allocation-free: it makes a single pass over
-/// the table tracking only the running best per-CPU victim, so no candidate
-/// vector is built.
-fn pick_balance_victim(
-    guards: &[Option<spin::MutexGuard<'_, PerCpuRq>>],
+/// Live load-balance victim selection over the thread table.  Allocation-free:
+/// a single pass over the table tracking only the running best per-CPU victim,
+/// so no candidate vector is built.
+///
+/// Per-CPU load is read through the `load` closure rather than off the guards
+/// directly, matching the wake-edge spread ([`wake_target_with_loads`]): the
+/// LIVE caller wires `|cpu| cpu_wake_load(cpu, nr_running)` so a dead-LAPIC-timer
+/// CPU reports `u32::MAX` and is never selected as the migration destination
+/// (the Leg-3 dead-timer exclusion).  Reading `nr_running` raw here would rank a
+/// dead-timer CPU — whose runqueue stays shallow precisely because the wake edge
+/// already steers work away from it — as `emptiest` and strand the most-overdue
+/// thread on the one CPU that cannot drain it.  The closure also lets the
+/// selection be driven directly over caller-supplied loads in Test 707 (the same
+/// pattern [`test_wake_target`] uses for the wake edge).
+pub(crate) fn pick_balance_victim(
+    load: impl Fn(usize) -> u32,
     ncpus: usize,
     threads: &[proc::Thread],
-    _here: usize,
 ) -> Option<(usize, usize, Tid, u8)> {
-    // Loads from the held guards (no re-lock).
     let mut loads = [0u32; MAX_CPUS];
     for cpu in 0..ncpus {
-        loads[cpu] = guards[cpu].as_ref().map_or(0, |g| g.nr_running());
+        loads[cpu] = load(cpu);
     }
     // Identify busiest/emptiest + threshold via the pure core over a single
     // synthesized best-victim-per-busiest candidate.  We first find busiest so
@@ -925,6 +932,48 @@ fn pick_balance_victim(
         }
     }
     best.map(|(tid, prio, _)| (busiest, emptiest, tid, prio))
+}
+
+/// Apply a chosen load-balance migration under the runqueue guards
+/// `mirror_maintain` already holds: dequeue the victim from its source rq,
+/// enqueue it on the destination, re-home its `last_cpu` + `mirror_slot`, and
+/// bump [`BALANCE_MIGRATIONS`].  Returns `true` iff the migration was applied.
+///
+/// Extracted from [`mirror_maintain`] so the live apply and the migration
+/// counter are exercised directly (Test 707) rather than only inferred from the
+/// pure decision.  The mirror invariant (slot ⟺ rq membership, stable across the
+/// held guards) guarantees the victim is present in the source bucket; the
+/// dequeue result is asserted and, if it ever misses, the enqueue/re-home/count
+/// are skipped so a weakened invariant cannot silently double-count `nr_running`
+/// on the destination (the same defensiveness [`try_steal_to`] applies).
+pub(crate) fn apply_balance_migration(
+    guards: &mut [Option<spin::MutexGuard<'_, PerCpuRq>>],
+    threads: &mut [proc::Thread],
+    src: usize,
+    dst: usize,
+    victim_tid: Tid,
+    victim_prio: u8,
+) -> bool {
+    let removed = match guards.get_mut(src).and_then(|g| g.as_mut()) {
+        Some(g) => g.dequeue(victim_tid, victim_prio),
+        None => false,
+    };
+    debug_assert!(
+        removed,
+        "balance victim absent from its source runqueue (mirror invariant weakened)"
+    );
+    if !removed {
+        return false;
+    }
+    if let Some(g) = guards.get_mut(dst).and_then(|g| g.as_mut()) {
+        g.enqueue(victim_tid, victim_prio);
+    }
+    if let Some(t) = threads.iter_mut().find(|t| t.tid == victim_tid) {
+        t.last_cpu = dst as u8;
+        t.mirror_slot = Some((dst as u8, victim_prio));
+    }
+    BALANCE_MIGRATIONS.fetch_add(1, Ordering::Relaxed);
+    true
 }
 
 /// True iff `t` belongs to the mirrored non-idle runnable pool: a `Ready`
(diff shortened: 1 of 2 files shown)


me1iissa:
> **Independent adversarial review — VERDICT: REQUEST CHANGES.** Posted as a comment because the review account owns this PR and GitHub blocks a formal request-changes on one's own PR; treat this as the formal verdict.

## Adversarial review — per-CPU runqueue load balancing (phase 3d)

Reviewed the migration logic against the invariants the scheduler already
maintains (derived from the code, not the PR description). One correctness
regression that is invisible to the tests, one substantive test-coverage gap,
and some minor notes. Requesting changes on the first.

The good news up front — the two hazards this codebase has been bitten by
before are **not** re-introduced (details in "Checked and cleared" at the end),
so this is not a crash/double-dispatch finding. It is a fairness/liveness
regression on the SMP=2 substrate the phase series targets.

---

### 1. (merge-blocking for SMP=2) Balance destination selection does not exclude a dead-timer CPU — it strands the most-overdue thread on a CPU that cannot run it

`pick_balance_victim` computes per-CPU load from the **raw** runqueue depth:

`kernel/src/sched/percpu.rs:878`
```rust
loads[cpu] = guards[cpu].as_ref().map_or(0, |g| g.nr_running());
```
and then picks `emptiest` = argmin(`loads`) as the migration destination.

Contrast the wake-edge target selection a few dozen lines below, which reads
load through `cpu_wake_load`:

`kernel/src/sched/percpu.rs:1145`
```rust
|cpu| cpu_wake_load(cpu, guards[cpu].as_ref().map_or(0, |g| g.nr_running())),
```

`cpu_wake_load` reports a CPU whose LAPIC periodic timer has been declared dead
(`irq::cpu_timer_dead`) as `u32::MAX`, precisely so the load-aware spread never
routes work onto a CPU that cannot drain its own runqueue (the "Leg 3"
dead-timer exclusion documented in `mirror_maintain`, ~lines 1133–1142). The
balance path re-implements least-loaded destination selection but **omits that
exclusion**.

**Failure scenario (SMP=2):** CPU0's periodic timer is dead (a real, recurring
mode on this substrate — the BSP-vCPU timer-injection artifact). Because the
wake edge already steers new wakeups away from CPU0, CPU0's `nr_running` stays
low, so `pick_balance_victim` ranks CPU0 as `emptiest`. Balance then selects the
**most-overdue** eligible thread on the busiest CPU (smallest absolute
force-deadline — i.e. the thread closest to starving, `percpu.rs:909`) and
migrates it onto CPU0. CPU0's picker does not run, so the thread sits on CPU0's
runqueue unrun. Balance thus takes the thread with the least starvation headroom
and parks it on the one CPU least able to schedule it — actively undoing the
Leg-3 protection every `BALANCE_EVERY_PASSES` passes. Recovery depends on a
*live* peer happening to go fully idle and stealing it back via `try_steal_to`;
until then the thread is stranded.

Note the busiest side is fine (a dead CPU that piled up work looks busiest and
gets drained), so this is purely a destination-side omission.

**Fix:** read the balance loads through `cpu_wake_load` exactly as the wake edge
does, so a dead-timer CPU is never chosen as `emptiest`. One-line change; keeps
the two destination-selection sites consistent.

This is gated on `ncpus > 1`, so it does not affect the current SMP=1 default —
but it is a latent regression that ships the moment SMP=2 is enabled, which is
the stated end-goal of this phase series, and it is cheap to fix now.

---

### 2. (should-fix) The live migration path has zero test coverage; Test 707 exercises a parallel reimplementation, not the code that runs

`pick_balance_victim` (the live selector, `percpu.rs:869`) does **not** call the
tested `pick_balance_core` (`percpu.rs:799`) — it is a hand-duplicated second
implementation of busiest/emptiest + victim selection over the thread table.
Test 707 drives only `test_pick_balance` → `pick_balance_core`. Grep confirms no
test references `pick_balance_victim`, `BALANCE_MIGRATIONS`, or
`balance_migrations`.

Consequences:

- The two implementations are kept in sync **by hand**. The dead-timer omission
  in finding #1 is exactly the kind of divergence Test 707 cannot catch: the
  test supplies clean `loads` as a parameter, so it never exercises the
  raw-`nr_running` read that the live path uses.
- Everything that only exists in the live path is untested: the eligibility
  guards (`ctx_rsp_valid`/`cpu_affinity`/`mirror_slot` homing), the actual
  dequeue/enqueue/re-home + `last_cpu`/`mirror_slot` rewrite, the `resched_mask`
  poke, and the "Running-task is excluded by construction" ordering invariant
  (the PR body itself acknowledges this last one is only "covered by the
  by-construction argument in the code").

The pure-decision cases in Test 707 are correct and worth keeping, but the
riskiest code — live migration under the mirror invariant — has no automated
check. At minimum, drive `pick_balance_victim` over a synthetic
`&[proc::Thread]` slice (the pattern `test_steal_to`/Test 657 already
establishes for `try_steal_to`), and assert `BALANCE_MIGRATIONS` advances so
there is a live-path signal that a migration ever happens. Ideally, have
`pick_balance_victim` delegate its selection to `pick_balance_core` so there is
one implementation and the test covers it.

---

### Minor / non-blocking

- **`dequeue` return value ignored** (`percpu.rs:1242`): the balance migration
  discards the `bool` from `g.dequeue(victim_tid, victim_prio)`, relying on the
  mirror invariant (slot ⟺ rq membership, held stable across the guards) to
  guarantee the victim is present. That reasoning is sound *today*, but
  `try_steal_to` deliberately checks this return and skips on a miss to avoid a
  double-enqueue. If the invariant is ever weakened, the silent path here
  double-counts `nr_running` on `dst`. Cheap to make it defensive (skip the
  enqueue/re-home if the dequeue missed).

- **Possible balance ↔ work-steal oscillation:** balance can move a thread
  busiest→emptiest, after which an idle peer's `try_steal_to` can pull it back.
  Bounded (balance fires once per `BALANCE_EVERY_PASSES`; steal only on an empty
  mirror set) and not a correctness issue, but worth a comment that the two
  migration sources can fight under some load shapes.

- **Stale "Test 652" references:** the PR body says the test was renumbered
  652 → 707, but the doc comments on `pick_balance_core`/`test_pick_balance`
  still cite "Test 652" (`percpu.rs:831` and `:852`). Cosmetic; please update so
  the reference matches the actual test number.

---

### Checked and cleared (so a future reviewer need not re-derive these)

- **#655 double-dispatch is NOT re-introduced.** The balance block adds no new
  dispatch site — it only re-homes a thread's `mirror_slot`. Every actual resume
  of a Ready thread still funnels through the pick-time on-CPU interlock
  (`kernel/src/sched/mod.rs:2609`,
  `.filter(|&idx| !defer_if_live_on_other_cpu(...))`), which defers any
  candidate reported `Current`/`OnStack` on another CPU
  (`proc::other_cpu_liveness`). So even if balance migrates a thread that is
  still on-stack in a source CPU's `switch_context_asm` epilogue, the
  destination CPU will not resume it until that CPU is provably off the stack.
  Migration ≠ dispatch; the chokepoint backstops it.

- **Running-task exclusion is sound.** The delta loop runs before the balance
  block under a single `THREAD_TABLE` hold and sets `mirror_slot =
  desired_slot(t)` for every thread; `desired_slot` is `None` for any non-Ready
  thread, so by the balance point every Running (and every non-Ready) thread has
  `mirror_slot = None` and cannot match the `Some((busiest, _))` victim
  predicate. The `Some(...)` slot therefore also implies Ready-and-non-idle, so
  the absent explicit state/idle checks are covered.

- **TID 0 (the BSP reactor) cannot be balanced.** It is pinned
  `cpu_affinity = Some(0)` (`kernel/src/proc/mod.rs:1359`), so
  `pick_balance_victim`'s `cpu_affinity.is_some()` guard skips it. This closes
  the concern that `other_cpu_liveness` special-cases `tid == 0` to `None`
  (`proc/mod.rs:1572`) — that interlock hole is unreachable via balance.

- **No new lock surface / lock-order regression.** Migration happens under the
  `RQS` guards `mirror_maintain` already holds in ascending CPU-index order; the
  `resched_mask` IPI is drained only after `drop(guards)` (`percpu.rs:1349`).

- **Gated audit stays consistent after a migration.** Setting
  `last_cpu = dst` makes `desired_slot(victim) == Some((dst, prio)) ==` the new
  `mirror_slot`, since `target_cpu_for(None, dst) == dst` (`percpu.rs:643`). So
  the every-64th-pass audit and the next delta pass both agree with the moved
  state — no spurious `MIRROR_AUDIT_FAILURES`, and the migration is not
  immediately undone.

- **Index/shift safety.** `MAX_CPUS == 16`, so `busiest`/`emptiest`/`dst` are
  all `< 32` and the `1u32 << dst` poke cannot overshift; the explicit `dst < 32`
  guard is redundant but harmless.

- **`min_deadline` staleness after a migration is inert** — it has no live
  consumer (`overdue()`/`min_deadline()` are read only by the test suite), so
  the one-pass staleness the migration introduces before the next recompute has
  no scheduling effect.

me1iissa:
> **Re-review of the fix-up delta (d7831a79 → 3d1e42b1, +211/−33) — VERDICT: APPROVE.** Posted as a comment for the same ownership reason as the first review; treat as the formal verdict. Both prior findings are resolved and the refactor is behavior-preserving. Confidence ~0.85.

### F1 (dead-timer destination) — fixed correctly

`pick_balance_victim` now takes a `load: impl Fn(usize) -> u32` reader and the live `mirror_maintain` call site wires `|cpu| cpu_wake_load(cpu, guards[cpu]…nr_running())` (`percpu.rs`, the balance block). A dead-LAPIC-timer CPU therefore reports `u32::MAX` and can never be `emptiest`, so it is never chosen as a migration destination — identical to the wake-edge spread. Verified:

- **Both ends behave sanely.** With `cpu_wake_load`, a dead-timer CPU reports `u32::MAX`, so it now ranks *busiest*, not emptiest. That is the desirable direction: balance migrates the most-overdue eligible thread *off* the dead CPU onto a live one (un-stranding it). It terminates — once the dead CPU has no eligible homed victim, selection returns `None`; the `u32::MAX` load does not depend on the (now-shrinking) real depth, so there is no oscillation. All-CPUs-dead degenerates to `busiest == emptiest → None` (no-op). No pathological loop, no double-count.
- **One minor limitation (non-blocking):** because a dead CPU always pins the `busiest` slot at `u32::MAX`, on a 3+-CPU config with one dead CPU, an imbalance between two *live* CPUs is not corrected until the dead CPU recovers — balancing degrades to "drain the dead CPU only." Moot for the SMP=2 target (only one live CPU when the other's timer is dead), and strictly better than the pre-fix behavior (which stranded onto the dead CPU). Reasonable tradeoff; worth a one-line comment but not a change request.

### F2 (live-path coverage) — addressed

`apply_balance_migration` is extracted so the dequeue/enqueue/re-home/counter path is callable directly, and Test 708 now drives the code that actually runs:

- **Case A** exercises the LIVE `pick_balance_victim` over a synthetic `&[Thread]` slice — genuinely hitting the eligibility guards (pinned tid 102 skipped, mid-switch tid 103 skipped, most-overdue tid 101 chosen) — then the LIVE `apply_balance_migration` over held `PerCpuRq` guards, asserting rq depths (1/1), the re-home (`mirror_slot=(1,8)`, `last_cpu=1`), and that `BALANCE_MIGRATIONS` advances. This is real live-path coverage, not the pure core.
- **Case B** is a sound regression contrast: it first proves the input genuinely produces the hazard under raw load (`dst == 1`), then proves a `u32::MAX`-returning closure excludes that CPU (`None`/`dst != 1`).

### Behavior-preservation of the extraction — verified

- Migration still happens under the `RQS` guards `mirror_maintain` already holds in ascending CPU-index order; `apply_balance_migration` takes `&mut guards` and acquires no new lock. No new lock surface.
- The reschedule poke is now gated on `apply_balance_migration(...) && dst != here && dst < 32`, so the `resched_mask` bit is set only when the migration was actually applied — and the IPI is still drained after `drop(guards)` (`percpu.rs:1394`/`1402`), outside the locked window. Correct.
- The balance block was not moved relative to the delta loop / `min_deadline` recompute, so every previously-cleared item still holds: #655 double-dispatch is still backstopped by the pick-time interlock (`mod.rs:2609`); the Running-task-by-construction exclusion is intact; the gated audit stays consistent (`last_cpu=dst ⇒ desired_slot==mirror_slot`); TID 0 remains pinned and unbalanceable.
- `dequeue`-miss is now defensive: `apply_balance_migration` asserts `removed` and skips the enqueue/re-home/count on a miss, so a weakened invariant cannot double-count `nr_running` (matching `try_steal_to`). Minor #1 resolved. The "Test 652 → 707" doc references are corrected; no residual `652` mislabel remains in `percpu.rs`.
- Build safety: Test 708 is `#[cfg(any(feature = "firefox-test-core", feature = "test-mode"))]` and its call site sits inside the identically-gated dispatch block, so non-firefox builds are unaffected.

### One residual to record (not blocking)

Test 708 Case B validates the *selector's* response to a `u32::MAX` load via a simulated `dead1` closure; it does not execute the live `mirror_maintain → cpu_wake_load` wiring (that would need a booted dead-timer CPU, which the host constraint precludes). So the actual F1 wiring rests on code inspection — which I performed and confirm is correct — rather than an executable assertion; a hypothetical future revert of just the live call-site closure back to raw `nr_running` would not be caught by Test 708. Acceptable given the constraint, but the coverage boundary should be understood.

Nothing in the +211/−33 outside these areas; net a clean, behavior-preserving hardening. Approving.

PR #724, opened 2026-07-17
mm/madvise: stop zeroing anonymous frames in place (W215 root cause)

me1iissa:
## What

Removes the in-place zero-fill from `sys_madvise`'s `MADV_DONTNEED` / `MADV_FREE`
path. This is the root cause of the long-running **W215** page-corruption class.

## Mechanism

The anonymous arm of `sys_madvise` wrote `write_bytes(PHYS_OFF+phys, 0, PAGE)`
into each frame **in place** before the deferred free.

- **Unshared frame** — redundant. Clearing the PTE already makes the next access
  demand-fault a fresh zero-filled frame (POSIX `madvise(2)`: *"subsequent
  accesses ... will result in ... zero-fill-on-demand pages"*).
- **Shared frame** — corrupting. After `fork`, a copy-on-write anonymous frame is
  co-mapped by parent and child, so its refcount is ≥ 2. When one side
  `madvise(MADV_DONTNEED)`s its view, the in-place zero scribbles the frame while
  the co-mapper **still maps it**: `page_ref_dec` only lowers the count to ≥ 1, so
  the frame is **not** freed and the sibling keeps reading it — now all zeros where
  its allocator metadata used to be. The sibling's next metadata read faults.

That is precisely the W215 "victim frame served all-zero" fingerprint: a musl
mallocng group-header read of a **live, never-freed** anonymous frame that came
back zeroed (the `CR2=0x10` / `0x20` ld-musl face), plus the `CR2=0x0`
pointer-chase and `MOZ_RELEASE_ASSERT` faces.

## Evidence — positive attribution

An `exp#2` per-pfn zero-fill provenance probe (diagnostic-only, on the
`diag/w215-victim-classify` branch) placed a ring stamp immediately **before** the
in-place write. On a corrupting Wikipedia boot it named this exact site:

```
[W215/ZERO-LIVE] phys=0x53049000 pid=1 tid=20 off=0x0 len=0x1000 rc=1 sc=1 site=.../subsys/linux/syscall.rs  (the anon in-place zero)
... 30+ consecutive frames of a contiguous anon arena, pid 1 ...
```

This is the writer the saga never had, named by site + phys + refcount-at-zero-time.
Because the madvise zero is the **last** writer of a victim frame, a faulting
victim's `[W215/ZERO-PROV]` read reports `state=EXACT` at this site.

**Honest nuance (surfaced, not hidden):** `pte_share_count` is a thin alias of
`page_ref_count`, so `rc` and `sc` in that line are the same number. The
rate-limited live sample happened to capture the common `rc=1` traffic (pid 1
recycling whole anon arenas), not the rarer `rc≥2` *shared* event the corruption
requires. The shared-frame (`rc≥2`) clobber is therefore proven **deterministically**
by Test 723(a) below rather than relying on a lucky live capture; the live
`ZERO-LIVE` establishes that this site zeroes *live, mapped* frames at all. This is
consistent with the `exp#1` verdict (victim `INPLACE_NEVERFREED`, `rc≥1`, partial
zeros over surviving data) and with the refutations that must hold (free-shadow
`EMPTY` ⇒ not recycle/UAF; `[PMM/ALLOC-NONZERO-RC]=0` ⇒ not PMM double-alloc).

## Fix

Delete the in-place zero. Both advices are served by clearing the PTE and dropping
this mapping's page reference; the frame returns to the PMM only at refcount zero.
Shared frames are left intact for their remaining mappers; unshared frames still
refault to fresh zeros. The file-backed/anonymous VMA snapshot and its
`MADV_ZERO_SUPPRESSED` telemetry — whose sole purpose was to gate the in-place
zero — are removed with it; the clear-PTE path is identical for both backings.
The three-pass clear → shootdown → free structure is unchanged.

## Test plan — Test 723 (kernel suite, drives the REAL `sys_madvise` via dispatch nr=28)

- **(a) shared-frame survival** — a refcount-2 frame `madvise(DONTNEED)`'d through
  one mapping keeps its bytes and is dec'd (not freed).
  - **Fails on current master**: the in-place zero clobbers the frame → the check
    reads it back all-zero.
  - **Passes with this fix**: frame `rc 2→1`, bytes intact.
- **(b) DONTNEED semantics** — an unshared frame `madvise(DONTNEED)`'d is
  PTE-cleared and the next access demand-faults fresh zeros.

`cargo +nightly check` clean under both `--features test-mode` and
`--features firefox-test-core`. CI is the gate.

## Review

This is an mm-path change; it warrants an independent review of the refcount /
shootdown accounting around the deletion. Requesting one author-side review; a
second unseeded review is expected per project policy. **Do not merge** — the
coordinator merges.

## Citations

POSIX.1-2017 `madvise(2)` (`MADV_DONTNEED` / `MADV_FREE`); Intel SDM Vol. 3A
§4.10.5 (propagate paging-structure changes before a frame is reused).

### DIFF
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -6239,19 +6239,6 @@ fn sys_select_linux(
     ready
 }
 
-/// W215 telemetry: number of MADV_DONTNEED/MADV_FREE per-page zero-fills
-/// suppressed by the file-backed guard.  Read from kdb / serial logging.
-#[cfg(feature = "firefox-test-core")]
-static MADV_ZERO_SUPPRESSED: core::sync::atomic::AtomicU64 =
-    core::sync::atomic::AtomicU64::new(0);
-
-/// Number of zero-fills the W215 file-backed guard has suppressed.  Always 0
-/// on non-firefox-test builds.
-#[cfg(feature = "firefox-test-core")]
-pub fn madv_zero_suppressed_count() -> u64 {
-    MADV_ZERO_SUPPRESSED.load(core::sync::atomic::Ordering::Relaxed)
-}
-
 /// madvise(addr, len, advice) — memory usage hint.
 ///
 /// MADV_DONTNEED (4) and MADV_FREE (8): free physical pages in range so the
@@ -6266,35 +6253,17 @@ fn sys_madvise(addr: u64, len: u64, advice: u64) -> i64 {
     let start = addr & !0xFFF;
     let end   = (addr + len + 0xFFF) & !0xFFF;
 
-    // Capture cr3 AND a snapshot of file-backed VMA ranges that overlap
-    // [start, end).  The snapshot is used below to suppress the per-page
-    // zero-fill on file-backed pages, which (when a frame is also held by
-    // the page cache) would silently corrupt the cache content shared with
-    // other VmSpaces and future faulters — the W215 root cause.  See the
-    // detailed rationale at the zero-fill site for the bug shape.
-    //
-    // Snapshot size is tiny in practice: a single madvise range typically
-    // overlaps one VMA, occasionally two when straddling an arena boundary.
-    let (cr3_opt, file_ranges) = {
+    // Resolve the calling process's CR3.  No per-VMA snapshot is needed: the
+    // clear-PTE path below is identical for anonymous and file-backed pages —
+    // neither is written in place, so there is nothing to gate on the backing
+    // type.  Each page simply re-faults on the next access (fresh zeros for
+    // anon, authoritative cache content for file-backed).
+    let cr3 = {
         let procs = crate::proc::PROCESS_TABLE.lock();
         let p = procs.iter().find(|p| p.pid == pid);
-        let cr3 = p.and_then(|p| p.vm_space.as_ref()).map(|vs| vs.cr3);
-        let mut ranges: alloc::vec::Vec<(u64, u64)> = alloc::vec::Vec::new();
-        if let Some(vs) = p.and_then(|p| p.vm_space.as_ref()) {
-            for vma in vs.areas.iter() {
-                if !vma.overlaps(start, end - start) { continue; }
-                if matches!(vma.backing, crate::mm::vma::VmBacking::File { .. }) {
-                    ranges.push((vma.base, vma.end()));
-                }
-            }
-        }
-        (cr3, ranges)
+        p.and_then(|p| p.vm_space.as_ref()).map(|vs| vs.cr3)
     };
-    let cr3 = match cr3_opt { Some(c) => c, None => return 0 };
-    // Closure: O(file_ranges.len()) per page, where len is typically 0-2.
-    let is_file_backed = |page: u64| -> bool {
-        file_ranges.iter().any(|&(b, e)| page >= b && page < e)
-    };
+    let cr3 = match cr3 { Some(c) => c, None => return 0 };
 
     // W216 H_5j-B: MADV_DONTNEED clears PTEs and frees frames; the area list
     // itself is not mutated, but the PFH install loop must abort if it has
@@ -6329,7 +6298,7 @@ fn sys_madvise(addr: u64, len: u64, advice: u64) -> i64 {
     // the sibling thread reads kernel data through the stale mapping.
     //
     // Three-pass approach:
-    //   Pass 1 — zero and clear PTEs, collect physical addresses to free.
+    //   Pass 1 — clear PTEs, drop refs, collect physical addresses to free.
     //   Pass 2 — synchronous shootdown (IPI + ack) across all CPUs.
     //   Pass 3 — return frames to the PMM (safe after pass 2 completes).
     //
@@ -6343,7 +6312,6 @@ fn sys_madvise(addr: u64, len: u64, advice: u64) -> i64 {
     // no caller observes the batch boundary.
     //
     // Cite: madvise(2) (POSIX.1-2017); Intel SDM Vol. 3A §4.10.4-5.
-    const PHYS_OFF: u64 = 0xFFFF_8000_0000_0000;
     const BATCH: usize = 128;
 
     let mut page = start;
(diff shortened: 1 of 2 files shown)


me1iissa:
## Author-side review (context-rich pass) — **APPROVE**

Head `92693524`, base `b5c9d090`. Verified read-only against the PR tree objects; the PR was not compiled locally (my working tree is a different branch and the host is owned by a diagnostic campaign) — remote CI is the compile gate, and the completeness checks below are against the committed tree, not a local build. Findings are hypothesis → evidence → confidence.

### 1. Deletion completeness & side-effects — **clean (0.95)**
- All three removed symbols are orphan-free at the PR tree: `MADV_ZERO_SUPPRESSED`, `madv_zero_suppressed_count`, and the `is_file_backed`/`file_ranges` snapshot each have **zero** remaining references (tree-wide grep of `92693524`). The removed `pub fn madv_zero_suppressed_count` had **no callers even at base** — it was diagnostic-only, read via kdb/serial. No compile break from the removal.
- The function-local `const PHYS_OFF` deleted from `sys_madvise` was scoped to that function; the other `PHYS_OFF` occurrences in the file are independent local consts in unrelated functions. Correct to drop.
- The three-pass **clear → shootdown → free** structure, the `BATCH=128` boundary, the `mm_sem` read-lock hold, `bump_generation_for_cr3`, and the **quarantine-on-dirty-shootdown** path are behavior-preserved for every case except the deleted zero (`subsys/linux/syscall.rs`, the `sys_madvise` Pass-1/2/3 body). Confirmed by diffing the post-fix body against the twin `unmap_and_free_range_in` (`mm/vmm.rs:949`), which is structurally identical.

### 2. MADV_DONTNEED semantics, incl. MAP_SHARED anon — **no regression (0.9)**
- The change deletes a *write* and nothing else — it cannot introduce data loss for a shared mapping, only stop destroying it. For a co-mapped frame (refcount ≥ 2, e.g. across `fork`), DONTNEED clears the PTE and `page_ref_dec` lowers the count to ≥ 1, so the **co-mapper's frame survives intact** — which is the fix. Pre-fix, the in-place zero scribbled that live co-mapped frame (POSIX madvise(2): DONTNEED must not disturb other mappings' invariants).
- `VmBacking` has only `Anonymous` and `File` (`mm/vma.rs:50`) — there is no distinct shmem backing. So a `MAP_SHARED|MAP_ANONYMOUS` region's advising side refaults to **fresh zeros** rather than re-finding a shared page. This is a **pre-existing** property of the fault path, orthogonal to this PR, and acceptable for the *advising* side under POSIX DONTNEED ("subsequent accesses … result in … zero-fill-on-demand pages"). The PR does not worsen it; it removes the write that previously corrupted the *co-mapper's* view. Not a blocker.

### 3. MADV_FREE — **spec-legal, unchanged by this PR (0.85)**
- POSIX/`madvise(2)`: after `MADV_FREE` a subsequent read may legally return **either the old data or zeros**. Our path serves FREE *eagerly* (equivalent to DONTNEED → refault-zeros), which is within the permitted outcome set; musl mallocng tolerates both. The eager-vs-lazy choice and the absence of an `EINVAL` for non-anon `MADV_FREE` are both **pre-existing** and untouched here.
- **Non-blocking nit:** the PR body/commit assert the fix covers "both advices," but Test 723 only drives `MADV_DONTNEED` (advice=4). `sys_madvise` treats 4 and 8 identically (single guard, no per-advice branch below it), so advice=4 functionally exercises the advice=8 path — but a one-line advice=8 assertion would make the regression guard match the claim, and would document the eager-FREE choice.

### 4. Test 723 — **genuine regression guard (0.92)**
- Drives the **real dispatch** (`dispatch_linux_kernel(28, …)`) for both (a) and (b) — not a reimplementation.
- (a) constructs refcount=2 via one real PTE + one manual `page_ref_inc`. This faithfully mirrors the **only** invariant `sys_madvise` inspects — the count returned by `page_ref_dec` (it does not walk sibling PTEs) — so the manual inc is a sound model of the CoW-sibling case.
- (a) is a true **fail-without / pass-with**: pre-fix the in-place zero runs regardless of refcount, so the `pat(off)` byte check trips at `off=0` (`pat(0)=0x00C0FFEE00000000`, non-zero) → clobber detected. Refcount accounting is balanced (2 refs, 2 decs, freed exactly once); the trailing `munmap` sees a cleared PTE, so no double-free.
- (b) genuinely observes a **refault**: the DONTNEED shootdown flushes the local TLB, so the read faults to a fresh zero frame (reads 0), rather than a stale-TLB read of the old `0xA5..` frame.

### 5. Refcount / shootdown ordering — **preserved (0.95)**
- Free still happens in Pass 3, strictly **after** the Pass-2 synchronous shootdown; batch boundaries and the quarantine fallback are undisturbed. No new free-before-shootdown → re-alloc → stale-TLB-write window (Intel SDM Vol. 3A §4.10.5). The deleted `write_bytes` had no barrier semantics, so its removal cannot affect ordering.
- **Disclosure check:** the twin munmap free path (`mm/vmm.rs:949`) also frees anon frames **without** zeroing them first. The kernel's stale-data-disclosure boundary is therefore **zero-on-allocation (demand fault)**, not zero-on-free — proven by Test 723(b) and by munmap's existing behavior. Deleting the madvise pre-free zero makes `sys_madvise` consistent with munmap and opens **no** new disclosure surface.

### 6. Hygiene — **clean**
- Commit and PR body cite POSIX `madvise(2)` and Intel SDM Vol. 3A §4.10.5 only; no foreign-source attribution. Trailer is `Co-Authored-By: … <noreply@anthropic.com>` only — no session URL, no real email; author is a `noreply` address. Diff scope matches (−77/+39 in `syscall.rs`, +164 test).

### Verdict
**APPROVE.** The deletion is complete and side-effect-free; the removed zero was an active corruptor of shared frames, not a semantics or security mechanism; the surviving path matches POSIX DONTNEED and the SDM ordering rule; Test 723 is a genuine fail-without/pass-with guard. The two nits (add an advice=8 case; a one-line note that MADV_FREE is served eagerly and MAP_SHARED-anon DONTNEED refaults to zeros) are non-blocking. Merge on CI green.

me1iissa:
## Independent review (unseeded, blind) — **APPROVE**

Reviewed the diff against invariants derived from the code itself (refcount lifecycle, PTE clear→shootdown→free ordering, the anon/file demand-fault paths, and `clone_for_fork` CoW), not from the PR body. **No merge-blocking issues.** The change is a genuine, tightly-scoped correctness fix.

### What I verified (load-bearing, for the next reviewer)

1. **The diff is minimal and behaviour-preserving except for the one deletion.** Against base `b5c9d090`, the *only* behavioural change is removing the in-place `write_bytes(PHYS_OFF+phys, 0, PAGE)` on the anonymous arm (plus the now-dead file-backed gate and the `MADV_ZERO_SUPPRESSED` telemetry that existed solely to conditionalise that write). The `write_pte(cr3,page,0)` → `page_ref_dec(phys)` → collect-iff-`new_rc==0` → **Pass 2** synchronous `shootdown_range` → **Pass 3** free/quarantine structure is byte-for-byte identical to the pre-fix code. Deleting a store cannot introduce a lock-order or ordering race.

2. **POSIX/`madvise(2)` `MADV_DONTNEED` zero-fill contract is preserved without the in-place zero.** The anonymous demand-fault path unconditionally zeroes each fresh frame (`idt.rs:3530`, `write_bytes(PHYS_OFF+phys,0,PAGE_SIZE)`) and sets `page_ref_set(phys,1)` at install (`idt.rs:3616`). So a `MADV_DONTNEED`-cleared PTE re-faults to fresh zeros whether or not the frame was zeroed at free time.

3. **No information-leak regression.** `alloc_page` (`pmm.rs:656`) does not zero. Every user-facing frame install zeroes or overwrites: anon fault (`idt.rs:3530`), file fault (`idt.rs:3066`/`:2542` zero-then-read, so partial/EOF reads leave a zero tail), page-table pages (`vmm.rs:500`). Crucially, `munmap`'s backend `unmap_and_free_range_in` (`vmm.rs:918`) already returns anonymous frames to the PMM via `page_ref_dec` **without** zeroing — so dirty-frame-to-PMM is the pre-existing norm and the demand-fault zero is the real security boundary. This PR makes `madvise` consistent with `munmap`; it does not widen any leak.

4. **The shared-frame corruption premise is real.** `clone_for_fork` (`vma.rs:960-975`) write-protects the parent PTE in place, installs the child PTE read-only to the same `phys`, and `page_ref_inc(phys)`. After fork an anonymous CoW frame is therefore rc≥2 and mapped by both sides. The old in-place zero on one side's `MADV_DONTNEED` (`page_ref_dec` 2→1, frame **not** freed) provably scribbles the co-mapper's live data — so this is a correctness fix, not merely a diagnostic tweak.

5. **Shootdown still fires for the shared (freed-nothing) case.** The `any_unmap` flag gates Pass 2 independently of `n` (frames-to-free), so a `MADV_DONTNEED` that clears a PTE but frees nothing (all pages shared) still shoots down the caller's stale TLB entry. Correct.

6. **This was the *unique* in-place zero of an already-mapped frame.** Every other `PHYS_OFF`-relative zero-fill in `mm/`/`arch/` (`vma.rs:726`, `idt.rs:2542/3066/3530`) targets a freshly-allocated pre-install frame (rc 0/unset, unshared). So this is the complete fix for the "in-place zero of a live, possibly-shared frame" class.

7. **No dangling references** to the removed `madv_zero_suppressed_count()` / `MADV_ZERO_SUPPRESSED` anywhere in `kernel/`, `scripts/`, `tools/`. Test 723's symbols all resolve; `dispatch_linux_kernel` is 7-arg matching the calls; happy-path refcount is balanced (mmap→touch rc1, manual inc→rc2, madvise dec→rc1, test dec→rc0+free, munmap no-op on the already-cleared PTE).

### Findings (all LOW / non-blocking)

**L1 — Citation over-attribution (committed prose).** `syscall.rs` comment (~6335-6355), the PR body "Citations", and the commit trailer attribute *"subsequent accesses … will result in … zero-fill-on-demand pages"* to **POSIX.1-2017 madvise(2)**. That clause is the `madvise(2)` **man-page** description of the (non-POSIX) `MADV_DONTNEED`. POSIX.1-2017 standardises `posix_madvise()`/`POSIX_MADV_DONTNEED`, which carries **no** destructive/zero-fill semantics (glibc even implements `POSIX_MADV_DONTNEED` as a no-op to avoid the destructive Linux behaviour). Recommend citing `madvise(2)` (man page) rather than POSIX for the zero-fill/destructive contract. Code correctness unaffected.

**L2 — `MADV_FREE` treated as eager `MADV_DONTNEED`.** The path (pre-existing, unchanged here) clears the PTE, defer-frees, and re-faults to zero for **both** advices. Per `madvise(2)`, `MADV_FREE` is *lazy*: pages are retained until memory pressure, a store before reclaim cancels the free (contents preserved), and only exclusively-mapped anon pages are eligible. Eager freeing is a permitted, stricter implementation, so this is **not** a regression and not introduced by this PR — but the shared comment and the PR's mechanism narrative describe the "re-fault to fresh zeros" contract as if it held for both advices, which is strictly true only for `MADV_DONTNEED`. Optional: scope the zero-fill claim to `MADV_DONTNEED`, or note `MADV_FREE` is implemented as eager-DONTNEED.

**L3 — Stale sibling test.** `test_madvise_dontneed` (`test_runner.rs:25437-25453`) is an *inline reimplementation* that still zeroes the frame in place (`:25442`) and comments *"same as sys_madvise MADV_DONTNEED"* (`:25441`) — now false. It passes independently (it never calls the real syscall), so no CI break, but it is misleading and its coverage is superseded by Test 723, which correctly drives the real dispatch (nr=28). Consider updating the comment or retiring the inline reimplementation.

### Sufficiency caveat (not a review blocker)

The *positive attribution* that this site is THE W215 root cause rests on (a) the code-level proof above that it corrupts shared frames and (b) a live `[W215/ZERO-LIVE]` capture that — by the PR's own honest admission — sampled the rc=1 (redundant, non-corrupting) traffic, not the rc≥2 corrupting instance; the rc≥2 case is demonstrated only by the synthetic Test 723(a). So the change is **necessary and correct on its own merits**, but "fully closes the W215 crash class" is an empirical claim to confirm by soak, not by this diff. Removing a provably-corrupting in-place write on shared frames is right regardless of whether other contributors remain.

**Verdict: APPROVE.** No merge-blocking findings; L1-L3 are optional prose/test cleanups. Confidence: 0.9 the diff is correct and safe to merge; ~0.6 it fully closes W215 (empirical, per the PR's own nuance).

Cite: `madvise(2)` (`MADV_DONTNEED`/`MADV_FREE`); POSIX.1-2017 `posix_madvise`; Intel SDM Vol. 3A §4.10.4-5 (propagate paging-structure changes before a frame is reused).

PR #725, opened 2026-07-17
test: #724 review nits (MADV_FREE coverage, stale twin, cites) + kuser_shared RTC flake fix

me1iissa:
Batch follow-up to #724 (merged `5e887019`). **Test/comment-only — no change to `sys_madvise` behaviour.** Two logical commits.

## Commit 1 — madvise review nits (from the #724 reviews)
- **MADV_FREE coverage (rev-724a):** Test 723's shared-frame-survival assertion now runs for **both** `MADV_DONTNEED` and `MADV_FREE` (advice=4 and advice=8). Both advices share the clear-PTE path, so neither may zero a shared frame in place — the loop proves it for each.
- **Retire the stale twin:** deleted `test_madvise_dontneed` (old Test 94). It reimplemented the *old* in-place-zero path inline and its "same as sys_madvise" comment is now false. Test 723 drives the real `sys_madvise` via dispatch nr=28, so the twin is redundant and misleading. (No hard-coded test total depends on it.)
- **L1 citation:** the `MADV_DONTNEED` zero-fill-on-demand contract is a **madvise(2) man-page** guarantee, not POSIX — cite the man page for it.
- **L2 over-claim:** softened the comment so it no longer implies a zero-fill *guarantee* for `MADV_FREE`, whose man-page semantics are laxer (lazy reclaim; absent an intervening write a later read may return old contents or zeros). Our eager-free is a conformant, stricter-than-required choice.

## Commit 2 — kuser_shared RTC flake
`test_kuser_shared_data` sub-test 7 read the RTC-derived `SystemTime` twice ~20 ms apart and required advance. `SystemTime` has **1-second granularity**, so the two reads usually land in the same second → false "SystemTime did not advance". Replaced the fixed 2-tick wait with a **bounded wait-for-edge** (~3 s / 300 ticks at TICK_HZ=100): advancing within the window passes; never advancing across the whole window is a genuine stall and still fails. Keeps the real regression signal; removes the race. Uses the same `get_ticks()` liveness the adjacent TickCountQuad sub-test already assumes.

## Test plan
`nice -n 19 cargo +nightly check --features test-mode` clean (only pre-existing warnings). Test-only; CI is the gate. Not mm-path — single review suffices.

### DIFF
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -6311,7 +6311,7 @@ fn sys_madvise(addr: u64, len: u64, advice: u64) -> i64 {
     // continuation across multiple batches, so this is a tuning change only:
     // no caller observes the batch boundary.
     //
-    // Cite: madvise(2) (POSIX.1-2017); Intel SDM Vol. 3A §4.10.4-5.
+    // Cite: madvise(2) man page (man7.org); Intel SDM Vol. 3A §4.10.4-5.
     const BATCH: usize = 128;
 
     let mut page = start;
@@ -6384,7 +6384,7 @@ fn sys_madvise(addr: u64, len: u64, advice: u64) -> i64 {
                 // this PTE's released reference.
                 //
                 // Cite: Intel SDM Vol. 3A §4.10.5 (TLB consistency);
-                // POSIX madvise(2) MADV_DONTNEED — the kernel may free
+                // madvise(2) (man page) MADV_DONTNEED — the kernel may free
                 // the pages but must do so without violating concurrent
                 // mapping invariants.
                 let new_rc = crate::mm::refcount::page_ref_dec(phys);
--- a/kernel/src/subsys/linux/syscall.rs
+++ b/kernel/src/subsys/linux/syscall.rs
@@ -6334,9 +6334,15 @@ fn sys_madvise(addr: u64, len: u64, advice: u64) -> i64 {
                 // re-faults — an anonymous page demand-faults a FRESH
                 // zero-filled frame, and a file-backed page re-faults through
                 // the page cache to the authoritative file content — so the
-                // observable result already matches POSIX madvise(2): after
-                // MADV_DONTNEED "subsequent accesses ... will succeed, but will
-                // result in ... zero-fill-on-demand pages".
+                // observable result already matches the madvise(2) man-page
+                // MADV_DONTNEED contract: "subsequent accesses ... will
+                // succeed, but will result in ... zero-fill-on-demand pages".
+                // MADV_FREE's man-page contract is laxer — the pages may be
+                // lazily reclaimed and, absent an intervening write, a later
+                // read may return the old contents OR zeros.  We take the
+                // eager option (free now → refault to zeros): a conformant,
+                // stricter-than-required choice.  NOT writing the frame is
+                // correct under both contracts.
                 //
                 // Zeroing the frame in place (as a prior revision did for the
                 // anonymous arm) is both redundant and unsafe.  Redundant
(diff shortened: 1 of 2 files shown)


me1iissa:
## Independent review — APPROVE (non-blocking nit)

Verified against the diff + surrounding code in `kernel/src/subsys/linux/syscall.rs` and `kernel/src/test_runner.rs` (branch `qa/w215-review-nits-rtc-flake` @ `814b5df5`), plus a clean `qemu-harness.py check --features test-mode` (rc=0) in an isolated worktree.

**1. MADV_FREE loop correctness — confirmed.** `sys_madvise` (syscall.rs:6246-6249) gates only on `advice ∈ {DONTNEED, FREE}` and then runs one identical clear-PTE + `page_ref_dec` path for both — no advice-specific branch in the per-page loop. The `test_723` loop gives each iteration a fully independent lifecycle (fresh `mmap`, fresh demand-fault, fresh pattern stamp, fresh `page_ref_inc` to model the CoW sibling, fresh `munmap`); refcount arithmetic is self-balanced per iteration (1 implicit + 1 explicit inc, madvise-dec + explicit-dec, ends at 0 and freed). No shared/leaked state between the two passes.

**2. Test 94 removal — complete.** No remaining reference to `test_madvise_dontneed` anywhere in the tree (kernel, `ci/`, `scripts/`, docs). `total`/`passed` in `test_runner.rs::run()` are purely dynamic (`total += 1` per call site, `passed == total` compare) — nothing hardcodes an expected count. The deleted test only touched its own locally-created blocked process (and unblocked it before returning), so no other test depended on its side effects.

**3. Citations — accurate but incomplete within the same function.** `madvise()`+`MADV_DONTNEED`/`MADV_FREE` genuinely isn't a POSIX interface (POSIX only standardizes the weaker, advisory-only `posix_madvise()`/`POSIX_MADV_DONTNEED`), so retargeting the citation to the madvise(2) man page is the right correction, and the MADV_FREE softening accurately reflects its laxer lazy-reclaim contract. Nit: two other comment blocks in the *same* `sys_madvise` function still carry the identical incorrect "POSIX" attribution and weren't touched by this diff:
   - `kernel/src/subsys/linux/syscall.rs:6314` — `// Cite: madvise(2) (POSIX.1-2017); Intel SDM Vol. 3A §4.10.4-5.`
   - `kernel/src/subsys/linux/syscall.rs:6387` — `// POSIX madvise(2) MADV_DONTNEED — the kernel may free...`
   Comment-only, non-blocking, but worth a fast follow-up so the function is internally consistent. No forbidden internal-corpus attribution anywhere in the diff.

**4. RTC wait-for-edge — sound, though the stated mechanism is imprecise.** Traced `current_system_time()` → `update_time_fields()` (`kuser_shared.rs:303-333`): the field is RTC-seconds*1e7 **plus a sub-second residue driven by the monotonic tick counter**, so in practice it visibly increments roughly every tick (~10ms at TICK_HZ=100), not strictly once per RTC second — the commit message's "1-second RTC granularity" framing is a bit off; the real trigger for the old flake was more likely incidental scheduling/timing jitter around the read pair. That said, it doesn't matter for correctness: the fix (break-on-edge with a bounded 300-tick/~3s cap) is robust regardless of the precise mechanism — passes almost immediately in the healthy case and only spends the full budget on a genuine frozen-clock regression, which is exactly the failure mode it must still catch. It reuses the same `get_ticks()` liveness assumption sub-test 6 (TickCountQuad) already relies on, so no new ordering dependency, and `saturating_add` guards the deadline math. No monotonicity/backwards-time assertion is added or removed relative to before.

**5. Hygiene — clean.** Both commits are noreply-only authored (`...@users.noreply.github.com` / `noreply@anthropic.com`), no session trailer/URL. The `syscall.rs` hunks touch comment lines exclusively — verified line-by-line that no statement or control flow changed, confirming "no behaviour change to `sys_madvise`."

Verdict: **APPROVE**. Nothing blocking; recommend a quick follow-up to fix the two leftover POSIX mis-citations at `syscall.rs:6314` and `:6387` while this is fresh.

